PR #174, opened 2015-01-11
Initial attempt at a talk creation API endpoint

chartjes:


classes/OpenCFP/Http/Controller/Api/TalksController.php
```diff
@@ -0,0 +1,77 @@
...
+            'slides' => $sanitized_data['slides'],
+            'other' => $sanitized_data['other'],
+            'sponsor' => $sanitized_data['sponsor'],
+            'user_id' => $sanitized_data['user_id'],
+        );
+
+        $talk_mapper = $this->app['spot']->mapper('OpenCFP\Domain\Entity\Talk');
+        $talk = $talk_mapper->create($data);
+
+        return new Response(json_encode(['msg' => 'Talk created']), 200);
```
jakefolio: Should be 201 for the status code.

mdwheele:
It's a great first pass, I believe. There are a few things I would add:

1. If there's a reason to use users' email addresses in key-generation, disregard! I prefer having a service to generate tokens that isn't attached to any factor represented in existing user data. Not that anybody would (because data in a CFP isn't exactly something folks would target as valuable), but technically someone could troll through accounts and brute-force tokens pretty easily.  Here's a diff of what I'm talking about: https://github.com/mdwheele/opencfp/compare/chartjes:api...api-thoughts
2. In `Api\TalksController`, both the token and email are required in order to authenticate the user; treating the token more as a password rather than as *the* credential itself. Is this the intent; to treat the token as a password rather than how you might treat an OAuth2 bearer token?  If it's still up in the air, you could essentially just `getUserByToken` rather than searching by email and checking for a matching token. Also, depending, it might be a good idea to suggest usage of `Authorization` header rather than `POST` or query parameters.
3. Would like to see a way to regenerate token, but think it's outside this initial implementation (because I want something to do! :P)
4. Error reporting for API responses could lump all request validation errors into one response rather than bailing early. Something like.  
```json
{
  "errors": [
    "token": "Missing API Token",
    "email": "..."
  ]
}

{
  "errors": [
    "title": "You must submit a title."
    "description": "...",
    ...
  ]
}
```

chartjes:
On Sun, Jan 11, 2015, at 08:33 PM, Dustin Wheeler wrote:
> It's a great first pass, I believe. There are a few things I
> would add:


>  1. If there's a reason to use users' email addresses in
>     key-generation, disregard! I prefer having a service to generate
>     tokens that isn't attached to any factor represented in existing
>     user data. Not that anybody would (because data in a CFP isn't
>     exactly something folks would target as valuable), but technically
>     someone could troll through accounts and brute-force tokens pretty
>     easily. Here's a diff of what I'm talking about:
>     mdwheele@chartjes:api...api-thoughts[1]

I am not married to any one method of generating tokens, I just picked
one that I thought would create unique enough ones by basing them on the
time the account was crated. Not sure how easily that could be
brute-forced.

>  1. In Api\TalksController, both the token and email are required in
>     order to authenticate the user; treating the token more as a
>     password rather than as *the* credential itself. Is this the
>     intent; to treat the token as a password rather than how you might
>     treat an OAuth2 bearer token? If it's still up in the air, you
>     could essentially just getUserByToken rather than searching by
>     email and checking for a matching token. Also, depending, it might
>     be a good idea to suggest usage of Authorization header rather
>     than POST or query parameters.

The intent was to not allow people to just blindly submit using a token.
If they start guessing md5 values, without the proper email address the
system would at least reject their attempts to create new talks. If they
have the proper token AND the proper email address, they are good to go.

>  1. Would like to see a way to regenerate token, but think it's
>     outside this initial implementation (because I want something to
>     do! :P)
>  2. Error reporting for API responses could lump all request
>     validation errors into one response rather than bailing early.
>     Something like. {
  "errors": [ "token": "Missing API Token", "email": "..." ] }

{ "errors": [ "title": "You must submit a title." "description":
"...", ... ] }

Can't lump them all together without redoing the validation system IIRC
-- it bails on on the first failed thing.





Links:

  1. https://github.com/mdwheele/opencfp/compare/chartjes:api...api-thoughts

classes/OpenCFP/Provider/RouteServiceProvider.php
```diff
@@ -70,6 +70,10 @@ public function register(Application $app)
         // Admin::Review
         $secureRoutes[] = $app->get('/admin/review', 'OpenCFP\Http\Controller\Admin\ReviewController::indexAction')->bind('admin_reviews');
 
+        // API end points
```
mdwheele: Thoughts on following?

| method | route | description |
| --- | --- | --- |
| `POST` | `/api/talks` | Given JSON payload representing a talk, creates talk for authenticated user and issues a 201 Created upon success, appropriate error otherwise |
| `GET` | `/api/talks` | Returns JSON collection of all talks for authenticated user |
| `GET` | `/api/talks/{id}` | Returns a particular talk for authenticated user. Returns appropriate responses for unauthorized or non-existent talks |
| `DELETE` | `/api/talks/{id}` | Removes a talk. |

This structure would allow functionality that would support a feature for "import from another OpenCFP instance" (in response to things like https://twitter.com/ocramius/status/554718836134780929) whereby user could simply provide a link to the root of CFP site and execution could take it from there. 

That said, adding something to `GET /api` would be a good idea to identify the API root (and the fact that the API is enabled or not in the case you make the API able to be toggled by configuration)

chartjes: On Mon, Jan 12, 2015, at 10:03 PM, Dustin Wheeler wrote:
> In classes/OpenCFP/Provider/RouteServiceProvider.php[1]:


> > @@ -70,6 +70,10 @@ public function register(Application $app)
>          // Admin::Review $secureRoutes[] = $app->get('/admin/review',
>          'OpenCFP\Http\Controller\Admin\ReviewController::indexAction')->bind('admin_reviews');
>
> +        // API end points Thoughts on following?


> method route description POST /api/talks Given JSON payload
> representing a talk, creates talk for authenticated user and issues a
> 201 Created upon success, appropriate error otherwise GET /api/talks
> Returns JSON collection of all talks for authenticated user GET
> /api/talks/{id} Returns a particular talk for authenticated user.
> Returns appropriate responses for unauthorized or non-existent talks
> DELETE /api/talks/{id} Removes a talk. This structure would allow
> functionality that would support a feature for "import from another
> OpenCFP instance" (in response to things like
> https://twitter.com/ocramius/status/554718836134780929) whereby user
> could simply provide a link to the root of CFP site and execution
> could take it from there.


> That said, adding something to GET /api would be a good idea to
> identify the API root (and the fact that the API is enabled or not in
> the case you make the API able to be toggled by configuration)


No objections here, and it never occurred to me to make having the API
available a configurable option. Easy enough I imagine to add an entry.



Links:

  1. https://github.com/chartjes/opencfp/pull/174#discussion-diff-22840940

mdwheele:
Amount of work aside, how would you feel about using OAuth2 backed by our local account store for issuing tokens? This would give clients a standard way of doing anything we would open up. 

In the case of an external system publishing talks to cfp through the API, the flow would be more natural for the user of the connected system: "I want X connected system to publish talks for me... Oh it's asking me to authenticate to cfp; makes sense... Oh it's asking me to authorize connected system to publish talks on my behalf."

This is obviously more complex to set up, but I am willing to help there and get some of that foundational stuff done. If so, would probably look into abstracting auth as a service and implement sumpleauth or whatever Jake recommended in lieu of Sentry. 

Thoughts? I know this is a truckload.

PR #176, opened 2015-01-22
Fix swiftmailer configuration not being passed correctly

stevecoug:
The swiftmailer configuration in config/*.yml was not being passed to the swiftmailer provider correctly. This commit puts the swiftmailer configuration inside a swiftmailer.options array, which fixes the problem.


chartjes:
Thanks @stevecoug

PR #178, opened 2015-01-29
Allow sorting on admin talks page

stevecoug:
We have a lot of talks in our CFP (399 so far), and we need to be able to go through them in a more non-random order.

I'd like to set up filters, too, but that's a bit more ambitious.


stevecoug:
Needs a fix

chartjes:
Hi Steve,
The Travis CI build is broken and complaining about this http://i.imgur.com/OpRLT9u.png
If you can fix that and resubmit the PR, that would be good.

chartjes:
Thanks @stevecoug

PR #180, opened 2015-02-24
Update badges of honor!

mdwheele:
Just repointed my repositories and found some loose-ends from the move. 22a1c99 requires registration / update of the repository with Code Climate. TravisCI seems to have picked up everything alright.


chartjes:
Thanks @mdwheele

PR #181, opened 2015-02-26
Added unit test for Signup controller

chartjes:
Wrote test in response to issue179, was a good exercise in figuring out correct mocking of the Application object to remove dependencies on using the database in the unit tests.


jakefolio:
:+1:

Issue #107, opened 2014-08-16
CFP closed view


pmeth:
Once the CFP end date has passed, there is no visual indicaton on the homepage.  The only indication I see is when someone tries to login or create an account.

Also I think we should still probably let people create accounts and login after the CFP is closed, but just not let them submit talks.  For example, how is an administrator supposed to login to perform administrative tasks?

mdwheele:
@pmeth This is pretty old at this point. I've got a few updates and some next steps:

- [ ] There should be visual indication on homepage once the CFP is closed.
  - What should that look like? e.g. "Sorry folks, [...]. Try again next year!"?
- [x] Speakers and admins are able to authenticate after the CFP is closed.
- [x] Speakers cannot submit, edit, or remove talks after CFP is closed.
- [x] Should be able to promote existing users to admin after CFP ends.
- [ ] Should be able to create admin users when they don't already exist with CLI utility.
  - `bin/opencfp admin:promote email` could create the account, emailing user to continue the activation / specifying-password flow. Shouldn't be that difficult and would actually be a good spring-board for wrapping authentication.

If visual indication of closure were implemented in addition to a separate issue detailing "admin user creation" being created, do you think we could close this out as resolved?

mdwheele:
@pmeth Are we good to close this issue out? I'm going to create an issue for the incomplete bullet point above, but since it's not entirely related to this issue, I think everything else is accomplished?

PR #184, opened 2015-03-04
Adds visual indication of CFP ending on the homepage

mdwheele:
When the CFP is on-going, everything is rendered as currently in `master`. When CFP has closed, an alternative message is shown on the homepage that includes how many talks were submitted. Took queue from [phptek](http://tek.phparch.com/call-for-papers-ended-with-620-submissions/) as far as whether or not number of submissions was "a standard thing to report".

Fixes part of #107.

This also implements a global Twig variable that indicates whether or not the CFP is open (`cfp_open`). This would be helpful for #183 to use rather than [sending it in the Twig context manually](https://github.com/mdwheele/opencfp/commit/2a2b3b0eb4a086defff0e265c502a1a115b99b94).

image

### DIFF
--- a/classes/OpenCFP/Http/Controller/PagesController.php
+++ b/classes/OpenCFP/Http/Controller/PagesController.php
@@ -6,16 +6,24 @@ class PagesController extends BaseController
 {
     public function showHomepage()
     {
-        return $this->render('home.twig');
+        return $this->render('home.twig', $this->getContextWithTalksCount());
     }
 
     public function showSpeakerPackage()
     {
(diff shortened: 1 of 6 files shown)


chartjes:
Can we make displaying the number of submitted talks a configurable option?

mdwheele:
Sure thing. Added in fa061d2 and default disabled in previous.

chartjes:
Yeah, the 'Register Now" button should be removed when the CfP is over. Good idea

chartjes:
Thanks @mdwheele

PR #185, opened 2015-03-06
Removes remaining references to previous repository location.

mdwheele:
Noticed this working on #184. A `sed` replace later... :smile:


chartjes:
Thanks @mdwheele

PR #186, opened 2015-03-07
Adds missing test for positive behaviour on Speakers#getTalk.

mdwheele:
Missed writing a test for the new method on `Speakers` to get a specific talk. Added here.


chartjes:
Thanks @mdwheele

PR #187, opened 2015-03-21
Symfony/security-csrf add, ENV add, date formats, and readme update

adamculp:
A few items required for functionality.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -3,3 +3,5 @@
    RewriteRule    ^$ web/    [L]
    RewriteRule    (.*) web/$1 [L]
 </IfModule>
+
+SetEnv CFP-ENV development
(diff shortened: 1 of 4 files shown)


chartjes:
@adamculp Could you add a note to the documentation that we are now setting a default environment value? Some people might miss this and end up with a production install being in dev mode

adamculp:
Very valid point. I see a few things that could be updated in the README. I'll add the note on env value as well and add it to this request.

mdwheele:
Hey @adamculp, wanted to give you a heads up that `CFP-ENV` should be `CFP_ENV`. Here's where that is being resolved: https://github.com/opencfp/opencfp/blob/master/classes/OpenCFP/Environment.php#L49-L54

It's currently working for ya because it's defaulting to `development`.

adamculp:
Thanks. Will note that when updating the readme.

adamculp:
OK, made some updates. Look OK? Ready to merge in?

chartjes:
Thanks @adamculp

Issue #48, opened 2013-06-29
Composer.lock file should be in the repository


hhamon:
When running the `install` command, Composer uses the `composer.lock` file first to install dependencies exactly as they are in this file. This is useful to have this file in the repository in order to always install a stable version of OpenCFP.

PR #192, opened 2015-05-01
Include composer.lock in the repo

lornajane:
I think it would be better to have composer.lock in the repo - it has been put into the repo in the past (see #48 and #49) but looks like it was added to the .gitignore in commit 478c7ffe856fbe0e213de26f7aa34ec444151d86

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,6 @@ opencfp.iml
 .project
 .settings
 composer.phar
-composer.lock
 deployment.properties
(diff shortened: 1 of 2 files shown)


pmeth:
totally agree. This was also talked about in #48.

chartjes:
@pmeth

Yeah, you need to get added in as a contributor now that it belongs to the overall org instead of just me. I will get it done for you

PR #195, opened 2015-05-14
[WIP] Discussion on API Documentation

mdwheele:
**Do not merge**

I'm submitting this as a line of work on a first iteration of API documentation. I will be merging this into the implementation branch once finalized. 

The point of this is to put an overview in front of folks of how all of this will fit together. I'm looking to pare this down a bit as I feel it is wordy and, in places, justifies implementation details rather than just saying, "This is how the shit works." for users to consume. Please rip it apart on those terms.

That said, it also suggests a rough technical direction for the API and there are some assumptions being made, for sure. I wish for those to be challenged / questioned where appropriate. Things that come to mind:

1. We're implementing client registration in compliance with the current [OAuth 2.0 Dynamic Client Registration Protocol](https://tools.ietf.org/html/draft-ietf-oauth-dyn-reg-23) draft. This means that we're able to move to a full implementation later, but what we're allowing (at this point) is for any arbitrary web application / client to register client application on any instance of OpenCFP where the API is enabled. There is no approval process. This is a trade-off. On the one hand, it is MUCH simpler to implement for a consuming application as there is no need for a pre-arranged access token / handshake. On the other hand, it leaves an instance of OpenCFP open for exploitation; where the "exploit" is the ability to create infinite client applications. We can actually take measures against this as a compensating control without penalizing nice users. Just don't want to be driven by FUD here, because it is very convenient, as described currently. Also note that simply "having a client application" buys the application nothing. They will still need to put users through an authorization flow before receiving tokens to do anything.
2. I make assumptions on what is valuable to a consumer as part of the API. This should be challenged by potential consumers.

https://github.com/mdwheele/opencfp/blob/json-api-docs/README.md#json-api (link to pretty)

**Additional Thoughts**

- [ ] It'd be nice to have an endpoint dedicated to serving API status. Is it on? Is it off? Does it only allow certain API actions (future)? In this way, a client could branch on how to handle talk submission if the API is turned off. Perhaps it just redirects them to the CFP site after putting talk contents in clipboard and we support some pasting of that encoded content? WHO KNOWS!? Magic!
- [ ] Double check all anchor links for jumping around. A few of the "see below" links are broken because anchor names changed.
- [ ] Will need a method for Client Application to update its Client Metadata in a next iteration, specifically redirect URIs.

I may add to this. I'm on a break and on-the-move.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,8 +181,9 @@ Note: For updating previously installed instances only run migrations as needed.
 <a name="json-api" />
 ## JSON API
 
-OpenCFP has a JSON API that can be used by third-party applications to take advantage of a set of features on behalf
-of a user. The API is enabled by default, but can be disabled if not needed for your instance of OpenCFP.
+OpenCFP has a JSON API (not to be confused with the [json-api specification](http://jsonapi.org/)) that can be used by
+third-party applications to take advantage of a set of features on behalf of a user. The API is enabled by default, but
+can be disabled if not needed for your instance of OpenCFP.
 
 <a name="json-api-configuration" />
 ### API Configuration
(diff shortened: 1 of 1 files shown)


mattstauffer:
@mdwheele Still reading, but, wow, you're a baller.

mdwheele:
About to fold these into the implementation branch. Closing PR.

PR #196, opened 2015-05-19
Photo Upload Vulnerability

mdwheele:
This pull request patches a flaw in how profile images are managed that creates risk of a persistent remote code execution vulnerability. Details of the vulnerability may be released at the maintainers' discretion.

- Uses `ircmaxell/random-lib` to generate reasonably random filenames whereby no part of the filename, at-rest, is controlled by a user of OpenCFP.
- Adds additional checks to cleanup profile photo uploads that have yet to be cropped and permanently stored.
- Adds a migration to clean existing installs of OpenCFP with regards to this vulnerability; bringing stored profile photos inline with proposed storage scheme.

**Potential Future Work**

- Would like to implement tests that attempt to exploit this vulnerability from an e2e perspective as well as tests that force exceptions to occur at various points in the upload/cropping process to verify that things are kept clean.
- We are currently implicitly protecting ourselves from this vulnerability by executing a cropping process that removes the RCE vector. It would be nice to implement a service responsible for explicitly detecting and cleaning uploaded profile images so that in the event we ever stopped cropping, we would not become exposed again.

### DIFF
--- a/classes/OpenCFP/Domain/Services/ProfileImageProcessor.php
+++ b/classes/OpenCFP/Domain/Services/ProfileImageProcessor.php
@@ -57,7 +57,7 @@ class ProfileImageProcessor
 
         try {
             $file->move($this->publishDir, $tempFilename);
-
+            
             $speakerPhoto = Image::make($this->publishDir . '/' . $tempFilename);
 
             if ($speakerPhoto->height > $speakerPhoto->width) {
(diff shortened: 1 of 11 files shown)


mdwheele:
The migration submitted with this should be placed under high scrutiny. It makes destructive changes and while I have hand-tested all branches the migration covers, it deserves more than one set of eyes.

PR #197, opened 2015-05-19
Creation of OAuth2-backed JSON API

mdwheele:
This pull request continues previous work to flesh-out a service layer exposing a set of use-case implementations to be shared by both existing web pages and a newly formed JSON API. There is a significant amount of documentation submitted as part of this PR that describes the API and how it is expected to function mostly from a consumer's perspective. 

Most importantly, this PR marks a minimum viable product that would allow any third-party client to interact with an instance of OpenCFP in an automated way to do work on behalf of OpenCFP users. Immediately, [Symposium](https://github.com/tightenco/symposium) aims to offer speakers a means of collecting abstracts and other materials behind a centralized profile of sorts. They wish to be able to submit those talks to instances of OpenCFP on behalf of the speaker as part of a "one-button send it all" type of feature. This is now possible.

Authentication to the service is provided by an [OAuth2 Authorization/Resource Server](https://github.com/thephpleague/oauth2-server) implementation. This functions as any other OAuth2 service is expected to and is further documented by a usage scenario in the submitted documentation. The only "non-standard" thing about this OAuth2 implementation is that we do not require a client developer to create an account on instances of OpenCFP in order to register Client Applications. Client Application registration is done in accordance with a partial implementation of the [OAuth 2.0 Core Dynamic Client Registration](https://tools.ietf.org/html/draft-richer-oauth-dyn-reg-core-00) RFC draft. There are some caveats to this described in #195 

I'm at `#phptek` this year if folks want to discuss this at all before merging. 

**Odds and Ends**

- HTML Purifier is running as a middleware that cleans `query` and `request` parts of every `Request`. This is implemented as part of the routing layer.
- Controller factories are used to simplify how we were achieving enforcing SSL on all endpoints. Now, a single directive is applied to each applicable "route group" if configuration forces SSL usage. This is also how we're able to specify group-specific middleware.
- A middleware is defined to overwrite requests with `Content-Type: application/json`; registering JSON payload as part of the request so that consumers will not need to decode payload every time.

**Future Work**

- [ ] Talk submission validation errors should be reported all-at-once. Work has been done to support this.
- [x] Organizers should have the option of disabling the API through configuration. This should prevent the registration of not-only API endpoints, but also OAuth authorization endpoints.
- [ ] I want to work on how service providers are used to help better organize endpoint registration, controller registration, service binding.
- [ ] Continue to swap web controller implementations to use service layer. This will also come with an increase in test coverage and the tests should start to look a little easier to digest.
- [ ] Parts of the API need improved test coverage as well as better factoring out.
- [ ] It'd be nice to have an endpoint dedicated to serving API status. Is it on? Is it off? Does it only allow certain API actions (future)? In this way, a client could branch on how to handle talk submission if the API is turned off. Perhaps it just redirects them to the CFP site after putting talk contents in clipboard and we support some pasting of that encoded content? WHO KNOWS!? Magic!
- [ ] Will need a method for Client Application to update its Client Metadata in a next iteration, specifically redirect URIs.


mdwheele:
Hmm... getting inconsistent results locally from test suite versus what's running on Travis as far as the Behat tests go. Looking into it and will be pushing more soon.

chartjes:
The more I think about it, the more I would like to see those Behat tests removed because I think we are better off on wanting solid unit tests and great unit test coverage.

mdwheele:
I agree 100% here. The original value in those tests were to provide e2e coverage for the big service layer refactor. They are actually a misuse of Behat in the way they are formed. I think focusing on solid unit tests as we move forward would be a better route. Related to other comment I made on "Sentry/IdentityProvider". Can discuss more in person as a talking point. It'd be nice to set some forward direction on next "todos"

mdwheele:
Going to fix merge conflict and resolve whitespace issues.

PR #198, opened 2015-05-21
Run php-cs-fixer on classes and tests for code standards.

mdwheele:
This will likely break any open pull requests as it dramatically changes lines; causing merge conflicts. Probably a good idea to only merge when nothing is open.


chartjes:
Thanks @mdwheele

PR #199, opened 2015-05-21
Added in test for getAdminFavorites method in the Talk mapper

chartjes:
CodeClimate indicated the method was missing any coverage at all for that method


mdwheele:
Awesome! Thanks @chartjes!

PR #200, opened 2015-05-21
Add PHPUnit Speedtrap to detect slow tests while in development.

mdwheele:
This adds a listener for PHPUnit that detects tests that run over some threshold. I have set it to 300ms as that seemed to be a healthy start. There **are** some tests (3-4) that run around ~250ms so maybe bumping down is good? In addition, you can override this threshold for specific test methods. See more docs at https://github.com/johnkary/phpunit-speedtrap.

Build will not fail if it detects heavy tests, it's just to help improve when there is time. 

Example output:

``` php
PHPUnit 4.6.6 by Sebastian Bergmann and contributors.

Configuration read from /var/www/html/opencfp/phpunit.xml

...............................................................  63 / 108 ( 58%)
.............................................

You should really fix these slow tests (>300ms)...
 1. 828ms to run ForgotControllerTest:indexDisplaysCorrectForm
 2. 499ms to run OpenCFP\ApplicationTest:it_should_run_and_have_output


Time: 4.23 seconds, Memory: 92.50Mb

OK (108 tests, 167 assertions)
```

This look like something that'd be valuable?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,11 +42,8 @@
         "phpunit/phpunit": "3.7.*",
         "phpunit/dbunit": "1.3.*@dev",
         "mockery/mockery": "dev-master@dev",
-        "behat/behat": "~3.0,>=3.0.5",
-        "behat/mink-extension": "~2.0@dev",
-        "behat/mink-selenium2-driver": "*",
-        "behat/mink-goutte-driver": "*",
         "fzaninotto/faker": "1.2.*",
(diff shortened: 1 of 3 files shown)


mdwheele:
Also removes Behat dev dependencies and an updated `composer.lock` file.

chartjes:
Thanks @mdwheele

PR #203, opened 2015-05-22
Add Vagrant support for Development

mheap:
This is an initial pass at a Vagrant/Ansible powered dev environment. You can `vagrant up` and start hacking. It could be broken down into roles and parameterised more, but for now it works for me

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,10 @@ to your event. However, anything you are willing to push back should be updated 
 keep the master branch generic for future event organizers that choose to use the system. You would then be able to
 merge master to your private branch and get updates when desired!
 
+### Vagrant
+
+For those of you that want to dive in and start contributing quickly, we provide a Vagrant box to get you started. You'll need [Vagrant](https://www.vagrantup.com/) and [Ansible](http://www.ansible.com/home) installed. Once you have those, run `vagrant up` and get hacking!
+
 <a name="requirements" />
 ## Requirements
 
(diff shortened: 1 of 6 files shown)


chartjes:
Hey @mheap, we *used* to have a Vagrant/Puppet setup but we got rid of it because nobody was maintaining it. Let me think a while on if I want this as an option again.

mheap:
No worries, I built this mainly for myself and thought I'd contribute it back. I'm also happy to maintain it going forward.

If you want to remove the dependency on having Ansible installed locally, I can move that requirement into the VM to make it closer to the Vagrant/Puppet way of doing things.

mdwheele:
Here's another option that will is maintained by Laravel: http://www.joeferguson.me/install-homestead-into-your-project. Joe just added support for including homestead as a dev-dependency for exactly this use-case; an open-source project that wants a common baseline development environment. We don't run into a lot of problems here due to environmental differences, so we haven't had as much of a need for it I guess.

My primary motivations would be:

* I either want chosen tech to be something that anybody can quickly pick up and maintain. (no craaaaazy puppet modules, super complicated ansible playbooks, etc.)
* I'd rather not maintain it at all! :smile: I need whatever is chosen to put the focus on OpenCFP immediately. Not something I have to fight with before I get to work on OpenCFP.
* Having a recommended development environment would help tremendously at hackathons; especially for newer contributors.

I think that if we want to include a default development environment, we should look into what Joe's done. Looks pretty solid.

chartjes:
Providing a dev environment is out-of-scope for the project in my opinion, closing the PR

Issue #202, opened 2015-05-22
Group service registrations according to purpose. 


mdwheele:
Currently, there are service providers for days and things can be confusing to deal with. Also there are ordering problems where by things can be registered in sequences that will actually cause the test suite to fail. 

I would like to reorganize service providers according to more semantic roles as follows:

1. Route / HTTP gateways should be registered in gateway specific providers (web, API, OAuth) along with any applicable "controllers as services". This will help in being able to turn on / off whole blocks of the application as well as help locate. 
2. Application services and some dependencies should be grouped together; providing service to gateways. 
3. Infrastructural services / integrations with third party code should be grouped as such; vendor services. These might be bridges or implementations of domain services etc. 

This collapses things down to a few clearly intentioned service providers. I have work towards this I will be submitting after I get off train.

Issue #201, opened 2015-05-22
Improved error handling covering web, API and OAuth2 end points


mdwheele:
I have an implementation of an error handler that does a few things for us:

1. Helps start reducing the ordering problem that exists in current service providers. 
2. Responds appropriately when I caught exceptions occur. Currently an uncaught exception in the API renders a twig template. This is unexpected as a consumer of an API. 

This is foundational to continued work simplifying / consolidating the boot process. 

I'm on a train at the moment but will be submitting a PR at some point soon. Just need the webs.

PR #205, opened 2015-05-23
Improve error handling and providers

mdwheele:
This work approaches a solution to #201 and #202. 

- Error handling is centralized into the `Application` rather than hidden away as part of the `TwigServiceProvider`.
- Status codes from `OAuthException` are now reflected in all cases through conditional check for exception type.
- Status codes from exception implementations of `HttpExceptionInterface` are now used when rendering `JsonResponse` on-error. (Note: Our exceptions are not currently implementing this yet. We have some work to do in organizing an exception hierarchy for consistency)
- There are now three endpoint gateways for the API, OAuth and Web endpoints. I suggest that our current `Http` namespace begin to move towards that structure as a measure to clearly document entry-points to service layer.
- Rather than pass around a big closure duplicated across gateways, `RequestCleaner` was introduced as an invokable that wraps HTMLPurifier for the purposes of recursively cleaning the `Request` object.

Miscellaneous:

- Adjust some tests that relied on Twig environment variables being set. Because these are set in a before filter, the application must be `run` to execute callbacks. These are wrapped in an output buffer so that output is not rendered in middle of test-suite running. This whole problem goes away as we move controllers to being services with explicit dependencies as we will not need to boot the entire application.
- Whitespace fix for PSR-2


mdwheele:
Just more information; this work is done in preparation building a PHP client for the API that third-parties can take advantage of. This also helps shield them from changes to what honestly is an infant API. Specifically, this is formalizing error handling at the boundary of HTTP. I want to take this even a step further to formalize error handling at service boundaries. For example, current Sentry and OAuth identity providers *CAN* throw different sets of exceptions. This should be fixed in the future as the exceptions a unit throws is part of its public API and since they implement the same interface, that causes confusion.

chartjes:
Thanks @mdwheele

Issue #208, opened 2015-06-09
Command cache:clear missing env option


bcremer:
Given `config/development.yml` does not exist, only `config/production.yml` it is not possible to run the `cache:clear` command because it lacks the `env` option:

```
$ ./bin/opencfp cache:clear 
PHP Fatal error:  Uncaught exception 'InvalidArgumentException' with message 'The config file '/home/bcremer/www/opencfp/config/development.yml' does not exist.' 
```

```
$ ./bin/opencfp cache:clear --env=production
                                     
  [RuntimeException]                  
  The "--env" option does not exist.  
```

As a workaround one can use the environment variable prior executing the command:
```
CFP_ENV=production ./bin/opencfp cache:clear
Clearing all caches...
  ✓ /home/bcremer/www/opencfp/cache/htmlpurifier
  ✓ /home/bcremer/www/opencfp/cache/twig
Done!
```

bcremer:
Fixed in PR #209

PR #209, opened 2015-06-09
Add global --env option. Fix issue #208.

bcremer:
Fix for issue #208.


bcremer:
The `env` option is evaluated globally in https://github.com/opencfp/opencfp/blob/master/bin/opencfp#L15 but is not propagated to all commands.

Fix inspired by:
https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/FrameworkBundle/Console/Application.php#L46

chartjes:
Thanks @bcremer

PR #210, opened 2015-06-16
Turn off https default for development environment

zymsys:
While I agree that https is a sensible default for production, I think that normal http makes more sense in development environments. Feel free to reject this PR if you disagree.

### DIFF
--- a/config/development.dist.yml
+++ b/config/development.dist.yml
@@ -8,7 +8,7 @@ application:
   airport: MIA
   arrival: 2014-10-26
   departure: 2014-10-31
-  secure_ssl: true
+  secure_ssl: false
 
 api:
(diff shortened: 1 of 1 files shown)


mdwheele:
I don't really have an opinion one way or another on this. I run a self-signed cert locally just to not have to worry with things like this. The only caveat here would be that if the work in #211 is merged, the default environment is development. If someone *only* copies `.htaccess.dist` into it's production location and neglects to specify the production environment, by default we wouldn't be enforcing SSL.

All that said... there's a lot of FUD / "what-ifs" there that could be just as easily cleared up with documentation. I think it's a wash either way.

chartjes:
Thanks @zymsys

Issue #193, opened 2015-05-10
.htaccess in repo


zymsys:
Could we reconsider having the .htaccess files in the repo? I would prefer supplying a distribution .htaccess file with a different name and adding something to the installation documentation about copying that suggested version and modifying it if needed.

Also, is there any reason for .htaccess files in both the root folder and the web folder? Can we pick one location and merge them?

I'd be happy to submit a PR for this if the idea is acceptable. If not I can just disallow htaccess in my apache config and put the required setup there - but maybe I'm not the only one who could face this problem.

mdwheele:
Note also that the file in web is missing APP_ENV directive.

mdwheele:
> Could we reconsider having the .htaccess files in the repo? I would prefer supplying a distribution .htaccess file with a different name and adding something to the installation documentation about copying that suggested version and modifying it if needed.

This makes sense if it buys anything for user. @chartjes will have to comment on what he feels the "installation experience" should be for a new user as he has the best grasp on who is using this and how they're using it. Theoretically, it makes sense though.

> Also, is there any reason for .htaccess files in both the root folder and the web folder? Can we pick one location and merge them?

Many shared hosts disallow the ability to specify Apache/Nginx document root path. Specifying document root at the `web` directory is safer than setting the document root at the root of the project. Both methodologies require `mod_rewrite`, which may also be disabled by particular shared hosts. So perhaps the argument for shared hosts is moot.

I believe it is simply a trade-off for **ease-of-use / installation** (for perhaps a developer that doesn't have much web server experience) at **the cost of** perhaps applying the application to a particular infrastructure. Chris should comment, but I think that a line of work that did anything with this should obviously make it dead simple as far as documentation. This type of thing would be pretty awesome to have as part of an automated installation process as well. (unrelated perhaps)

pmeth:
My 2 cents:
1. project root should not be accessible so no need for .htaccess there.  @dustin I understand your point about different hosting companies, but we should not be catering to bad hosting companies, we should be encouraging users to leave them.
2. APP_ENV does not belong in .htaccess.  This should be configured by the server
3. The current .htaccess in /web is part of the front controller and is essential to the functioning of the application.  Thus it belongs in the project.  Someone should not have to rename it to install.  There is nothing in there that is specific to the environment.
4. I think we should consider using the Symfony standard .htaccess file (and probably renaming our index.php to app.php).  It accounts for a few more situations and is one less thing for us to worry about.

[...]

PR #211, opened 2015-06-16
Merge htaccess files into one suggested default

zymsys:
As discussed in #193 this PR merges the two .htaccess files into one .dist file and updates the README.md to instruct users to deploy this configuration as needed.


mdwheele:
@zymsys could you do the following:

- Rename `htaccess.dist` to `.htaccess.dist`
- Add `web/.htaccess` to `.gitignore` since it is now considered instance-specific configuration
- Update docs to reference new filename for dist.

Also, this should kick another build off. Looks like Travis failed because of a dropped connection from one of the VMs that were spun up to test php54. Looks good to me, otherwise.

chartjes:
Thanks @zymsys

PR #212, opened 2015-06-19
Make it easier to use the PHP built-in web server.

simensen:
After looking at the requirements for Apache in the htaccess file, we discovered that it would be pretty easy to just run with PHP's built-in web server for development.


mdwheele:
Looks good! Thanks @simensen.

PR #213, opened 2015-06-24
Uploading user photo and form error cause internal server error

andrewshell:
The variable `preview_photo` is only generated in ProfileController and not in SignupController, but they both use forms/_user.twig

If you upload a user photo during signup and have a form error (like invalid password) it fails because `preview_photo` doesn't exist.

### DIFF
--- a/templates/forms/_user.twig
+++ b/templates/forms/_user.twig
@@ -91,7 +91,7 @@
         <div class="col-sm-10">
             <input type="file" id="form-user-photo" name="speaker_photo" class="form-control" placeholder="Photo at least 300px wide by 300px high" accept="image/jpg, image/jpeg, image/png">
             <p class="help-block">Photo should be square, and must be at least 300px wide by 300px high. JPG and PNG are the only accepted formats.</p>
-            {% if speaker_photo is defined and speaker_photo is not empty %}
+            {% if preview_photo is defined and preview_photo is not empty %}
                 <img src="{{ preview_photo }}" class="profile-photo">
             {% endif %}
         </div>


mdwheele:
@andrewshell any chance we could get a test that reproduces this along with the patch? As it's an uncaught error, would like to be able to guard against regressions. If you don't have time, that's fine. I can send something along before merge.

andrewshell:
I'll see if I can figure out how to test it. The error is generated by Twig.

mdwheele:
If not that's fine. I can wrangle something together. Thanks for the work!

andrewshell:
As I'm attempting to write a test, I realized that preview_image isn't being defined ever for that form, so I'll try to fix (and test) that too.

andrewshell:
I've tried a couple of different things and I can't figure out how to get twig injected properly with all its extensions and such in order to fail because of a missing variable.

mdwheele:
Writing up a test for this now. This will actually fail when updating your profile as well for the same reasons you've described. However, I believe the problem is that `preview_photo` is an older name for `speaker_photo` and this looks to be an oversight when a template change was made in a87169ad (see that previously `speaker_photo` was used and `preview_photo` was subbed in. Sorry I didn't look into this a little deeper upfront. I've been mobile for quite a bit lately.

andrewshell:
In ProfileController it has speaker_photo and preview_photo.  The only difference is that preview_photo is prefixed with /uploads/

In SignupController speaker_photo isn't a string, it's the response from $req->files->get('speaker_photo'); so you don't want to change references to preview_photo to speaker_photo.

When the form template is displayed after an error in SignupController, the image hasn't been processed yet, so there is no public path that could be passed to the template to show the image.

chartjes:
So, do we have a test to verify this? I can't merge anything in without a passing test.

andrewshell:
I was not able to get a test working with all the twig stuff.  I'm not sure if anyone else is working on getting a test created.

mdwheele:
I started one and will finish and get it in tomorrow. End to end tests are not the easiest to write for this case for sure.

PR #216, opened 2015-07-02
Fix defaults for users table.

simensen:
When using more strict MySQL settings it is not possible to add records without specifying values for NOT NULL columns unless a default is specified. Phinx creates NOT NULL columns by default so many fields were giving us problems when trying to register a user without giving a proper value for many fields that the signup controller wasn't asking about.

This fixes the problem by altering the database schema such that the fields in question will either allow null values or specify a default value in the case that NULL is used or no value is provided.

# Important Note

For people with strict MySQL settings, this migration will most likely not be reversible as columns will end up with NULL values and reversing this migration results in making those columns NOT NULL.

People with less-strict MySQL settings will not have any problems with this, though.

### DIFF
--- a/migrations/20150702132854_fix_defaults_for_users_table.php
+++ b/migrations/20150702132854_fix_defaults_for_users_table.php
@@ -1,0 +1,78 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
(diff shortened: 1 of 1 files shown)


davedevelopment:
:+1:

simensen:
Thanks, @chartjes!

PR #217, opened 2015-07-03
Update test instructions and build

yitznewton:
Adding things to the README which I needed to do in order to run the full test suite

* Create test database
* Copy some scripts from .travis.yml to the README

Also, this moves `phpunit.xml` to `phpunit.xml.dist`, and gitignores `phpunit.xml`. Doing that allows local customization of the phpunit configuration in `phpunit.xml`. The explicit specification of the config file in .travis.yml is also not needed

More info on the config file names and fallback is [here](https://phpunit.de/manual/current/en/textui.html#textui.clioptions); unfortunately there is no link directly to that option


mdwheele:
Thanks @yitznewton!

PR #218, opened 2015-07-06
Enhancement: Switch to PSR-4 autoloading

localheinz:
This PR

* [x] switches from PSR-0 to PSR-4 autoloading


mdwheele:
Thanks! @chartjes wanna make a call here? Only thing to think about is merging or closing other PRs before this because conflicts from file moves. Since this PR is pretty lightweight it shouldn't be that bad to merge after a while.

chartjes:
Thanks @localheinz

PR #219, opened 2015-07-06
Fix: Remove import for non-existent class

localheinz:
This PR

* [x] removes an import for a non-existent class `OpenCFP\ProfileImageProcessor`
* [x] fixes an inline comment referring to a non-existent class `OpenCFP\ProfileImageProcessor`

:bulb: `OpenCFP\ProfileImageProcessor` doesn't exist, but `OpenCFP\Domain\Services\ProfileImageProcessor` does.


mdwheele:
Thanks @localheinz!

PR #220, opened 2015-07-06
Fix: Speaker::findProfile() doesn't accept any arguments

localheinz:
This PR

* [x] removes arguments from calls to `Speaker::findProfile()` as the method doesn't accept any
* [x] removes a fat finger


mdwheele:
Looks good! Looks like an oversight from recent API addition. Waiting for tests to come back green and merging. Thanks @localheinz!

localheinz:
@mdwheele 

You're welcome!

PR #221, opened 2015-07-06
Fix: Reduce visibility from public to protected

localheinz:
This PR

* [x] reduces the visibility of `setup()` and `tearDown()` from `public` to `protected`, as this is the visibility used in `PHPUnit_Framework_Testcase`


mdwheele:
You're a machine! Thanks @localheinz.

PR #222, opened 2015-07-06
Fix: Move YamlConfigProvider into appropriate location

localheinz:
This PR

* [x] moves `OpenCFP\Provider\YamlConfigDriver` into the appropriate location

Follows #218.


localheinz:
@chartjes 

This should fix the breaking build as reported for https://travis-ci.org/opencfp/opencfp/jobs/69736790!

PR #224, opened 2015-07-06
Fix: Cache directory for dependencies installed with Composer

localheinz:
This PR

* [x] fixes the directory which should be used for caching dependencies installed with Composer on Travis


localheinz:
Ha, builds fails, of course, because it still needs #222.

PR #225, opened 2015-07-07
Fix: Clean up Travis configuration and build script

localheinz:
This PR

* [x] updates the Composer binary in the `before_install` section
* [x] installs dependencies with Composer in the `install` section
* [x] defines the environment variable `TRAVIS_DB` in the `env` section
* [x] creates database in `before_script` section (and doesn't worry about dropping it)
* [x] runs database migrations in `before_script` section (and doesn't worry about renaming the migration configuration file)
* [x] runs `phpunit` in `script` section
* [x] uses `phpunit` as installed with development dependencies (rather than the version provided on Travis)
* [x] removes unused scripts `tools/travis/build.sh` and `tools/travis/reset-database.sh`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,14 +15,16 @@ cache:
 addons:
   code_climate:
     repo_token: 9deb249b01a414d979959cfd05a4c351b19a5959858653b20276454d4189edc3
-    
+
 before_install:
   - composer self-update
-  
+
+install:
+  - composer install
+
 before_script:
   - sh tools/travis/setup-mail.sh
   - cp phinx.yml.dist phinx.yml
-  - composer install
 
 script:
   - sh tools/travis/build.sh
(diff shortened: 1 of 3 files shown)


localheinz:
@chartjes 

With the cache kicking in now, builds are a lot faster!

localheinz:
@chartjes @mdwheele 

We can speed them up a lot more by only collection code coverage 

* during a merge to `master` **and**
* during a build on one PHP version

localheinz:
@chartjes @mdwheele 

Not sure about deleting the shell scripts, are they used for the actual deployment, too?

chartjes:
I think deleting the shell scripts are fine because they are only used by Travis. Right now the deploy for the TrueNorthPHP install is done manually by me via git pull since I don't have any admin privileges on that box to set up rsync deploy etc

localheinz:
@chartjes 

:+1:

chartjes:
Thanks @localheinz

PR #226, opened 2015-07-13
Enhancement: Run composer --validate on Travis

localheinz:
This PR

* [x] runs `composer validate --no-check-publish` in `before_install` section

:information_desk_person: Following https://github.com/composer/composer/pull/4219, this will break the build if `composer.lock` is not up to date.


chartjes:
Vielen dank @localheinz

PR #228, opened 2015-07-19
Use ambiguous success message when reset-password works

yitznewton:
In 50d1ed25, @chartjes changed the no-matching-user message to be
ambiguous, but we are still using a positive message when the email does
exist, so we haven't closed the loop.

This PR changes to use the same message for both cases: attempted reset
of an unknown email, and reset of a known email.


chartjes:
Thanks Yitz

PR #230, opened 2015-07-24
Update default dev config to match phinx

yitznewton:
See https://github.com/opencfp/opencfp/blob/master/phinx.yml.dist#L14-L19


mdwheele:
Thanks @yitznewton!

PR #232, opened 2015-07-28
Refactor and add test coverage for sending reset emails

yitznewton:
As discussed with @chartjes on Twitter a few weeks ago, here's the result of looking for some code with no test coverage. (Meetup video [here!](livestream.com/accounts/3104006/events/4213735))

This involved extracting the email-sending code into its own class, and wrapping that via a service provider.

As a next step, I would add controller tests which mock out the `reset_emailer` and assert that it's called with `send()`, or just jump to abstracting this into a "use case" where the controller knows less about what's going on, but I have my doubts about getting to that.

As far as testing the refactor, I spun up a Vagrant instance locally with Postfix, and I still receive the reset emails in my GMail account, and they look right.

Let me know if this is :astonished: with your ways of doing things, or if there are established manual test cases that I should do, or the like.


classes/Provider/ResetEmailerServiceProvider.php
```diff
@@ -0,0 +1,32 @@
...
+use OpenCFP\Domain\Services\ResetEmailer;
+use Silex\Application;
+use Silex\ServiceProviderInterface;
+
+class ResetEmailerServiceProvider implements ServiceProviderInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function register(Application $app)
```
yitznewton: This is a little sneaky... we're calling `$app->config()` below, which is from `\OpenCFP\Application` not `\Silex\Application` as in the `use`... but the interface requires `\Silex\Application` and PHP doesn't complain, so... ?

tests/OpenCFP/Domain/Services/ResetEmailerTest.php
```diff
@@ -0,0 +1,51 @@
...
+    /** @test */
+    public function it_sends_the_expected_email()
+    {
+        $this->mailer->send($this->user_id, $this->user_email, $this->reset_code);
+    }
+
+    private function validateEmail()
+    {
+        return function ($message) {
+            return $message->getTo() === array($this->user_email => null);
```
yitznewton: A basic mock to sanity-check the Swift message being sent; we could make more assertions against the message, but because we're using a Twig template, it would quickly become a hairy test

classes/Domain/Services/ResetEmailer.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace OpenCFP\Domain\Services;
+
+class ResetEmailer
```
mdwheele: Could we make this a mailing abstraction, in general? Thought would be we'd start to migrate other code to depend on this abstraction rather than explicitly using `Swift_Mailer`.

Would also mean that (unless there is specific logic going on here) this would become merely a contract that would be implemented in `OpenCFP\Infrastructure\Mail` (or similar). Check out the `IdentityProvider` and implementations in `OpenCFP\Infrastructure\Auth` for an example of this style.

tests/OpenCFP/Domain/Services/ResetEmailerTest.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->mailer = \Mockery::mock('mailer')->shouldReceive('send')->once()
+            ->with(\Mockery::on($this->validateEmail()))->getMock();
+
+        $this->template = \Mockery::mock('template')->shouldIgnoreMissing();
+        $this->config_email = 'admin@example.com';
+        $this->config_title = 'Reset';
+
+        $this->user_email = 'user@example.com';
+        $this->user_id = 123;
+        $this->reset_code = '987abc';
```
yitznewton: I put `user_email` up here so that we can easily grab it in `validateEmail()`; we don't really need the other two, unless we decide to add those assertions against the created Swift message

classes/Domain/Services/ResetEmailer.php
```diff
@@ -0,0 +1,64 @@
...
+
+    public function __construct($mailer, $template, $configEmail, $configTitle)
+    {
+        $this->mailer = $mailer;
+        $this->template = $template;
+        $this->config_email = $configEmail;
+        $this->config_title = $configTitle;
+    }
+
+    public function send($user_id, $email, $reset_code)
```
mdwheele: Prefer `camelCase` parameter names.

mdwheele:
Thanks @yitznewton for the work and I hope the meetup went well! Lemme know what you think about "mailing abstraction" above. If you don't have time to bite it off, that's fine. This is a step in the right direction at any rate.

classes/Domain/Services/ResetEmailer.php
yitznewton: Most of what this class is doing is actually cobbling together the message using variables and Twig templates... it also does call the Swift mailer service. It feels like this would end up being a layer of abstraction on top of Swift, both message and mailer service... would that actually gain us anything?

yitznewton: IOW, this would be useful if we're required to switch out Swift for something else, but would we ever?

yitznewton:
Oops, I found a horrible bug, hang on....

yitznewton:
OK :sweat_smile:

classes/Domain/Services/ResetEmailer.php
mdwheele: Wouldn't gain us anything as far as switching to another mailing implementation. Hrm... can't find words over text so may just drop it and think some more. I'll try to explain better.

This is 100% the right abstraction for the job of "sending an account reset email". If I need to send a mail, in general (say as a result of some action in the API), I might not want to fulfill a dependency on Twig to send a plain-text email that a client application was registered. I probably wouldn't use this service and that's okay.

What *would* happen is that I would have to (at some point) grab references to `Swift_Mailer` in terms of the next best Service to do the "job of today". I may use that work as an opportunity to refactor the duplication I'm anticipating today.

However, I guess what I'm getting at more than anything is that the job of sending a account password reset notification is prompted by user action and could be orchestrated (with the right abstractions) as a controller action without a need to implement an service explicitly purposed for that one job.

Either way, it doesn't really matter so much I suppose. Just a point of discussion. I'd like to eventually move more and more towards something like:

```php
class AccountController extends Controller 
{
    public function processPasswordResetRequest(Request $request)
    {
        // Load up the user.
        $user = $this->identityProvider->getCurrentUser();
        $event = $user->resetPassword();

        $this->dispatcher->send($event);
    }
}

class AccountNotifier 
{
    public function whenPasswordWasReset(PasswordWasReset $event)
    {
        // Do the needful with whatever mailing abstraction / message construction.
        $this->mailer->send($event->getUserEmail(), $event->getMessage());  // Or similar...
        
        // ...
    }
}
```

All that said. I fully understand that this PR also represents a refactor for the purposes of demonstrating introducing tests into a legacy project, so I can appreciate that. We can continue down the path you've cut so far! :smile: Thanks!

yitznewton: Ooh, eventy. Nice!

mdwheele: From the API-perspective, I think moving towards (and we're 120% outside the scope of this work now) that type of model offers some pretty interesting integration opportunities for connected systems. Imagine System A being able to ask an OpenCFP instance, "Hey! What are all the things that `yitznewton` did since yesterday?" ... "Hey! I know I know... I was down for a few weeks, could you be an awesome OpenCFP API and let me know all the account registrations that happened since 3 weeks ago?" etc.

mdwheele: Let's just leave it as-is for now. I'm fine to merge this otherwise, though I want to give it a little bit deeper look when I get home. More tests and tests passing is always a good thing.

yitznewton: Ah, sounds like [event sourcing](http://martinfowler.com/eaaDev/EventSourcing.html)? Would that require a substantial data refactor? #gleefullyofftopic

mdwheele: You're on the right track. For this project, you might just be building a domain history of arbitrary use-cases and making it easily addressable. Sourcing entities from that history isn't necessarily required. You get the best of both worlds for as long as you want.

chartjes:
Thank you again @yitznewton @mdwheele

yitznewton:
@chartjes thank you again for the suggestion on Twitter; this is an example of a great PHP project! Glad I've been able to contribute some

Issue #234, opened 2015-07-29
Adding code style/quality tools


yitznewton:
I propose we add style and quality tools to the Travis build, which would both communicate the project's code requirements to developers automatically, and alleviate the need for reviewers to comment on style violations. My recommendations:

* http://phpmd.org/
* https://github.com/FriendsOfPHP/PHP-CS-Fixer

I'll be happy to set those up

yitznewton:
There's also codesniffer; at my last PHP shop we used all three of these. I'm not sure what PHP_CS adds over phpmd

mdwheele:
PHP_CodeSniffer detects violations of specified code style. PHP-CS-Fixer simply builds on top of that to auto-fix those violations for you. Not sure if it would ever make sense to run both since PHPCSF covers that entire use-case (identification (through `--dry-run`) of violations as well as fixing). PHPMD isn't as much concerned with code-style as it is an attempt at static analysis of code. It's going to look for duplication of code. It's going to look at method cyclomatic complexity. Etc. It will also probably light up like Las Vegas when pointed at OpenCFP so it might feel odd for **someone** submitting a [simple fix](https://github.com/opencfp/opencfp/pull/233) to get notes about completely unrelated code "being inefficient". Food for thought. Maybe we just want to introduce code-style checks for now?

In addition, we also have (a not well publicized) Codeclimate registration at https://codeclimate.com/github/opencfp/opencfp that I think is really a better solution for the job. Rather, it's already covering the case of "inefficient code" etc etc. Through configuration, I believe we can get it to report status changes on PRs as a check as well. We mentioned something similar to them at php[tek] this year, but I lost track of where that went.

Questions for consideration:

* Will we have PHP CS Fixer doing dry-runs detecting code style violations or will we be somehow configuring it to actually make changes + commits to PRs. FWIW I think I'd prefer the former; having it simply communicate the violation.
* Will we have PR merges **require** that these code-style tests pass or do we make it less strict than that. 
* Rather than hitting every PR with `phpmd` what if we better publicized the Codeclimate registration and pointed folks there as a "low hanging fruit" or opportunity to contribute. We could also better track that over time. 

Take nothing above as assertive / conclusive. I'm in a rush and was just getting some ideas out! Definitely open for discussion.

yitznewton:
I'm not a fan of autocorrect in the background; I was trying out the tools on a different codebase, and it did break some things.

[...]

PR #235, opened 2015-07-29
Add code style tools

yitznewton:
Examples for #234

## PHPMD

Minimal fixes for some style issues, mostly turning off warnings. We can selectively & progressively turn on and fix things

## PHP-CS-Fixer

The basic PSR2-level fixes


yitznewton:
PHPMD rules deactivated:

* Superglobals
* Camel-case var name (contrary to OpenCFP style)
* Exit expression
* Short variable (`$id` violates)
* Unused formal parameter

yitznewton:
Take a look at the camel-case parameter name changes; not sure if there's any library magic happening with those parameters. In the controllers, I suppressed rather than changed.

.travis.yml
```diff
@@ -32,7 +32,10 @@ before_script:
   - vendor/bin/phinx --configuration=phinx.yml.dist migrate -e testing
 
 script:
-  - vendor/bin/phpunit --coverage-clover build/logs/clover.xml
+  - ./vendor/bin/phpunit --coverage-clover build/logs/clover.xml
+  - ./vendor/bin/php-cs-fixer fix classes/  --verbose --level=psr2 --dry-run
+  - ./vendor/bin/php-cs-fixer fix tests/  --verbose --level=psr2 --dry-run
```
localheinz: @yitznewton 

Instead of having two runs, you can also specify a configuration file in which you can

* configure a `Finder` to use for iterating over (directory to start in, exclusions, extensions, etc.)
* base fixer level
* additional fixers
* excluded fixers

localheinz:
@yitznewton 

I think it's a better idea to do it this way:

1. enable a tool / fixer on Travis (breaks the build)
2. apply the fixer (to fix the build)

What do you think?

composer.json
```diff
@@ -44,6 +44,8 @@
         "mockery/mockery": "1.0.*@dev",
         "fzaninotto/faker": "1.2.*",
         "codeclimate/php-test-reporter": "dev-master",
-        "johnkary/phpunit-speedtrap": "~1.0@dev"
+        "johnkary/phpunit-speedtrap": "~1.0@dev",
+        "fabpot/php-cs-fixer": "^1.10",
+        "phpmd/phpmd": "^2.2"
```
localheinz: @yitznewton 

Just as an FYI, I have contributed a small feature to `composer/composer`, the `--sort-packages` flag!

Not sure if you are as pedantic as I am, but I like it, it keeps platform requirements and dependencies sorted alphabetically:

```
$ composer require --dev --sort-packages fabpot/php-cs-fixer:^1.10
```

yitznewton:
@localheinz do you mean red-green style with Travis? :shrug: I did it that way locally, not sure how Travis makes it better?

composer.json
yitznewton: Heh that's cool, I also see it has auto-chosen the `^` versioner... what is best?

classes/Provider/Gateways/RequestCleaner.php
```diff
@@ -32,7 +32,8 @@ private function clean(array $data)
             if (is_array($value)) {
                 $sanitized[$key] = $this->clean($value);
             } else {
-                $sanitized[$key] = $this->purifier->purify($value);;
+                $sanitized[$key] = $this->purifier->purify($value);
+                ;
```
mdwheele: Can remove extra semi-colon here.

localheinz:
@yitznewton 

:+1:

I think it's similar to the red-green-refactor loop:

* add a build step that fails
* fix just enough to fix the build

I'm convinced the outcome is the same (trust you on this one)!

mdwheele:
> I think it's a better idea to do it this way:
> 
> enable a tool / fixer on Travis (breaks the build)
> apply the fixer (to fix the build)
> What do you think?

I'm a fan of this. Let it do its work and if it breaks the tests, we'll fix from there. The preference *is* for camel case parameters and we can always roll a commit back :smile:

composer.json
localheinz: @yitznewton 

From what I heard and read, the `^` operator works best for packages which you trust to have embraced semantic versioning, as for example

```
$ composer require --dev fabpot/php-cs-fixer:^1.9.3`
```

will always resolve to `<=2.0.0` and `>=1.9.3`, while

```
$ composer require --dev fabpot/php-cs-fixer:~1.9.3
```

will resolve to anything `>=1.9.3` and `<1.10` (if I understood it correctly).

classes/Provider/Gateways/RequestCleaner.php
localheinz: @mdwheele 

There's a fixer for that, too, see https://github.com/FriendsOfPHP/PHP-CS-Fixer!

mdwheele: Then lets hook it up! Bam! Progress!

mdwheele:
Hey @yitznewton! I stopped tracking this PR and no longer have a good idea where we are. Could you send along a short punch-list of remaining items to fulfill spirit of PR? Looks last latest commit was checking some rules were actually working. I'm happy to help finish up, just a little behind! :smile:

yitznewton:
I think bases on Andreas's recent additions, we can close this 😍

Issue #237, opened 2015-08-14
Admin section wishlist


pmeth:
Using admin section for first time this year, some things jump out at me as being missing.  Note I don't have time right now to look at any of the code to see why or fix any of these things, just putting here in case someone gets ambitious and wants to address these:
Menu:
- page you are on should have a "selected" class
- names should match the headings of the pages (eg: menu says "Talks", page says "Submitted Talks")

Dashboard:
- totals should update in real-time as I select/favourite talks

Admins:
- ability to add admins

Submitted Talks:
- should be able to sort asc/desc (right now only asc is available)
- sort column should be identified somehow (arrow would make most sense assuming the previous item is done
- submitted by should be sortable
- filters would be quite useful
- created on is not really important and should be removed to create more real estate
- add columns for level, desired, sponsor, airport, travel costs (could be a calculate button if not already known), speaker # of previous selections
- should have yes/no/maybe type selection instead of "selected" & "star"
- should have a totals section at top (yes talks, yes speakers, yes cost, no, maybe)
- needs an export button (excel, csv, whatever)
- delete option
- view all (ps, i dislike pagination in general)
- selections and favourites on a per-admin basis

Speakers:
- i feel like we need an option to yes/no/maybe speakers
- ability to sort/filter columns
- option to remove pagination (view all)

Review Talks:
- I don't know what this section does. It never seems to have any talks populated

chartjes:
Closing this because of @jakefolio's huge admin PR fixed a bunch of these issues, we can revisit individual ones after

PR #241, opened 2015-08-29
speakers list makes more sense alphabetized by first name

adamculp:
The speakers names in the admin speaker listing are displayed as full name (first last) format but are sorted on last name, causing some confusion when attempting to find the speakers in the list. I altered the order to be on firstname and it makes finding speakers much easier.


chartjes:
Thanks @adamculp

pmeth:
Note issue #237 has a wishlist item to be able to sort this column... in case you get ambitious ;)

PR #242, opened 2015-08-29
Enhancement: Run php-cs-fixer on Travis

localheinz:
This PR

* [x] requires `fabpot/php-cs-fixer`
* [x] adds a simple configuration, effectively checking PSR-2 compliance
* [x] performs a dry-run of `php-cs-fixer` on Travis to break the build
* [x] disables PSR-0 fixer (see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1329)
* [x] runs `php-cs-fixer` to fix issues

Follows #198.
Related to #234.
Related to #235.

PR #243, opened 2015-08-29
Fix: Remove unused imports

localheinz:
This PR

* [x] removes unused imports

:information_desk_person: Alternatively, one could add the `unused_use` fixer in #242.


chartjes:
thanks @localheinz

PR #244, opened 2015-08-29
Fix: Reduce visibility of instance property

localheinz:
This PR

* [x] reduces the visibility of an instance property to `protected`
* [x] fixes a docblock


chartjes:
Thanks @localheinz

PR #247, opened 2015-08-29
Fix: Remove unused parameter

localheinz:
This PR

* [x] removes an unused parameter from `AdminAccessTrait::userHasAccess()`

:person_with_pouting_face: The method never uses it, so no need to pass it in, either.


localheinz:
Closing in favor of #251, because there we got tests!

PR #249, opened 2015-08-30
Fix: Add coverage for ContainerAware

localheinz:
This PR

* [x] adds an `autoload-dev` section to `composer.json` so we can autoload test resources
* [x] adds coverage for `ContainerAware`, because it can break, too


chartjes:
Thanks @localheinz

PR #250, opened 2015-08-30
Fix: Add coverage for FlashableTrait

localheinz:
This PR

* [x] cherry-picks 749c0df so we can autoload test resources
* [x] asserts that the `FlashableTrait` works as it claims it does

Follows #249.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,6 +38,11 @@
     "autoload": {
         "psr-4": {"OpenCFP\\": "classes/"}
     },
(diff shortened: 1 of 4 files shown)


localheinz:
@chartjes 

Not sure if 60b784d should be in a separate PR?

Also, I think it makes sense to use the `Flashable` trait in all the controllers, maybe?

chartjes:
Thanks @localheinz

PR #251, opened 2015-08-30
Fix: Add coverage for AdminAccessTrait

localheinz:
This PR

* [x] cherry-picks 749c0df so we can autoload test resources
* [x] asserts that the `AdminAccessTrait` works as it claims it does
* [x] asserts that `__call()` never intercepts a method call to an existing method
* [x] removes useless `__call()`
* [x] asserts that method argument is never used
* [x] removes unused method argument
* [x] removes unused method argument from method calls

Follows #249.
Related to #247.


chartjes:
Thanks @localheinz

PR #253, opened 2015-09-02
Enhancement: Add Makefile with cs target

localheinz:
This PR

* [x] adds a `Makefile` with a `cs` target, so it's easier to fix issues locally

Follows #242.
### Usage

Run

```
$ make cs
```


chartjes:
Thanks @localheinz

PR #261, opened 2015-11-15
Create CONTRIBUTING.md documentation

mdwheele:
We didn't have one and after grabbing (and whiffing) for one in our project was surprised when it wasn't there.

I lifted this one from The PHP League's skeleton project and made slight modifications in light of some of the conventions I've experienced contributing here. Please correct me if wrong about the last point to "not close issues with GitHub commit-message magic". I'm recollecting us not doing that but if inaccurate, let's remove.


chartjes:
Thanks @mdwheele

PR #263, opened 2015-11-16
Missing query alias

jakefolio:
The missing alias, total_rating, was causing an uncaught exception. I'm assuming this snuck it from an old version of the application.

Also, I added the use of getId instead of just the id property for consistency sake.


chartjes:
Thanks @jakefolio

Issue #207, opened 2015-05-25
Removing flight/travel cost for online conferences


chrispecoraro:
I hacked up openCFP to hide the travel information from the speaker registration when the conference is online.
Is it worth it to do a pull request for this? I'd have to figure out how it work it into the settings.

mdwheele:
Could you send a link to implementation? I think it depends on what "hacked up" looks like. Ideally this would be done at the service level as part of some DTO returned to be exposed by both web and API end points.

chrispecoraro:
I hacked it up for my own needs, but what I would do is actually write it in a proper way and then send the pull request. For the API, I'm not sure if it matters since interface designers could just use whichever information that they want.

mdwheele:
Ah! I misread what this was was actually referring to. I thought it was directly related to the other issue you submitted re: anonymous abstract submissions. You're actually talking about implementation of variable account creation interface (front-end) display, correct?

If so, the fields related to flight/travel cost are already optional so having a configuration option for show/hiding fields on the registration (and all display) pages should "work", but it begs a larger question of handling variations in OpenCFP instances like this. If you have the free time to send a PR along, even if in raw form, it would be a good starting point for discussion. If not, I think this warrants more discussion than a simple boolean on/off? In other words, I see that type of implementation getting worked on sooner for something that handles more of the use-case. @chartjes has more experience with the types of modifications folks usually make to instances of OpenCFP. Perhaps he'll have some ideas on direction here.

That said, could you send along your idea in some rough form as part of this issue? Ignore what I said re: DTO+API. I misunderstood the issue as this seems purely front-end focused (show/hiding individual form elements)

chrispecoraro:
There's not just to show. All I did was comment out wherever "submitted by" shows the name. Since I know most of the submitters to the online conference, I didn't want that to bias the selection. After we pick the winning abstracts, we'll then show the names. Since it's an online conference, flight/travels don't matter.

chartjes:
@chrispecoraro Put together a PR and I'll take a look. Like @mdwheele mentioned we already have a configuration option for showing/hiding those fields and they are not considered required fields for validation purposes.

chrispecoraro:
I must have not read through things well enough then. How can I disable flights/travel related information?

chartjes:
Hrm, I don't see it in there either. Could've sworn that it was in there.

I would be happy to take a look at a PR that makes this possible

chartjes:
[...]

PR #265, opened 2015-11-20
Fixed issue #207

chartjes:
The request was to make it so that information about airport and hotel could be hidden, so I added a confg value (set to be false by default) if the conference is an online one.

### DIFF
--- a/config/development.dist.yml
+++ b/config/development.dist.yml
@@ -9,6 +9,7 @@ application:
   arrival: 2014-10-26
   departure: 2014-10-31
   secure_ssl: false
+  online_conference: false
 
 api:
   enabled: true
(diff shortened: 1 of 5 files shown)


mdwheele:
We need to put that variable in the distributable configs for other environments as well.

chartjes:
Okay, will add that to the other dist files

--
  Chris Hartjes
  chartjes@littlehart.net
 
 
 
On Thu, Nov 19, 2015, at 08:57 PM, Dustin Wheeler wrote:
> We need to put that variable in the distributable configs for other environments as well.


> —
> Reply to this email directly or view it on GitHub[1]. 


>  
 


Links:

  1. https://github.com/opencfp/opencfp/pull/265#issuecomment-158257297

PR #270, opened 2015-12-04
Added test coverage for admin talks controller

chartjes:
Part of the work I've been doing for my book.

@mdwheele Please review


mdwheele:
I'll be able to follow this up with examples if I don't make much sense right now. Still in the middle of moving things around so time is pressed. I believe this is an improvement over the current test for this controller, specifically in the way mocks are specified. We get clearer specification of the collaboration between actual objects in our domain rather than `stdClass` stubs to "fake it til' we make it".

Beyond some simple typos (I'll line comment em'), I think this is a great step in the right direction. We're moving from what I would have considered characterization tests into something that is more descriptive of behaviour of the domain of conference call for paper administration.

I'd like to see continued progress in describing the messaging patterns between objects in OpenCFP in terms of **why** those messages are being passed. These improved test examples show a clearer picture of the interactions between objects as part of a controller method executing, but I'd like to see language that describes the scenario / context under test.

[This test](https://github.com/opencfp/opencfp/blob/master/tests/OpenCFP/Application/SpeakersTest.php#L83-L91) is by no means perfect, but is somewhat in the direction I'm talking about. That said, I don't really like that my method name is so long, `trainIdentityProviderToReturnSampleSpeaker`, but it does read well and if I want to know *how* that happens, I can dive in. The test, at the highest level, describes context-first and hides coupling to implementation which is a necessary evil when it comes to legacy. We work to improve that over time.

https://github.com/opencfp/opencfp/blob/master/tests/OpenCFP/Application/SpeakersTest.php#L104-L116 is another good example, I think. It's very clear what the test's purpose is and it's easy to find the "arrange" / context for the particular scenario.

Sorry for wall of text. I think this is good work.

tests/OpenCFP/Http/Controller/Admin/TalksControllerTests.php
```diff
@@ -0,0 +1,184 @@
...
+use Symfony\Component\HttpFoundation\Session\Session;
+use Symfony\Component\HttpFoundation\Session\Storage\MockFileSessionStorage;
+use OpenCFP\Application;
+use OpenCFP\Environment;
+
+class AdminTalkControllerTest extends \PHPUnit_Framework_TestCase
+{
+    public $app;
+
+    public function setup()
```
localheinz: @chartjes 

Visibility of parent is `protected`, should the visibility be the same here?

tests/OpenCFP/Http/Controller/Admin/TalksControllerTests.php
```diff
@@ -0,0 +1,184 @@
...
+
+use Mockery as m;
+use Symfony\Component\HttpFoundation\Session\Session;
+use Symfony\Component\HttpFoundation\Session\Storage\MockFileSessionStorage;
+use OpenCFP\Application;
+use OpenCFP\Environment;
+
+class AdminTalkControllerTest extends \PHPUnit_Framework_TestCase
+{
+    public $app;
```
localheinz: @chartjes 

Would `private` visibility suffice here?

tests/OpenCFP/Http/Controller/Admin/TalksControllerTests.php
```diff
@@ -0,0 +1,184 @@
...
+            );
+
+        ob_start();
+        $this->app->run();
+        ob_end_clean();
+
+        $controller = new \OpenCFP\Http\Controller\Admin\TalksController();
+        $controller->setApplication($this->app);
+        $response = $controller->indexAction($req);
+        $this->assertContains('Test Title', (string) $response);
```
localheinz: @chartjes 

Would it make sense to introduce an explaining variable to avoid repeating its value, and possibly, using `fzaninotto/faker` for generating fake data?

tests/OpenCFP/Http/Controller/Admin/TalksControllerTests.php
```diff
@@ -0,0 +1,184 @@
...
+{
+    public $app;
+
+    public function setup()
+    {
+        // Create our Application object
+        $this->app = new Application(BASE_PATH, Environment::testing());
+
+        // Create a test double for our User entity
+        $user = m::mock('OpenCFP\Domain\Entity\User');
```
localheinz: @chartjes 

`README.md` says PHP5.5+ is required, `composer.json` requires `php:>=5.4`, and `.travis.yml` runs builds against at least PHP5.5, so maybe the tests could benefit from using the `class` keyword? Makes it a lot easier to find usages, and also, to refactor. 

What do you think?

classes/Http/Controller/Admin/TalksController.php
```diff
@@ -67,6 +67,7 @@ public function indexAction(Request $req)
             'order_by' => $req->get('order_by'),
         ];
 
+
```
localheinz: @chartjes 

Extra empty line is extra, isn't it?

tests/OpenCFP/Http/Controller/Admin/TalksControllerTests.php
mdwheele: I'm getting ready to merge this in. I've created https://github.com/opencfp/opencfp/issues/271 as an effort to make a sweep through on stuff like this.

tests/OpenCFP/Http/Controller/Admin/TalksControllerTests.php
mdwheele: Gonna leave it for now. This is how it's done throughout the codebase so it'd make more sense to make an effort across the codebase to do this if we want to. I also want to make sure there's no sketchy reason we have a 5.4 allowance in composer but aren't testing for it. I'll create an issue to do so after we figure out why there's a difference. Pretty sure that's an oversight, but want to double check.

chartjes: We should definitely upgrade PHP support in composer.json to 5.5 and redo the lock file.

tests/OpenCFP/Http/Controller/Admin/TalksControllerTests.php
mdwheele: I think there's value in cleaning up tests like this to introduce some level of abstraction that more clearly defines context. This could include factories of any sort, better test helpers, etc. More of a general "improve readability of tests" effort

tests/OpenCFP/Http/Controller/Admin/TalksControllerTests.php
mdwheele: Fair enough. No need to tweet now :smile:

mdwheele:
Waiting on green and planning to merge.

mdwheele:
Thanks, @chartjes!

Issue #272, opened 2015-12-14
`bin/opencfp list` has inaccurate description for `admin:promote`


eryno:
## Steps to Reproduce
1. Run `bin/opencfp list` on the CLI
2. See the text reads: 

    `admin:promote  Promote an existing (or new) user to be an admin`

3. Run `bin/opencfp admin:promote totally_new@person.com` (totally_new@person.com must not already exist in the database)

## Expected results
A user would be created with the email address totally_new@person.com and promoted to admin.

## Actual results
No account was created; instead an error was thrown.

    Retrieving account from totally_new@person.com...
    Error: Could not find user by totally_new@person.com

mdwheele:
Yeah the wording on that description is bad. It should not give the hint that it will create a user unless we want to do that. In that case, the admin would need to set a password for the user.

We could do that and then in the future, add a "must reset password on first login" feature. What do you think?

eryno:
I think that would be a fantastic enhancement, and is, in fact, what I was hoping it would do. But the quick solution was obviously to bring the hint text in line with reality. I'll add a ticket for the improved version.

mdwheele:
I appreciate it. Thanks!

PR #273, opened 2015-12-14
Fixes github issue #272: "bin/opencfp list" has inaccurate description for admin:promote

eryno:
I fixed the issue in #272 by removing the "(or new)" parenthetical from the help text, so no future intrepid programmers will be tripped up by it again.


mdwheele:
Thanks @eryno!

PR #275, opened 2015-12-14
Fixes broken column sorting on admin/talks page.

eryno:
* Clicking on column headings should sort the table by that column, but
   it was passing in the column heading in the "sort" parameter instead
   of the "order_by"

 * The default sort order was DESC, which means string sorted Z-A.
   Changed the default to ASC.


chartjes:
Thanks @eryno

Issue #285, opened 2015-12-19
All talk rating (up/down vote) was removed


jakefolio:
I recently rebased and found that all of the talk rating UI had been removed from Admin Talks page. As of the following hash: afdae85e1ee2fbbe66b93319498dc9a1fb8d316a, this UI was still there.

https://github.com/opencfp/opencfp/commit/afdae85e1ee2fbbe66b93319498dc9a1fb8d316a#diff-2dda05548a15dd6baf3c71e98c897bb2R25

*UPDATE*
This appears to be the point where a lot of work was reverted: https://github.com/opencfp/opencfp/commit/aafc9e5b201fd6ca52c6d999cad699a5b174b846

I'm assuming the submitter was not working off of an up to date version of OpenCFP when creating PR.

mdwheele:
Resolved in #286

PR #286, opened 2015-12-19
Re-adding Sorting and Rating UI

jakefolio:
Per issue: https://github.com/opencfp/opencfp/issues/285

Sorting UI was removed and a different one added instead. I don't believe this was done on purpose, but merely product of an older version of OpenCFP being worked on. This PR re-adds the sort UI that has sort direction arrows.

Also, the up and down vote rating icons are being re-added in this PR.


mdwheele:
Sorry about that @jakefolio. Hope all is well. Thanks!

PR #287, opened 2015-12-19
Fix: Require codeclimate/php-test-reporter:0.2.0

localheinz:
This PR

* [x] requires `codeclimate/php-test-reporter:0.2.0` instead of `dev-master`
* [x] submits coverage to Code Climate the regular way


mdwheele:
Thanks @localheinz!

PR #289, opened 2015-12-19
Fix: Use a whitelist filter instead of blacklisting

localheinz:
This PR

* [x] uses a whitelist filter instead of a blacklist filter when collecting coverage

See https://travis-ci.org/opencfp/opencfp/jobs/97889795#L493-L495:

```
Runtime:	PHP 5.5.21 with Xdebug 2.2.7
Configuration:	/home/travis/build/opencfp/opencfp/phpunit.xml.dist
Warning:	No whitelist configured for code coverage
```

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -10,10 +10,9 @@
          processIsolation="false"
          stopOnFailure="false">
     <filter>
-        <blacklist>
-            <directory>vendor</directory>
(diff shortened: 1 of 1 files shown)


mdwheele:
Made an adjustment to this to exclude `classes/Providers` as it was previously. Otherwise, looks like the warning from Travis is gone! Thanks @localheinz!

localheinz:
@mdwheele 

:+1:

Issue #288, opened 2015-12-19
Speed up Travis build


localheinz:
There are a bunch of possibilities which could further speed up the builds on Travis:

- [x] only collect and report coverage on a merge to `master` (everything else doesn't make sense)
- [x] only collect and report coverage on one PHP version
- [x] only report coverage when the build was a success
- [x] disable xdebug when it's not needed (we only need it when collecting coverage)
- [x] only run `php-cs-fixer` on one PHP version
- [x] do not try to run `php-cs-fixer` on PHP7, it's not supported

:information_desk_person: Doesn't speed up builds, but provides feedback earlier if one build failed:

- [x] allow builds to finish early, see https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing

mdwheele:
@localheinz 

I changed your list to a checklist so that we could track progress a little bit easier. Thanks for all the PRs.

mdwheele:
Everything looks good. The only thing that may get reconsidered down the line is when coverage is generated (which is why I think computing an environment variable that captures "when we generate coverage" is useful). 

I would like it if Codeclimate could report coverage +/- back on PRs to `master` and this is something we've talked to them about before at conferences here and there. Something similar to what coveralls will do. It's not a big priority, but more a thought. For now, I think what you've sent over the past few days helps out a lot and we can always adjust if Codeclimate supports this feature. Honestly, progress over time is already tracked well in CC, it's just the cases where contributors send uncovered code that an extra check helps with. Isn't really a big problem.

Anywho, if you come up with any other ideas on speeding things up, send them our way. Switching to the container infrastructure 5-or-so months ago helped tremendously and I'm sure these changes will shave even more time off.

Thanks again, @localheinz! :hand:

localheinz:
@mdwheele 

We've asked Code Climate for this feature, too - would be really nice to have it!

Closing the issue as we're done here.

PR #290, opened 2015-12-19
Fix: Run php-cs-fixer on PHP5.6 only

localheinz:
This PR

* [x] runs `php-cs-fixer` on PHP5.6 only

Related to #288.


mdwheele:
Thanks @localheinz!

PR #291, opened 2015-12-19
Fix: Do not bother reporting coverage if the build failed

localheinz:
This PR

* [x] moves coverage reporting to the `after_success` section

Related to #288.

:information_desk_person: There's no use for reporting coverage if the build failed, eh?

For reference, see https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle.


mdwheele:
Thanks!

PR #292, opened 2015-12-19
Fix: Use class keyword

localheinz:
This PR

* [x] makes use of the `class` keyword instead of using string literals to reference FQCNs

Follows https://github.com/opencfp/opencfp/pull/270.

:information_desk_person: Makes sense since we've dropped support for PHP5.4, doesn't it?

### DIFF
--- a/classes/Domain/Entity/Favorite.php
+++ b/classes/Domain/Entity/Favorite.php
@@ -21,7 +21,7 @@ class Favorite extends Entity
     public static function relations(\Spot\MapperInterface $mapper, \Spot\EntityInterface $entity)
     {
         return [
-            'talk' => $mapper->belongsTo($entity, 'OpenCFP\Domain\Entity\Talk', 'talk_id'),
+            'talk' => $mapper->belongsTo($entity, \OpenCFP\Domain\Entity\Talk::class, 'talk_id'),
         ];
     }
 }
--- a/classes/Domain/Entity/Group.php
+++ b/classes/Domain/Entity/Group.php
@@ -22,7 +22,7 @@ class Group extends Entity
     public static function relations(\Spot\MapperInterface $mapper, \Spot\EntityInterface $entity)
     {
         return [
-            'users' => $mapper->hasManyThrough($entity, 'OpenCFP\Domain\Entity\User', 'OpenCFP\Domain\Entity\UserGroup', 'user_id', 'group_id'),
+            'users' => $mapper->hasManyThrough($entity, \OpenCFP\Domain\Entity\User::class, \OpenCFP\Domain\Entity\UserGroup::class, 'user_id', 'group_id'),
         ];
     }
 }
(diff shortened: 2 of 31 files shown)


mdwheele:
I intend to merge this PR but I want to go through it a little more carefully than I have time for right now and verify that we have test coverage over each of these refactors before merging. Thanks for the work!

localheinz:
@mdwheele 

You're much welcome, it's fun!

mdwheele:
Sorry about the delay on this. I'm pulling it down today to poke around and verify all refactors are covered by tests. If they aren't, I will be covering them.

mdwheele:
The following classes are currently either completely uncovered or applicable lines uncovered by tests of any fashion:

- `OpenCFP\Domain\Entity\Group`
- `OpenCFP\Domain\Entity\Talk`
- `OpenCFP\Domain\Entity\TalkComment`
- `OpenCFP\Domain\Entity\TalkMeta`
- `OpenCFP\Domain\OAuth\Client`
- `OpenCFP\Http\Controller\Admin\AdminsController`
- `OpenCFP\Http\Controller\Admin\DashboardController`
- `OpenCFP\Http\Controller\Admin\ReviewController`
- `OpenCFP\Http\Controller\Admin\SpeakersController`
- `OpenCFP\Http\Controller\Admin\TalksController`

Before merging this, we should write some tests getting coverage on effected lines. I'll go ahead and write those up and merge this work in with that when it's ready.

localheinz:
@mdwheele 

Sounds great!

Following #301, I'm happy to help adding coverage so that we can refactor without fear!

chartjes:
Any ETA on this? I don't like long-running PR's hanging around

mdwheele:
We don't have coverage on a lot of affected classes. I mean to add coverage and merge, but have been sick. I don't have a good ETA. If you want, we can close and open and issue as this one already has conflicts from surrounding PRs from @localheinz.

That, or we eye-ball these hard enough that we're comfortable there aren't any mistakes and merge without tests :astonished: Don't really feel comfortable doing that.

localheinz:
@chartjes @mdwheele 

Two options:

1. I make it easier to eyeball this by breaking d8abbda6b704e2ad8d568b2457c26ef345c06014 apart into two commits (one that replaces class names with FQCNs using the class keyword, another one which simplifies the FQCNs)
2. I only replace class names where we have coverage.

Thoughts?

Ha, of course, we could also just leave it as is.

mdwheele:
FWIW, I reviewed this line by line and since every refactor uses absolute namespace references, there is no risk of accidentally forgetting to import a namespace partial. This passes tests, albeit vacuously as these are uncovered, but I think we can merge this and @localheinz and I start working on coverage to get the test suite in better order.

You have the conn, @chartjes :rocket:

chartjes:
Carry on

PR #293, opened 2015-12-19
Fix: Order use statements

localheinz:
This PR

* [x] enables `ordered_use` fixer
* [x] runs `make cs`

Follows #242.

> **ordered_use [contrib]**
Ordering use statements.


localheinz:
@chartjes @mdwheele 

How about enabling more fixers, step by step?

mdwheele:
What if we turned on PSR2? I could take the time to work through any breaky breaky. Then you won't have to go fix by fix.

mdwheele:
Whoops. Looks like we already are. I'm down to add more step by step but wouldn't want to go overboard. PSR2 is pretty much good enough from my perspective but that's about the extent of my usage of phpcs honestly.

localheinz:
@mdwheele 

There's a ton of useful fixers which fix issues which go beyond PSR-2, and help maintaining a consistent coding style. Since it's not a lot of work, I wouldn't mind opening a few PRs, then you can decide whether you like the result, what do you think?

mdwheele:
Sounds great. Thanks @localheinz!

PR #294, opened 2015-12-20
Fix: Use short array syntax

localheinz:
This PR

* [x] enables the `short_array_syntax` fixer
* [x] runs `make cs`

Follows #242.

> **short_array_syntax [contrib]**
PHP arrays should use the PHP 5.4 short-syntax.


mdwheele:
Thanks @localheinz!

PR #227, opened 2015-07-16
Add editorconfig

yitznewton:
Support for universal editor settings; see http://editorconfig.org/

Let me know if there are more relevant details in the project's code style

PR #295, opened 2015-12-20
Fix: Set indent size to 2 for YAML files

localheinz:
This PR

* [x] sets `indent_size` to `2` in `.editorconfig`

Follows #227.

:information_desk_person: Throughout the project, YAML files are indented with 2 spaces. Since we're using `.editorconfig`, it makes sense to adjust it, as not to fight an IDE supporting it.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,3 +9,6 @@ indent_size = 4
 
 [*.js]
 indent_size = 2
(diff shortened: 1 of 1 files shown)


mdwheele:
Thanks @localheinz!

PR #296, opened 2015-12-20
Fix: Collect and report coverage on PHP7 only

localheinz:
This PR

* [x] collects and reports coverage on PHP7 only

Related to #288.


mdwheele:
Thanks!

PR #299, opened 2015-12-20
Fix: Allow builds to finish early

localheinz:
This PR

* [x] allows builds to finish early

Related to #288.

:information_desk_person: This, of course, only gives us faster feedback if the build has failed.


mdwheele:
Thanks!

PR #300, opened 2015-12-20
Fix: Disable xdebug when we don't need it

localheinz:
This PR

* [x] disables xdebug when we don't need it

Related to #288.

:information_desk_person: We only need it when we want to collect coverage, right?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ cache:
     - $HOME/.composer/cache
 
 before_install:
+  - if [[ "$WITH_COVERAGE" != "true" ]]; then phpenv config-rm xdebug.ini; fi
   - composer self-update
   - composer validate --no-check-publish
 


.travis.yml
```diff
@@ -21,6 +21,7 @@ cache:
     - $HOME/.composer/cache
 
 before_install:
+  - if [[ "$WITH_COVERAGE" != "true" ]]; then phpenv config-rm xdebug.ini; fi
```
localheinz: @mdwheele 

Depending on whether you like #298, this condition should be adjusted, of course!

mdwheele: Will do. Merging while mobile so I'm going to go ahead and pull this in and edit after the fact.

mdwheele:
Thanks @localheinz ! If you get a chance, please double-check ab55c404e7e96e89f0401a9c679cd7dd4a93e7d3 for me. It's late here and was editing on phone. I want to make sure I understood properly what you meant by adjusting the conditional. I assumed logical negation of the condition where we *would* generate / report coverage.

Wonder if there's a cleaner way for us to express "only generate coverage on a single build when code is merged into master". Probably not... wonder if we can compute it as part of setting up the environment and just reuse that variable further down. Thoughts?

localheinz:
@mdwheele 

Looks good!

PR #302, opened 2015-12-21
Fix: Introduce COLLECT_COVERAGE environment variable

localheinz:
This PR

* [x] sets an environment variable `COLLECT_COVERAGE` to `true` if it's a merge to `master` and to `false` otherwise

Follows https://github.com/opencfp/opencfp/pull/300#issuecomment-166194054.


chartjes:
Thanks @localheinz

PR #304, opened 2015-12-25
Add default logging

jakefolio:
Currently, all errors/exceptions are swallowed up and not logged anywhere. This is incredibly frustrating when trying to debug issues. This commit adds monolog and uses the built in silex provider. Default path will add log to "log/app.log". Log level and log path can be overriden in the environment config files.


mdwheele:
Looks good. Thanks for the work. Could you add the config setting to the development dist and testing configs?

jakefolio:
@mdwheele updated.

mdwheele:
Thanks!

PR #306, opened 2015-12-29
Fix: Each class should be in a file of its own

localheinz:
This PR

* [x] pulls out `SessionDouble` into a file of its own
* [x] adds classmap autoloading for classes in `tests/unit`

Related to https://github.com/opencfp/opencfp/issues/308.

See https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-1-basic-coding-standard.md#3-namespace-and-class-names:

> This means each class is in a file by itself, and is in a namespace of at least one level: a top-level vendor name.

:information_desk_person: This could be part of a series cleaning up the general structure of existing tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,6 +40,9 @@
         "psr-4": {"OpenCFP\\": "classes/"}
     },
     "autoload-dev": {
+        "classmap": [
+            "tests/unit"
+        ],
         "psr-4": {
             "OpenCFP\\": "tests/OpenCFP/"
         }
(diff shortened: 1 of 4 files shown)


mdwheele:
Agreed on continued restructuring of tests. Ideally, things start moving towards the mirror directory structure away from unit. Things under that namespace *shoild* be unit tests but I know we're still working many angles here. Hold off on significant restructure PRs, I'm sending a PR soon to get coverage back within an acceptable range as well as tagging tests that end up under that unit space with a group that really sets them apart. Getting coverage on a Spot2 entity is kinda questionable because we're testing code we don't own. Some test better than none for sure, I'd just like them set aside as integration tests but that's personal preference mostly.

mdwheele:
Related #307

localheinz:
@mdwheele 

:+1:

Issue #279, opened 2015-12-14
Comments should display poster name, not just avatar


eryno:
If a reviewer leaves a comment on a talk, you see their avatar but not their name.

This is especially a problem when a site has multiple reviewers and not everyone has added an avatar.

Comments should display First and Last Names.

chartjes:
Closed thanks to @jakefolio's submission

PR #309, opened 2015-12-30
Add comment author full name to comment meta section

jakefolio:
This makes it easier to determine who wrote which comment. Currently, the only identifier of comment author is avatar. This PR addresses: https://github.com/opencfp/opencfp/issues/279


mdwheele:
Thanks @jakefolio!

PR #311, opened 2015-12-30
Fix: Increase visibility of actions

localheinz:
This PR

* [x] raises the visibilities of actions from `private` to `public`

:information_desk_person: Not sure how the routing and dispatching works, but I don't think that the actions dispatched to should be `private` or `protected`.


mdwheele:
Thanks @localheinz!

Issue #307, opened 2015-12-29
Tests need to be sorted out


localheinz:
A :eyes: into 

* https://github.com/opencfp/opencfp/tree/master/tests/OpenCFP and
* https://github.com/opencfp/opencfp/tree/master/tests/unit

reveals

* test classes are using inconsistent namespaces
* some test classes do not even reside within a namespace
* some files declare multiple classes

I propose to 

* consistently use `OpenCFP\Test` as a namespace prefix
* pull out classes into separate files, following PSR-1

and then continue to sort out the existing tests.

mdwheele:
Agree on namespacing tests though I'd like to hold off long enough to get a PR I'm working on in that improves coverage a bit; truthfully, for my own sanity in resolving conflicts. 

What do you think about mirroring the namespace of production code with expectation that there will be 1:1 match with production class to testcase? Reason I ask is it saves a number of imports in testcases. Really a matter of personal preference for me and the IDE handles most of it anyway so if there's a reason to separate, I'm down.

chartjes:
I see we've done a bunch of merges related to this issue. @localheinz and @mdwheele, can we close this issue or is there more work to be done? If so, maybe we should break this done into smaller issues and actually assign them to people.

mdwheele:
Honestly, an agreement that we start to merge tests into the `tests/OpenCFP` location rather than splitting would be good enough. 

> Any new tests must only be merged if applied under the `tests/OpenCFP` directory / namespace. There will be an ongoing effort to move remaining tests from `tests/unit` into that namespace. In addition, we value clarity in meaning / purposes behind tests. If there is excessive setup required for a test, it should be hidden behind a semantic (and possibly re-usable) method.

Any thoughts on the sentiments of that statement? I can draft a basic `CONTRIBUTING.md` to get us started.

mdwheele:
Hahahahahaha... I never noticed this. Check out https://github.com/opencfp/opencfp/pull/261. I could have SWORE we added a basic `CONTRIBUTING` document. This was closed with un-merged commits. Want me to send another PR? It should be good to pull over, but is 300+ commits, 1 ahead of `upstream/master`.

mdwheele:
I believe we've merged enough work against for right now. The tests are structurally in a better place and we're ready to continue work to improve coverage. Next steps for me will be to go through common duplicated boilerplate for tests that hit `Application` to execute a controller method and find ways of making that easier to work with. 

[...]

PR #312, opened 2016-01-01
Fix: Visibility should be protected

localheinz:
This PR

* [x] reduces the visibility of `setUp()` and `tearDown()` from `public` to `protected`

Related to #307.

:information_desk_person: For reference, see 

* [PHPUnit_Framework_TestCase::setUp()](https://github.com/sebastianbergmann/phpunit/blob/4.8/src/Framework/TestCase.php#L1905-L1911)
* [PHPUnit_Framework_TestCase::tearDown()](https://github.com/sebastianbergmann/phpunit/blob/4.8/src/Framework/TestCase.php#L1937-L1943)

### DIFF
--- a/tests/OpenCFP/Http/Controller/Admin/SpeakersControllerTest.php
+++ b/tests/OpenCFP/Http/Controller/Admin/SpeakersControllerTest.php
@@ -11,7 +11,7 @@ class SpeakersControllerTest extends \PHPUnit_Framework_TestCase
 {
     public $app;
 
-    public function setup()
+    protected function setup()
     {
         // Create our Application object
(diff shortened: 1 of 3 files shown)


tests/OpenCFP/Http/Controller/Admin/SpeakersControllerTest.php
```diff
@@ -11,7 +11,7 @@ class SpeakersControllerTest extends \PHPUnit_Framework_TestCase
 {
     public $app;
 
-    public function setup()
+    protected function setup()
```
localheinz: Method should also be named `setUp()`, but I like to keep both PRs and commits small, so happy to follow up with a separate PR.

PR #315, opened 2016-01-01
Fix: Cover Domain\Services\Login

localheinz:
This PR

* [x] covers `Services\Domain\Login`

Related to #307.


localheinz:
Ha, I see it is somewhat covered in https://github.com/opencfp/opencfp/blob/master/tests/unit/LoginTest.php.

PR #317, opened 2016-01-01
Fix: Cover Infrastructure\Persistence\SpotSpeakerRepository

localheinz:
This PR

* [x] covers the previously uncovered `Infrastructure\Persistence\SpotSpeakerRepository`

Related to #307.


mdwheele:
Would you mind updating this to use Mockery mock long API? It's used in many other places in the test suite  and just want to be consistent. Tests look very readable. Love it!

mdwheele:
Alternatively, this could be written as a true integration test as this class is at the edge of the system and implements persistence with a third-party package. Up to you as this does solve a problem in coverage.

localheinz:
@mdwheele 

:+1:

PR #319, opened 2016-01-01
Enhancement: Enable unused_use fixer

localheinz:
This PR

* [x] enables the `unused_use` fixer
* [x] runs `make cs`

Follows #242.

> **unused_use [symfony]**
Unused use statements must be removed.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -12,6 +12,7 @@ return Symfony\CS\Config\Config::create()
         'multiline_array_trailing_comma',
         'ordered_use',
         'short_array_syntax',
+        'unused_use',
     ])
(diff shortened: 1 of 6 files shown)


mdwheele:
We've got a conflict on `ContainerAwareTest` from a previous commit you recently sent 04f2a6b2f5c247d0ca59022704dfb2f698634667 that was probably based before you sent the updates in that commit. If we get that fixed, this is good to go. 

Thanks for all the help with coverage, @localheinz!

localheinz:
@mdwheele 

You're welcome, it's fun!

PR #326, opened 2016-01-02
Fix: Partially cover Domain\Speaker\SpeakerProfile

localheinz:
This PR

* [x] partially covers `Domain\Speaker\SpeakerProfile`

Related to #307.


mdwheele:
Very good! Thanks!

PR #260, opened 2015-11-15
Export feature

michaelcullum:
Closes #257

Sample CSVs: https://www.dropbox.com/sh/e1pfioq7fcjmli0/AADUx-0_txio4iqhmS-Gszgaa?dl=0

This allows for:
* A full list of talks complete with speaker info (one talk a row)
* A full list of talks anonymised (Speaker info fields, slides and 'More Information' removed)
* A list of all talk names, whether they were selected, speaker names and emails for use in sending acceptance/rejection emails
* A list of all selected talks with full information including speaker info

What must be done before merge
- [ ] It need an icon for the admin menu on the left

There is some neatening up that should probably be done but this is fully functional and I'm not sure if/when I'll have a chance to neaten it up so if someone wants to continue this feel free. I can certainly add on a revised navigation icon however if someone can make the icon (I'm no designer).

Issue #257, opened 2015-11-14
Add export talks list feature


michaelcullum:
- [ ] Export list of talks to CSV (one talk per row) with all information including speaker bio etc.
- [ ] Export list of anonmised talks to CSV (Talk title, length, category and abstract)
- [ ] Export list of email addresses for accepted talks
- [ ] Export list of email addresses for speakers with only rejected talks

michaelcullum:
@jakefolio Would it be possible to get an icon done for the admin side menu?

michaelcullum:
POC: https://github.com/opencfp/opencfp/compare/master...michaelcullum:export-feature

Fully working but could use a little tidying but if the powers that be (@mdwheele?) are happy with that as-is then I'm happy to send up a PR.

chartjes:
Would love to see a PR please

mdwheele:
Yeah, send a PR and I'll gladly review. Few things from diff I'd immediately ask for change:

* Use of boolean `$userData` to expose additional formatted output off the Talk model begs to instead use a more appropriate method (or, better) some sort of Presenter responsible to provide an alternative formatted response.
* In HTTP routing, would it be a good idea to start all export options under something like `/admin/export`? `GET` to route might show all options for export. `POST` might process a request that specifies a particular option.
* Tests, of course!

michaelcullum:
Sure, I did warn you it is pretty nasty and all three of those things are things that need doing (and quite a bit else). Most of it was done just because that was the quickest way of doing it (POC), not because it was the right way. I don't really have much more time I can put into this unfortunately but I've thrown up a PR in case anyone wants to finish it off.

mdwheele:
No problem! Thanks for the effort and PR.

PR #328, opened 2016-01-02
Export feature

michaelcullum:
Old pull request: #260 
Issue: #257 

Still needs an icon for the menu at some point

@chartjes Fixed tests on the old (merged then reverted) PR

### DIFF
--- a/classes/Domain/Entity/Mapper/Talk.php
+++ b/classes/Domain/Entity/Mapper/Talk.php
@@ -28,7 +28,7 @@ class Talk extends Mapper
      * @internal param string $order_by
      * @internal param string $sort Sort Direction
      */
-    public function getAllPagerFormatted($admin_user_id, $options = [])
+    public function getAllPagerFormatted($admin_user_id, $sort, $userData = true, $where = null)
     {
         // Merge options with default options
         $options = $this->getSortOptions(
@@ -44,8 +44,18 @@ class Talk extends Mapper
             ->with(['favorites', 'comments']);
         $formatted = [];
 
+        if ($where) {
+            $talks = $this->all()
+                ->order($sort)
+                ->where($where);
+        } else {
+            $talks = $this->all()
+                ->order($sort)
+                ->with(['favorites']);
+        }
+
         foreach ($talks as $talk) {
-            $formatted[] = $this->createdFormattedOutput($talk, $admin_user_id);
+            $formatted[] = $this->createdFormattedOutput($talk, $admin_user_id, $userData);
         }
 
         return $formatted;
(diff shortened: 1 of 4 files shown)


mdwheele:
@michaelcullum Thanks much for sending this again. I'm about to pull it down to verify everything works. Otherwise, looks good to me. 

I would like to adjust the route registration to be mounted under `admin/export` to give us a easier path when we flesh out this feature-set. I will make these changes as part of the merge process if you don't see any problem doing so. 

In this specific case, the full routes would end up something like:

```php
$web->get('/admin/export/csv', 'OpenCFP\Http\Controller\Admin\ExportsController::attributedTalksExportAction')->bind('admin_export_csv');
$web->get('/admin/export/csv/anon', 'OpenCFP\Http\Controller\Admin\ExportsController::anonymousTalksExportAction')->bind('admin_export_csv_anon');
$web->get('/admin/export/csv/selected', 'OpenCFP\Http\Controller\Admin\ExportsController::selectedTalksExportAction')->bind('admin_export_csv_selected');
$web->get('/admin/export/csv/emails', 'OpenCFP\Http\Controller\Admin\ExportsController::emailExportAction')->bind('admin_export_csv_emails');
```

I'll let this hang for a day or so after validating 1c30637. Also gives me time to write a few basic tests. Thanks!

classes/Domain/Entity/Mapper/Talk.php
```diff
@@ -44,8 +44,18 @@ public function getAllPagerFormatted($admin_user_id, $options = [])
             ->with(['favorites', 'comments']);
         $formatted = [];
 
+        if ($where) {
```
mdwheele: After merging, I had to make quite a few modifications to this code to get it to function as I expected it to. Specifically, the incomplete introduction of `$sort` (replacing `$options`) caused several notices that it was not properly initialized and a fatal error on calls to `Talk::getSortOptions`. I have resolved this and think that this PR is salvageable, but would advise that we look at cleaner ways of implementing these 4 export features. Michael has already said this work was pretty-much proof-of-concept to scratch an itch.

mdwheele:
Even with above comment, I think we can merge the work as-is and revisit soon. It does work as expected once I applied a few patches here and there to remedy surface-level issues. It is not easy to write clear tests for this code. Since it is relatively self-contained as a feature (except for fact that it adds to `Talk` mapper method signatures that may need to be cleaned up, it's reasonable for merge.

mdwheele:
@michaelcullum I have taken your work forward at #328. I'm closing this PR as the extension includes all these commits and makes a few improvements on what you already have done. Thanks again for sending this. Awesome stuff!

michaelcullum:
@mdwheele Thanks for taking this forward. As I explained in my initial PR this was more of a POC just to get something working I could use but couldn't really put any more time into it.

PR #332, opened 2016-01-03
Fix: Invalid default values for DATETIME columns

localheinz:
This PR

* [x] modifies existing migrations to set `DATETIME` field defaults to values within accepted range
* [x] adds a migration which allows `DATETIME` fields to be nullable and changes their default to `NULL`
* [x] adds a migration which updates `DATETIME` fields with previous defaults to `NULL`

Somewhat related to #307.

:person_with_pouting_face: An attempt to run the migrations on a system where MySQL 5.7 is installed results in this:

```
$ vendor/bin/phinx migrate --configuration phinx.yml.dist

Phinx by Rob Morgan - https://phinx.org. version 0.5.0

using config file ./phinx.yml.dist
using config parser yaml
using migration path /Users/am/Sites/opencfp/opencfp/migrations
warning no environment specified, defaulting to: development
using adapter mysql
using database cfp

 == 20141201172522 AlterUserColumnDefaults: migrating

  [PDOException]
  SQLSTATE[42000]: Syntax error or access violation: 1067 Invalid default value for 'created_at'
```

## Server SQL Modes

As suggested by @mdwheele, this is apparently related to different default MySQL modes:

### MySQL 5.5 (as seen on Travis)

See http://dev.mysql.com/doc/refman/5.5/en/sql-mode.html#sql-mode-setting:

> The default SQL mode is empty (no modes set).

```
$ mysql -e "SHOW VARIABLES LIKE '%version%'" -uroot

+-------------------------+-----------------------------+
| Variable_name           | Value                       |
+-------------------------+-----------------------------+
| innodb_version          | 5.5.41                      |
| protocol_version        | 10                          |
| slave_type_conversions  |                             |
| version                 | 5.5.41-0ubuntu0.12.04.1-log |
| version_comment         | (Ubuntu)                    |
| version_compile_machine | x86_64                      |
| version_compile_os      | debian-linux-gnu            |
+-------------------------+-----------------------------+

$ mysql -e "SELECT @@GLOBAL.sql_mode;" -uroot

+-------------------+
| @@GLOBAL.sql_mode |
+-------------------+
|                   |
+-------------------+

$ mysql -e "SELECT @@SESSION.sql_mode;" -uroot

+--------------------+
| @@SESSION.sql_mode |
+--------------------+
|                    |
+--------------------+
```

For reference, see 

* https://travis-ci.org/opencfp/opencfp/jobs/100003130#L459-L470
* https://travis-ci.org/opencfp/opencfp/jobs/100003130#L471-L476
* https://travis-ci.org/opencfp/opencfp/jobs/100003130#L477-L482

### MySQL 5.7

See http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-setting:

> The default SQL mode in MySQL 5.7 includes these modes: [`ONLY_FULL_GROUP_BY`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by), [`STRICT_TRANS_TABLES`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_strict_trans_tables), [`NO_ZERO_IN_DATE`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_zero_in_date), [`NO_ZERO_DATE`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_zero_date), [`ERROR_FOR_DIVISION_BY_ZERO`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_error_for_division_by_zero), [`NO_AUTO_CREATE_USER`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_auto_create_user), and [`NO_ENGINE_SUBSTITUTION`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_engine_substitution).

> The [`ONLY_FULL_GROUP_BY`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by) and [`STRICT_TRANS_TABLES`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_strict_trans_tables) modes were added in MySQL 5.7.5. The [`NO_AUTO_CREATE_USER`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_auto_create_user) mode was added in MySQL 5.7.7. The [`ERROR_FOR_DIVISION_BY_ZERO`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_error_for_division_by_zero), [`NO_ZERO_DATE`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_zero_date), and [`NO_ZERO_IN_DATE`](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_zero_in_date) modes were added in MySQL 5.7.8. For additional discussion regarding these changes to the default SQL mode value, see [SQL Mode Changes in MySQL 5.7](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-changes).

```
$ mysql -e "SHOW VARIABLES LIKE '%version%'" -uroot 

+-------------------------+----------+
| Variable_name           | Value    |
+-------------------------+----------+
| innodb_version          | 5.7.9    |
| protocol_version        | 10       |
| slave_type_conversions  |          |
| version                 | 5.7.9    |
| version_comment         | Homebrew |
| version_compile_machine | x86_64   |
| version_compile_os      | osx10.11 |
+-------------------------+----------+

$ mysql -e "SELECT @@GLOBAL.sql_mode;" -uroot 

+-------------------------------------------------------------------------------------------------------------------------------------------+
| @@GLOBAL.sql_mode                                                                                                                         |
+-------------------------------------------------------------------------------------------------------------------------------------------+
| ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION |
+-------------------------------------------------------------------------------------------------------------------------------------------+

$ mysql -e "SELECT @@SESSION.sql_mode;" -uroot 

+-------------------------------------------------------------------------------------------------------------------------------------------+
| @@SESSION.sql_mode                                                                                                                        |
+-------------------------------------------------------------------------------------------------------------------------------------------+
| ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION |
+-------------------------------------------------------------------------------------------------------------------------------------------+
```

#### `NO_ZERO_DATE`

See http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_zero_date:

>The `NO_ZERO_DATE` mode affects whether the server permits `0000-00-00` as a valid date. Its effect also depends on whether strict SQL mode is enabled.
>
>* If this mode is not enabled, `0000-00-00` is permitted and inserts produce no warning.
>
* If this mode is enabled, `0000-00-00` is permitted and inserts produce a warning.
>
>If this mode and strict mode are enabled, `0000-00-00` is not permitted and inserts produce an error, unless `IGNORE` is given as well. For `INSERT IGNORE` and `UPDATE IGNORE`, `0000-00-00` is permitted and inserts produce a warning.

### DIFF
--- a/migrations/20141201172522_alter_user_column_defaults.php
+++ b/migrations/20141201172522_alter_user_column_defaults.php
@@ -19,8 +19,8 @@ class AlterUserColumnDefaults extends AbstractMigration
             ->changeColumn('reset_password_code', 'string', ['null' => true])
             ->changeColumn('first_name', 'string', ['null' => true])
             ->changeColumn('last_name', 'string', ['null' => true])
-            ->changeColumn('created_at', 'datetime', ['default' => '0000-00-00 00:00:00'])
-            ->changeColumn('updated_at', 'datetime', ['default' => '0000-00-00 00:00:00'])
+            ->changeColumn('created_at', 'datetime', ['default' => '1000-01-01 00:00:00'])
+            ->changeColumn('updated_at', 'datetime', ['default' => '1000-01-01 00:00:00'])
             ->save();
     }
 
--- a/migrations/20160103164746_set_date_time_fields_with_previous_default_to_null.php
+++ b/migrations/20160103164746_set_date_time_fields_with_previous_default_to_null.php
@@ -22,7 +22,13 @@ class SetDateTimeFieldsWithPreviousDefaultToNull extends AbstractMigration
 
     public function up()
     {
-        $this->run('UPDATE `%s` SET `%s` = NULL WHERE `%s` = "1000-01-01 00:00:00"');
+        if ($this->isNoZeroDateMode()) {
+            $sql = 'UPDATE `%s` SET `%s` = NULL WHERE `%s` = "1000-01-01 00:00:00"';
+        } else {
+            $sql = 'UPDATE `%s` SET `%s` = NULL WHERE `%s` IN ("1000-01-01 00:00:00", "0000-00-00 00:00:00")';
+        }
+
+        $this->run($sql);
     }
 
     public function down()
@@ -42,5 +48,23 @@ class SetDateTimeFieldsWithPreviousDefaultToNull extends AbstractMigration
                 ));
             });
         });
+    }
+
+    /**
+     * @return bool
+     */
+    private function isNoZeroDateMode()
+    {
+        $sql = 'SELECT @@GLOBAL.sql_mode AS modes';
+
+        /* @var PDOStatement $statement */
+        $statement = $this->query($sql);
+        $statement->execute();
+
+        $row = $statement->fetch(PDO::FETCH_ASSOC);
+
+        $modes = explode(',', $row['modes']);
+
+        return in_array('NO_ZERO_DATE', $modes);
     }
 }
--- a/migrations/20160103164746_set_date_time_fields_with_previous_default_to_null.php
+++ b/migrations/20160103164746_set_date_time_fields_with_previous_default_to_null.php
@@ -1,0 +1,46 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class SetDateTimeFieldsWithPreviousDefaultToNull extends AbstractMigration
+{
+    /**
+     * @var array
+     */
+    private $tables = [
+        'talks' => [
+            'created_at',
+            'updated_at',
+        ],
+        'users' => [
+            'activated_at',
+            'created_at',
+            'updated_at',
+            'last_login',
+        ],
+    ];
+
+    public function up()
+    {
+        $this->run('UPDATE `%s` SET `%s` = NULL WHERE `%s` = "1000-01-01 00:00:00"');
+    }
+
+    public function down()
+    {
+        $this->run('UPDATE `%s` SET `%s` = "1000-01-01 00:00:00" WHERE `%s` IS NULL');
+    }
+
+    private function run($sql)
+    {
+        array_walk($this->tables, function ($columnNames, $tableName) use ($sql) {
+            array_walk($columnNames, function ($columnName) use ($tableName, $sql) {
+                $this->execute(sprintf(
+                    $sql,
+                    $tableName,
+                    $columnName,
+                    $columnName
+                ));
+            });
+        });
+    }
+}
--- a/migrations/20141203111136_alter_talk_column_defaults.php
+++ b/migrations/20141203111136_alter_talk_column_defaults.php
@@ -14,8 +14,8 @@ class AlterTalkColumnDefaults extends AbstractMigration
             ->changeColumn('favorite', 'boolean', ['default' => 0])
             ->changeColumn('sponsor', 'boolean', ['default' => 0])
             ->changeColumn('selected', 'boolean', ['default' => 0])
-            ->changeColumn('created_at', 'datetime', ['default' => '0000-00-00 00:00:00'])
-            ->changeColumn('updated_at', 'datetime', ['default' => '0000-00-00 00:00:00'])
+            ->changeColumn('created_at', 'datetime', ['default' => '1000-01-01 00:00:00'])
+            ->changeColumn('updated_at', 'datetime', ['default' => '1000-01-01 00:00:00'])
             ->save();
     }
 
--- a/migrations/20150702132854_fix_defaults_for_users_table.php
+++ b/migrations/20150702132854_fix_defaults_for_users_table.php
@@ -21,14 +21,14 @@ class FixDefaultsForUsersTable extends AbstractMigration
         $users->changeColumn('permissions', 'text', ['null' => true]);
         $users->changeColumn('activated', 'boolean', ['default' => 0]);
         $users->changeColumn('activation_code', 'string', ['null' => true]);
-        $users->changeColumn('activated_at', 'datetime', ['null' => true, 'default' => '0000-00-00 00:00:00']);
-        $users->changeColumn('last_login', 'datetime', ['null' => true, 'default' => '0000-00-00 00:00:00']);
+        $users->changeColumn('activated_at', 'datetime', ['null' => true, 'default' => '1000-01-01 00:00:00']);
+        $users->changeColumn('last_login', 'datetime', ['null' => true, 'default' => '1000-01-01 00:00:00']);
         $users->changeColumn('persist_code', 'string', ['null' => true]);
         $users->changeColumn('reset_password_code', 'string', ['null' => true]);
         $users->changeColumn('first_name', 'string', ['null' => true]);
         $users->changeColumn('last_name', 'string', ['null' => true]);
-        $users->changeColumn('created_at', 'datetime', ['default' => '0000-00-00 00:00:00']);
-        $users->changeColumn('updated_at', 'datetime', ['default' => '0000-00-00 00:00:00']);
+        $users->changeColumn('created_at', 'datetime', ['default' => '1000-01-01 00:00:00']);
+        $users->changeColumn('updated_at', 'datetime', ['default' => '1000-01-01 00:00:00']);
 
         // Custom fields
         $users->changeColumn('company', 'string', ['null' => true]);
--- a/migrations/20160103163540_allow_date_time_fields_to_be_null.php
+++ b/migrations/20160103163540_allow_date_time_fields_to_be_null.php
@@ -1,0 +1,71 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class AllowDateTimeFieldsToBeNull extends AbstractMigration
+{
+    public function up()
+    {
+        $options = [
+            'null' => true,
+            'default' => null,
+        ];
+
+        $talks = $this->table('talks');
+
+        $talks->changeColumn('created_at', 'datetime', $options);
+        $talks->changeColumn('updated_at', 'datetime', $options);
+
+        $talks->save();
+
+        $users = $this->table('users');
+
+        $users->changeColumn('activated_at', 'datetime', $options);
+        $users->changeColumn('created_at', 'datetime', $options);
+        $users->changeColumn('updated_at', 'datetime', $options);
+        $users->changeColumn('last_login', 'datetime', $options);
+
+        $users->save();
+    }
+
+    public function down()
+    {
+        $talks = $this->table('talks');
+
+        $talks->changeColumn('created_at', 'datetime', [
+            'null' => false,
+            'default' => '1000-01-01 00:00:00',
+        ]);
+
+        $talks->changeColumn('updated_at', 'datetime', [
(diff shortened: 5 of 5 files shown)


migrations/20141201172522_alter_user_column_defaults.php
```diff
@@ -19,8 +19,8 @@ public function up()
             ->changeColumn('reset_password_code', 'string', ['null' => true])
             ->changeColumn('first_name', 'string', ['null' => true])
             ->changeColumn('last_name', 'string', ['null' => true])
-            ->changeColumn('created_at', 'datetime', ['default' => '0000-00-00 00:00:00'])
-            ->changeColumn('updated_at', 'datetime', ['default' => '0000-00-00 00:00:00'])
+            ->changeColumn('created_at', 'datetime', ['default' => '1000-01-01 00:00:00'])
```
localheinz: @mdwheele 

To be honest, I'm not sure how much sense a default value like that actually makes, how about defaulting to `NULL` instead?

localheinz:
@mdwheele 

What I'm concerned about - other than the default value that doesn't make much sense: modifying migrations! As far as I'm concerned, migrations, once executed, should probably never be changed. 

:person_with_pouting_face:

mdwheele:
Yeah, I agree. It's hard to say with OpenCFP due to the expectation for how it's deployed by users. They are encouraged that this is simply a starter project and to riff on it for their own needs. That said, I think we can still pull this off for new installs of OpenCFP on MySQL 5.7+ and bring older installs back in line.

Making `created_at` and `updated_at` nullable is probably the only version-compatible thing we could do. I believe 5.6+ allows you to set default values for datetime via function, but we're obviously unable to use that.

What if we write a migration that makes the fields nullable and migrates current data where `created_at` and `updated_at` are `0000-00-00 00:00:00` to be `NULL`. Then go back to the migration broken for new MySQL 5.7 installs and for `>=5.7` branch to create the fields as nullable, otherwise do the old. We would a note as to why we're doing this modification.

Alternatively, we could just update the default value as you've done in the older migration and write a new migration that make the fields nullable as well as cleans up data with the old "`null`" values `0000-00-00 00:00:00` and `1000-01-01 00:00:00`

Either of those, I think, would be acceptable. The latter is probably the cleanest. I don't know how I like the idea of branching logic around versions of software but that's because I haven't really ever had to for stuff like this, haha.

localheinz:
> Alternatively, we could just update the default value as you've done in the older migration and write a new migration that make the fields nullable as well as cleans up data with the old "null" values 0000-00-00 00:00:00 and 1000-01-01 00:00:00

:+1:

migrations/20160103164746_set_date_time_fields_with_previous_default_to_null.php
```diff
@@ -0,0 +1,50 @@
...
+            'last_login',
+        ],
+    ];
+
+    public function up()
+    {
+        array_walk($this->tables, function ($columns, $table) {
+            array_walk($columns, function ($column) use ($table) {
+                $this->execute(sprintf(
+                    'UPDATE `%s` SET `%s` = NULL WHERE `%s` = "1000-01-01 00:00:00"',
```
localheinz: I'm unfortunately unable to execute

```sql
UPDATE `%s` SET `%s` = NULL WHERE `%s` IN ("1000-01-01 00:00:00"', "0000-00-00 00:00:00"')
```

as the latter value is illegal.

migrations/20160103164746_set_date_time_fields_with_previous_default_to_null.php
```diff
@@ -0,0 +1,46 @@
...
+            'activated_at',
+            'created_at',
+            'updated_at',
+            'last_login',
+        ],
+    ];
+
+    public function up()
+    {
+        $this->run('UPDATE `%s` SET `%s` = NULL WHERE `%s` = "1000-01-01 00:00:00"');
```
localheinz: I'm unfortunately unable to execute

```sql
UPDATE `%s` SET `%s` = NULL WHERE `%s` IN ("1000-01-01 00:00:00"', "0000-00-00 00:00:00"')
```

as the latter value is illegal.

mdwheele: This could be where we have to do branching based on version of MySQL unless this is handled by some other convention. This can't really be a "new problem", I just haven't had to deal with it since nothing I own is running a version of MySQL that new.

mdwheele:
Related Information and How to reproduce on MySQL < 5.7: 

MySQL 5.7 is failing because the default SQL mode is different between it and all previous versions. `0000-00-00 00:00:00` is MySQL's representation of an invalid date as a "zero date". For example, setting a `DATETIME` field to `2016-04-60 00:00:00` would result in a "zero date" at-rest. On previous version of MySQL, you can set the SQL mode to 

Semantically, these many of these dates **SHOULD** be `NOW` by default, but we cannot do this easily due to previous versions of MySQL not supporting defaults specified as function calls. The best alternative is likely to specify them as "unset" by `NULL`-ing them as we're intending. Just gotta work around some warnings.

migrations/20160103164746_set_date_time_fields_with_previous_default_to_null.php
mdwheele: Found some stuff... but... I kinda want to pull this down and try it out as I have a good way of reproducing on previous versions of MySQL now; instead of thrashing a bit. Technically, you should be able to have that query run by ignoring the warning generated by `NO_ZERO_DATE` SQL mode i.e.:

```sql
UPDATE IGNORE `%s` SET `%s` = NULL WHERE `%s` IN ("1000-01-01 00:00:00"', "0000-00-00 00:00:00"')
```

But it's a bit of a hack. I think it's okay in this one-time case because we simply need to query by that value to be able to `NULL` its value.

mdwheele:
@localheinz When you get a chance, could you query the SQL modes used on your version of MySQL as well as what's on Travis and attach those to the top of the PR?

The following query should do it according to docs.

```sql
SELECT @@SESSION.sql_mode;
```

localheinz:
@mdwheele 

Will do! 🤓

mdwheele:
Hey @localheinz, sorry I didn't get to look at this yesterday. I'm coming off a cold and not really feeling all that well. 

I wonder if it'd be simpler to take advantage of `UPDATE IGNORE` as the documentation states:

> If this mode and strict mode are enabled, '0000-00-00' is not permitted and inserts produce an error, unless IGNORE is given as well. For INSERT IGNORE and UPDATE IGNORE, '0000-00-00' is permitted and inserts produce a warning.

This will produce a warning we could suppress (if Phinx doesn't already do so) and the data migration would run gracefully on any installation of MySQL that has `NO_ZERO_DATE` SQL mode enabled. It might be a good idea to go back and squash this down a bit if you agree with the following (so sorry about flip flopping here, but I think this is relevant being that be both learned why this was happening in this PR, haha):

The original problem surfaces on *new* installs of OpenCFP on a installation of MySQL with `NO_ZERO_DATE` enabled. This PR mitigates that by the following:

1. Modify original migration and set default to a valid date.
2. Introduce a new migration that updates the schema, making applicable fields nullable and defaulting them to null (which is, for our case, semantically correct)
3. Introduce a data migration that corrects all previously "null" (by zero date / minimum valid date) dates.

I think the following change might be appropriate instead as I feel our work here was perhaps over-focused on the immediate cause of errors before we understood why it was happening. We need something that solves this problem for both deployed and new installations, regardless of SQL mode setting:

1. Modify the base schema migration to allow applicable fields be nullable. (fix for new installations)
2. Modify the migrations that alter `users` and `talks` to set `default` to `NULL`, as expected. (this should technically already be the case, but for completion, we should do it)
3. Add new migration that brings deployed installations inline with baseline of 1 and 2.
  1. Make applicable fields nullable and set their defaults to `NULL`
  2. Search for zero dates (the previous default) and replace with `NULL`. The query for this update (on MySQL with `NO_ZERO_DATE` enabled) will require the use of `UPDATE IGNORE` i.e.

```sql
UPDATE IGNORE `%s` SET `%s` = NULL WHERE `%s` = "0000-00-00 00:00:00"
```

Unless I'm missing something, I think this is a "best of both worlds" solution that takes caution in modifying previous migrations and maintains parity between currently deployed instances of OpenCFP while allowing new installs on MySQL 5.7 that has forced this awesome default SQL mode on us! :smile: 

Sorry again for any thrashing about on potential solution. If you would like, I can prepare commits for this and you can pull them over here. I don't mean to cause extra work because you've been very helpful here. Honestly, we're probably getting to a good point soon to re-base migrations for reasons such as the difficulty in strategy we've experienced here.

chartjes:
Where are you folks on this?

mdwheele:
@localheinz was going to take another look given my last comment. Would like your thoughts on that last comment as well. We're trying to get OpenCFP working on MySQL 5.7, where the default SQL mode has changed, disallowing usage of zero dates.

localheinz:
@mdwheele 

Happy to wait for @chartjes' input before proceeding!

chartjes:
I am okay with the UPDATE IGNORE suggestion

localheinz:
@chartjes

:+1:

mdwheele:
@localheinz I have done the suggested work above on [mdwheele/pr332-datetime](https://github.com/opencfp/opencfp/compare/master...mdwheele:pr332-datetime). I ended up having to set the session SQL mode to <blank> to get this to run. Individual `ALTER TABLE` commands would not execute per-field because strict mode would disallow the transaction being that the other fields had invalid defaults (zero date). 

Actually, I think I could change my branch to not require `UPDATE IGNORE` at this point. If you would like, you can pull those commits here; replacing what we have so far and continue if you see any changes that need to be made. Otherwise, I can send a separate PR and we can obsolete this one. Let me know what you want to do.

Thanks again.

localheinz:
@mdwheele 

Let's not complicate things, feel free to open a PR for https://github.com/opencfp/opencfp/compare/master...mdwheele:pr332-datetime, I'm closing this one!

PR #333, opened 2016-01-06
Extension of Export Feature Introduced in #328

mdwheele:
This PR includes commits from #328 and obsoletes that request.

There were a few patches that had to be applied to get the PR to work on `master`:

- There was an assumed incomplete refactor of method signature for sorting on `Talk` mapper.
- In the case you exported "selected talks" when there were no selections, an internal server error occurred. 

Additionally, this adds a few style improvements to navigation and moves where the routes were mounted to `admin/export/*` from `admin/*`. This was done for the sake of future opportunities to expand the export feature.

image

There were no tests for the original pull request and there are no tests for this one. The way this is written is a bit hard to test, but it is possible. Rather than keep a long-running PR open, I think it's okay to merge this as it is relatively self-contained, but we should look to pay some TLC here soon.

@chartjes Could you review and if acceptable, merge?

Old PR: #328


chartjes:
I'm not happy about the lack of tests, but we will be having a testfest for this code base soon enough

PR #334, opened 2016-01-06
Follow-up: Invalid default values for DATETIME columns

mdwheele:
This is a follow-up to #332, where we hashed out a solution to migrating the database to support the `NO_ZERO_DATE` SQL mode that is set, by default, in MySQL 5.7 and is also available to be set in previous versions.

This PR makes changes to previous migrations in order to allow new installations of OpenCFP to be migrated under `NO_ZERO_DATE`. We introduce a new migration to bring older installations of OpenCFP in line with those changes.

Applicable fields in `users` and `talks` are set as `nullable` with a default value of `null`. All zero dates in those fields are converted to `null`. Please take care in reviewing this PR as it does make changes to previous migrations, which is generally not a good idea TM. An exception is reasonable in this case because of the compensating final migration that brings everything in line.

/cc @localheinz 

Old PR: #332


chartjes:
Thanks @localheinz and @mdwheele

localheinz:
@chartjes @mdwheele 

:+1:

PR #342, opened 2016-01-14
Apply Consistency Fixups to tests/OpenCFP

mdwheele:
Related: #307 

This PR applies a number of consistency fixes to all test cases under `tests/OpenCFP`, the intended migration location of all unit tests for the system. Here is a summary of things I looked for and changed.

* If a test talked to the database (sqlite) it has been marked as part of `group db`. This is so we can target these for improvement, isolation of real business logic. Thinking forward, these might actually be more clear if broken out as easier to maintain end-to-end tests.
* If a test class name differed from file name, corrections were made to have the test match the production system under test with `Test` appended to class name.
* Every test case under `tests/OpenCFP` is now namespaced under `OpenCFP\Test`. This caused a lot of package import breakage, which was also resolved. 
* Property visibility for systems under tests are `private`, down from `protected` or `public`.
* Visibility and casing of PHPUnit `setUp` and `tearDown` methods was set to `protected` to match parent, and camel-cased instead of lowercase.
* A comment was placed between test methods and factory methods / helpers as a search target for future refactoring of these common methods into either a **well thought out** trait or an abstract test case.

This PR purposefully does not make changes to any tests under `tests/unit` as there will be on-going effort to merge those tests under this namespace. It was just easier to get the destination cleaned up before doing so. Many tests under `tests/unit` overlap with cases already under `tests/OpenCFP`, so care must be taken to cull any double-coverage as well as characterize the tests for future work. Many of those tests are far-reaching, non-intention-revealing, hard to read. This would be improved as part of the ongoing improvements to the test suite.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,7 @@ We accept contributions via Pull Requests on [Github](https://github.com/opencfp
 
 - **[PSR-2 Coding Standard](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md)** - The easiest way to apply the conventions is to run `make cs`, which will apply all code-style fixes necessary.
 
-- **Add tests!** - Your patch probably won't be accepted if it doesn't have tests. Any new tests must be placed in the `tests/OpenCFP` directory and properly namespaced at `OpenCFP\Tests`. There will be an ongoing effort to move remaining tests from `tests/unit` into that namespace.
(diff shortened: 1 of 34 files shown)


mdwheele:
Fixing CS issues... for some reason, I'm getting different CS results on Travis than on my local machine running `make cs`. The specific one is due to an extra carriage return. Shrug. Fixed.

mdwheele:
@chartjes @localheinz If you get a chance tonight, review these. I've got another few PRs to send that have tests that might conflict if not based on this. I'm currently basing changes on this PR.

localheinz:
@mdwheele 

:+1: Looks good to me, great job!

chartjes:
@mdwheele Looks good to me -- *pushes red button*

mdwheele:
:rocket: Thanks, guys.

PR #344, opened 2016-01-19
Sort tests out of unit directory into namespaced location.

mdwheele:
This PR moves the remaining tests left over inside the `tests/unit` directory into the namespaced (`OpenCFP\Test`) location at `tests/OpenCFP`. Eventually, we might split integration tests from unit tests currently under the OpenCFP test-namespace.

I removed some classes from `tests/unit` that did not fail the sweet after being removed. If those need to be restored, let me know and I'll be glad to.

This is related to #307 and if merged, completes the work described there, I believe. The only thing in question is multiple classes per-file and I have applied a regex search that looks for multiple occurrences of the keyword, `class`. I found nothing. Should be good to go.


mdwheele:
Had an old autoloader from earlier commit that allowed suite to pass locally. Removed offending classmap from `composer.json` and have pushed a fix. I have another PR for adding airport information to Speakers administrative overview. I am holding that until this goes in as it m

localheinz:
@mdwheele 

:+1:

chartjes:
Thanks @mdwheele

PR #348, opened 2016-01-25
Added conference information from new config settings

GeeH:
One of the first things regular speakers look for on CFP page is the conference date and location. This PR replaces the text "For More Information About The Conference" with the text "GeeHCon is on 26/10/2014 to 31/10/2014 in Swansea, UK".

This only happens if the two new config values that are added are present, and so is backwards compatible.

The new config values are:

event_location - the location of the event in string format
date_format - a Twig `date()` filter compatible string to format the event dates

### DIFF
--- a/config/development.dist.yml
+++ b/config/development.dist.yml
@@ -3,6 +3,7 @@ application:
   url: http://localhost
   email: contact@openconf.com
   eventurl: http://localhost
+  event_location: Swansea, UK
   enddate: Oct. 14th, 2016
   show_submission_count: false
   airport: MIA
@@ -10,6 +11,8 @@ application:
   departure: 2014-10-31
   secure_ssl: false
   online_conference: false
+  date_format: d/m/Y
+
 
 api:
   enabled: true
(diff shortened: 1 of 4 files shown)


mdwheele:
Hey @GeeH! I'm presuming this is *not* troll, but I'm getting some strong "boy who cried wolf" vibes. :smile: 

Looks like a good addition. What do you think about making the default location `~` (nil) such that the feature is not enabled by default. Folks can enable it when they configure their instance for the first time.

Introducing configurable representation of dates is good as well. I think we should create an issue forward from this request to apply that filter to other dates as well as standardize the configuration file's date/time expectations.

Small things:

- [ ] The [testing configuration](https://github.com/opencfp/opencfp/blob/master/config/testing.yml) schema should be updated to reflect this change.

GeeH:
Yes, believe it or not this isn't a troll :smile:

I've updated PR as you suggested.

GeeH:
With regards to the defaults, I just copied the existing rules, for example the airport code defaults to MIA, and the start/end dates have data in them.

mdwheele:
Thanks @GeeH!

Issue #280, opened 2015-12-14
[Enhancement] Display airport name next to code


eryno:
## Current behavior
A speaker's airport code is displayed as a link to a google search for flights to and from that airport. (Which is definitely pretty cool.)

## Suggested enhancement
Display the full airport name + city, state, and country next to the airport code in order to save reviewers a click.

jsundquist:
There is an api for google flights which could possibly accomplish similar to what #239 was asking for.

pmeth:
Here is an API I found https://developers.google.com/qpx-express/

However, I think to just display the airport name it would be better to just have a static list of airport codes.  Really, how often do they change? I believe I already have such a table someplace. I can add it as a migration to partially address this feature request.

eryno:
I agree, @pmeth. Airport information can safely be in a table somewhere, along with location information (it's not always clear what city/country an airport is in just by the name).

pmeth:
anyone who wants to implement this feature can continue on with the 280_airport-lookup branch. it basically just has a migration to add a table with a list of airports

mdwheele:
I'd like to pick up initial work in doing a first pull, syncing it with new data, and exposing it to the rest of the codebase as a shared resource. Building the feature on top of that could be a continuance of that work.

pmeth:
good by me. it would be useful in other places too i'm sure.

mdwheele:
Where'd you pull this data from?

pmeth:
don't rightly remember, but probably here: http://www.nationsonline.org/oneworld/IATA_Codes/airport_code_list.htm

chartjes:
Closed because code for this issue got merged in

mdwheele:
A migration that added a bunch of airport codes was merged, but nothing was written to expose any of that data to the web interface.

chartjes:
Reopening

mdwheele:
I'm about finished with my stab at this. Just hooking it up to the web interface now. Most of the problems I hit were to do with vendor-specific SQL that Sqlite3 did not play well with. Unrelated, some of the current migrations absolutely do not run or impartially run against Sqlite3. I'll be sending a separate PR for those.

mdwheele:
All, What do you think of the following? Any advice on whether this can be improved?

![image](https://cloud.githubusercontent.com/assets/2453394/12368405/26fcb34a-bbb9-11e5-926c-be6b29cd0f36.png)

[...]

PR #349, opened 2016-01-25
Display Additional Airport Information in Admin Interface

mdwheele:
Related: #280 

This PR introduces a value object for Airport Information (`AirportInfo`). This includes the airport's IATA code, name and country. The work includes a new interface for retrieving airport information by IATA code as well as an implementation using `illuminate/database`. Airport information is shown as part of the admin interface to all speakers as well as on each speaker profile page:

image

This PR also adds the ability to migrate an in-memory Sqlite3 database using Phinx and then re-use that connection for relying infrastructure, such as `illuminate/database`. Previously, this was not possible because after migrations were run (via command-line by hand, or by the test-suite) the connection is dropped and the in-memory database goes away. The way we get around this is by executing Phinx as a console application in PHP, rather than from the command-line. If we do this, we are able to extract a reference to the instance of PDO it used to do migrations and can inject that anywhere that will take it for further work. This is what I have done here in [`DatabaseTestCase`](https://github.com/mdwheele/opencfp/blob/display-airport-info/tests/DatabaseTestCase.php).

Other small things:

- Added a trait for object immutability.
- Added a new Phinx environment called `memory`.
- Fixed an old migration that was incompatible with Sqlite3 through the use of `CONCAT` function. There are no backwards incompatible changes here as without this fix, a user cannot run OpenCFP on Sqlite3, period.


mdwheele:
I'll sort this out tonight. Getting reporting differences between local and Travis regarding Phinx environment.

mdwheele:
Sweet. I've got it figured out. I just need to pass additional configuration to `ArgvInput` to specify configuration path for the execution of `PhinxApplication` within the test case. Will do this next break I get and rebase the "test builds" on Travis.

mdwheele:
Alright, this PR is now passing and is ready for review on content of work.

chartjes:
Thanks @mdwheele

PR #357, opened 2016-02-27
Add timezone to .yml config files

davethegr8:
Allows users to specify their default timezone via config files. The value is stored in `application.date_timezone` and is set in the application as the last step in `__construct`.

Normally this is set on the server side in `php.ini` files, so no worries if you don't want to merge this. I could see value (especially on shared servers) in having this be configurable app-side so I wanted to suggest & provide a pr for it.

I'm not sure if any tests need to be added to this? I'm happy to add them if needed. Also totally happy to move things around if they would go in better places.

### DIFF
--- a/classes/Application.php
+++ b/classes/Application.php
@@ -93,7 +93,7 @@ class Application extends SilexApplication
 
         $this->registerGlobalErrorHandler($this);
 
-        if($timezone = $this->config('application.date_timezone')) {
+        if ($timezone = $this->config('application.date_timezone')) {
             date_default_timezone_set($timezone);
         }
     }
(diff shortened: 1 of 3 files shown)


classes/Application.php
```diff
@@ -92,6 +92,10 @@ public function __construct($basePath, Environment $environment)
         $this->register(new ApplicationServiceProvider);
 
         $this->registerGlobalErrorHandler($this);
+
+        if($timezone = $this->config('application.date_timezone')) {
```
localheinz: @davethegr8 

The build is failing because of a missing space after `if`, see https://travis-ci.org/opencfp/opencfp/jobs/112294546#L591.

davethegr8: Whoops, sorry about that.

mdwheele:
No worries. Thanks @davethegr8!

PR #364, opened 2016-02-28
Fix: Classname should end in Test

localheinz:
This PR

* [x] renames `TalksControllerTests` to `TalksControllerTest`


localheinz:
@chartjes @mdwheele @pmeth 

The build failing here indicates that the test has never actually been run before, doesn't it?

See https://travis-ci.org/opencfp/opencfp/builds/112438321.

mdwheele:
That seems correct as the test-case previously ended with `Tests` rather than `Test`. See note at https://phpunit.de/manual/current/en/organizing-tests.html#organizing-tests.filesystem.

It'd probably be best to go ahead and resolve the issue as part of this PR after determining expected behaviour.

localheinz:
@mdwheele 

Will have a look tomorrow!

Can you do me a big favour, though, and have a look at #368? Will make it a lot easier to run the tests locally!

mdwheele:
Done!

localheinz:
Thank you, @mdwheele!

localheinz:
@mdwheele 

Do you want to take a look again?

chartjes:
This has been hanging around long enough...

PR #365, opened 2016-02-28
Enhancement: Leverage GitHub issue and pull request template capabilities

localheinz:
This PR

* [x] leverages GitHub issue and pull request template capabilities

:information_desk_person: For reference, see https://github.com/blog/2111-issue-and-pull-request-templates.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,13 @@
+# Steps required to reproduce the problem
+
+1. 
+2. 
+3. 
+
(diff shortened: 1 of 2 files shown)


mdwheele:
Looks good to me. Only wondering if we should explicate fact that it is *just* a suggested template and that if someone needs to break from it, they totally can. Don't want to see "connextra syndrome" leaving someone who has a simple question or enhancement thinking they have to try to fit that square peg into a round hole, haha.

Probably a non-issue, but I'm fine with merging at any rate.

PR #366, opened 2016-02-28
Fix: Run php-cs-fixer with --diff on Travis

localheinz:
This PR

* [x] runs `php-cs-fixer` with the `--diff` option on Travis

Related to https://github.com/opencfp/opencfp/pull/357/files#r54354143.

:information_desk_person: We're using that option in [`Makefile`](https://github.com/opencfp/opencfp/blob/master/Makefile#L4), and using it on Travis at least helps to see what the actual problem is when the build fails due to CS issues.


mdwheele:
Thanks for all the work this afternoon, @localheinz. Looks great!

localheinz:
Thank you, @mdwheele!

PR #368, opened 2016-02-28
Enhancement: Streamline testing process

localheinz:
This PR

* [x] uses `cfp_test` as database name for testing instead of `cfp_travis`
* [x] adds `it`, `composer`, `database`, and `test` targets to `Makefile`

:information_desk_person: Run

```
$ make
```

to run all the things!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
 
 env:
   global:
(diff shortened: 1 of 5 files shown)


mdwheele:
Thanks!

PR #369, opened 2016-02-29
Changing value passed to google flights URL to match airport code

madasha:
While using the software I found this great functionality to check airfare cost for speakers flying in from somewhere through google flights. There was only a minor problem - the value passed to the 'f' param in the URL is supposed to be the code of the airport, instead of the full name to make it work. Like this:
https://www.google.com/flights/#search;f=BRU;t=SOF;d=2016-10-07;r=2016-10-09

Made a minor change in two templates in the admin area to reflect this in the composed URL.


chartjes:
Thanks @madasha

madasha:
It's a pleasure!

PR #374, opened 2016-03-22
Fix: Add missing docblocks

localheinz:
This PR

* [x] adds missing docblocks to `ResetEmailer`

Somewhat related to #373.


mdwheele:
Thanks @localheinz!

PR #375, opened 2016-03-22
Fix: Introduce explaining variable when fetching Twig_Environment from container

localheinz:
This PR

* [x] introduces an explaining variable `$twig` when fetching `Twig_Enviroment` from the container

:information_desk_person: This allows to provide an inline docblock, which is helpful with 

* auto-completion
* reducing errors related to undefined methods when running inspections
* finding usages

and it also micro-optimizes the application, yay!

:person_with_pouting_face: In related news, I'm happy to sort out a few more of these issues. Also, noted that there are a bunch of different ways of fetching services from the container

* sometimes using the `ContainerAware` trait in controllers, services are fetched from the field `$app`
* sometimes using the `ContainerAware` trait in controllers, services are fetched using the method `service()`
* sometimes the container is passed into a controller action

What do you think, would it be worth to streamline this?

### DIFF
--- a/classes/Application.php
+++ b/classes/Application.php
@@ -33,6 +33,7 @@ use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\HttpExceptionInterface;
+use Twig_Environment;
 
 class Application extends SilexApplication
 {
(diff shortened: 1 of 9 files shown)


mdwheele:
Changes look good to merge. If you want to include a sweep of other consistency changes in this PR, that's fine. 

So, if a controller uses `ContainerAware`, I think it should use the `service` method for one reason. I can quickly scan for usages of that for a future refactor with regards to your third point, "container passed into action".

That happens quite a bit and is a "Silex defaults" thing, as far as I can tell. Ultimately, I would like to move more and more towards a service layer that is shared by WebControllers and ApiControllers. With that, I would inject the applicable services from service layer into controllers and pass input from web / api to the service; catching exceptions.

I think a PR is in order to standardize how we're accessing the container through `ContainerAware` (to use the method instead of direct-access. I think the places where container is passed into action is worth refactoring to `ContainerAware` if we want to for consistency sake; otherwise, it's one of those "refactor as we expand the service layer" scenarios, I think.

localheinz:
@mdwheele 

Let me update the PR, then!

localheinz:
@mdwheele 

Quick heads-up, I've identified the following services:

* [ ] `application.speakers`
* [ ] `application.speakers.api`
* [ ] `controller.api.talk`
* [ ] `controller.oauth.authorization`
* [ ] `controller.oauth.clients`
* [ ] `controllers_factory`
* [ ] `db`
* [ ] `flash`
* [ ] `form.factory`
* [ ] `mailer`
* [ ] `oauth.resource`
* [ ] `profile_image_processor`
* [ ] `purifier`
* [ ] `request`
* [ ] `reset_emailer`
* [ ] `resolver`
* [ ] `sentry`
* [ ] `security.random`
* [ ] `session`
* [ ] `spot`
* [x] `twig`
* [ ] `url_generator`

In regard to the size of this PR, I believe introducing explaining variables for all of them in this PR would be too much, what do you think?

Issue #340, opened 2016-01-14
Upgrade from Cartalyst to Sentinel


chartjes:
Cartalyst is no longer supported but https://github.com/cartalyst/sentinel appears to be an OSS option we should move to with should cause less friction.

PR #376, opened 2016-03-23
Fix: Introduce explaining variable when fetching Sentry from container

localheinz:
This PR

* [x] introduces an explaining variable `$sentry` when fetching `Cartalyst\Sentry\Sentry` from container

Follows #375.


localheinz:
Also related to #340.

PR #377, opened 2016-03-23
Fix: Actually mock Sentry

localheinz:
This PR

* [x] actually uses `Cartalyst\Sentry\Sentry` (instead of `stdClass`) when mocking

Somewhat related to #376.


localheinz:
Also related to #340.

PR #380, opened 2016-03-24
Fix: Remove useless docblocks

localheinz:
This PR

* [x] removes useless docblocks from console commands 

:information_desk_person: No need to repeat what is automatically inherited from the parent. As a heads-up, this can be part of a series cleaning up console commands!

### DIFF
--- a/classes/Console/Command/AdminDemoteCommand.php
+++ b/classes/Console/Command/AdminDemoteCommand.php
@@ -11,11 +11,6 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 class AdminDemoteCommand extends BaseCommand
 {
-    /**
-     * Configure the command options.
-     *
-     * @return void
-     */
     protected function configure()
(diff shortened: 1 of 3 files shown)


classes/Console/Command/AdminDemoteCommand.php
```diff
@@ -33,13 +28,6 @@ protected function configure()
 );
     }
 
-    /**
-     * Execute the command.
-     *
-     * @param  \Symfony\Component\Console\Input\InputInterface   $input
-     * @param  \Symfony\Component\Console\Output\OutputInterface $output
-     * @return void
```
localheinz: This is actually outdated, too. 

Running

```
$ composer show | grep symfony/console
```

yields

```
symfony/console                   v2.8.0             Symfony Console Component
```

See https://github.com/symfony/console/blob/v2.8.0/Command/Command.php#L158.

PR #383, opened 2016-04-02
Fix: Cover Console\Application

localheinz:
This PR

* [x] covers `Console\Application`

Follows https://github.com/opencfp/opencfp/pull/380.


chartjes:
Thanks @localheinz

PR #387, opened 2016-04-17
Fix: Do not try too much

localheinz:
This PR

* [x] tries not to `try` too hard

Follows #380.

💁There's probably no need to extend the `try` and `catch` block further than trying to find a user, especially if we're only catching `UserNotFoundException`s.


chartjes:
Thanks @localheinz

PR #388, opened 2016-04-17
Fix: Undefined class

localheinz:
This PR

* [x] prepends a class reference with `\` so it can be resolved


chartjes:
Thanks @localheinz

PR #389, opened 2016-04-23
Enhancement: Use SymfonyStyle to streamline console output

localheinz:
This PR

* [x] uses `SymfonyStyle` to streamline the console application output

Follows #380.


mdwheele:
Sweet! Thanks @localheinz.

PR #390, opened 2016-04-25
Fix: Retrieve services using service()

localheinz:
This PR

* [x] retrieves service from the service container using `service()` as provided by the `ContainerAwareTrait`

@mdwheele Not sure in which PR, but I think we briefly talked about streamlining this some time ago, what do you think?

PR #392, opened 2016-04-28
Update dependencies

mdwheele:
This PR

* [x] Updates dependencies for OpenCFP in an effort to resolve what looks like rewritten history upstream in `symfony/config` conflicting with `composer.lock`.

See: https://travis-ci.org/opencfp/opencfp/builds/126466337
Related to #390.

I'm a little unsure on this one honestly, but updating dependencies should be done periodically anyway. I am unsure of root cause because **I** was actually able to install dependencies just fine on my local virtual machine. However, on Travis CI, when it pulls that specific commit, it bails. Also, when I clone `symfony/config` down, that commit is gone.

It's also very strange that #390 passed tests but the merge is what failed. I suppose that the commit could have been removed in the space between PR submission and merge? Shrug. Stranger things have happened, I suppose.

We'll see if this helps.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "91fceef1071670a4507c93a9d1790ba8",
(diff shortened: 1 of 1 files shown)


mdwheele:
Seems to have passed. I am going to go ahead and merge since this is such a small update so that we can have a passing baseline for further PR.

Issue #262, opened 2015-11-16
Twig Fatal Error due to missing variable


michaelcullum:
A fatal error now occurs on /admin/review

```
[16-Nov-2015 02:30:18 Europe/Paris] PHP Fatal error:  Uncaught exception 'Twig_Error_Runtime' with message 'Variable "current_page" does not exist in "layouts/default.twig" at line 12' in \var\www\opencfp\vendor\twig\twig\lib\Twig\Template.php:391
```

jakefolio:
I believe this PR should fix it: https://github.com/opencfp/opencfp/pull/263

michaelcullum:
@jakefolio Just pulled your fix and I'm still getting the exception.

jakefolio:
@michaelcullum really? I just branched off master, made the changes, and spun up local php test server. The URL you specified is working fine for me.'

Have you ran all the migrations?

chartjes:
Just tried with a fresh install, does not generate any errors. Closing as resolved

vlucas:
Heads up: I ran into this, but it was when I was using the built-in PHP server via:
```
php -S localhost:8080 web/index.php
```

The pages all work, but none of the static assets load - for some reason you get the "Variable "current_page" does not exist" errors instead. You probably just need to be clear in the docs that the PHP dev server is not supported (or fix the asset loading! :)).

chartjes:
It does state in the README that Apache 2+ with mod_rewrite enabled is a requirement. Anything else isn't supported.

vlucas:
Yeah - I saw the Apache/Nginx - I just didn't want to have to set those up on my local machine to make a few styling changes, so I tried the built-in PHP webserver first. Not a big issue, but that's what I imagine this guy did as well.

jakefolio:
@vlucas I think I've always used Homestead for developing with OpenCFP. If we setup a basic router script for the built-in PHP server, that should resolve the issues, no?

chartjes:
Yes, that would be great.

PR #393, opened 2016-05-23
Scripts to Rule Them All (PHP built-in web server support)

jakefolio:
This PR adds the Scripts to Rule Them All pattern, https://github.com/github/scripts-to-rule-them-all,  to allow for a more simplified development experience. Also, I've included a server.php router file for PHP built-in web server. I have re-organized the README to move local development higher up and created an entire section for Scripts to Rule Them All.

Related to #262

### DIFF
--- a/server.php
+++ b/server.php
@@ -1,0 +1,24 @@
+<?php
+
+define('WEB_PATH', __DIR__ . '/web/');
+
+if (preg_match('/\.css|\.js|\.jpg|\.png|\.map|\.svg|\.ico$/', $_SERVER['REQUEST_URI'], $match)) {
+    $mimeTypes = [
+        '.css' => 'text/css',
+        '.js'  => 'application/javascript',
+        '.ico' => 'image/ico',
+        '.jpg' => 'image/jpg',
+        '.png' => 'image/png',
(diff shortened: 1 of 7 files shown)


server.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+define('WEB_PATH', __DIR__ . '/web/');
+
+if (preg_match('/\.css|\.js|\.jpg|\.png|\.map|\.svg|\.ico$/', $_SERVER['REQUEST_URI'], $match)) {
+    $mimeTypes = [
```
vlucas: I think you can just `return false;` to serve static assets as-is, right?

jakefolio:
@vlucas the update isn't too simplistic is it? I tested locally and it responded correctly, correct headers too.

vlucas:
Not as long as you don't plan on hitting a `.php` file directly :)

jakefolio:
It should be locked down to the web directory, so I didn't really feel it would be necessary to filter out .php file requests, but can add that in.

vlucas:
Yeah, given that no .php files are under `web`, you honestly should be fine.

chartjes:
Thanks @jakefolio

Issue #398, opened 2016-06-28
Admin Dashboard fails to load after deleting a user with talks


ramsey:
# Steps required to reproduce the problem

1. Register a new speaker.
2. Under the speaker's account, create a new talk.
3. As an admin, log in and go to the Admin > Speakers page.
4. Click "Delete" next to the new speaker you created in Step 1.
5. Go back to the "Admin Dashboard."

# Expected Result

I expect to see the Admin Dashboard page render, listing recent talks.

# Actual Result

Instead of rendering the Admin Dashboard, you will see the main OpenCFP home page with the message "Internal Server Error!" on it.

In the logs, you'll see something like:

```
[2016-06-28 18:13:15] opencfp.INFO: Matched route "admin". {"route_parameters":{"_controller":"OpenCFP\\Http\\Controller\\Admin\\DashboardController::indexAction","_route":"admin"},"request_uri":"http://cfp.mysite.com/admin"} []
[2016-06-28 18:13:15] opencfp.INFO: > GET /admin [] []
[2016-06-28 18:13:15] opencfp.CRITICAL: Twig_Error_Runtime: An exception has been thrown during the rendering of a template ("Parameter "id" for route "admin_speaker_view" must match "[^/]++" ("" given) to generate a corresponding URL.") in "admin/talks/_table.twig" at line 22. (uncaught exception) at /home/mysite/opencfp/vendor/twig/twig/lib/Twig/Template.php line 396 {"exception":"[object] (Twig_Error_Runtime(code: 0): An exception has been thrown during the rendering of a template (\"Parameter \"id\" for route \"admin_speaker_view\" must match \"[^/]++\" (\"\" given) to generate a corresponding URL.\") in \"admin/talks/_table.twig\" at line 22. at /home/mysite/opencfp/vendor/twig/twig/lib/Twig/Template.php:396, Symfony\\Component\\Routing\\Exception\\InvalidParameterException(code: 0): Parameter \"id\" for route \"admin_speaker_view\" must match \"[^/]++\" (\"\" given) to generate a corresponding URL. at /home/mysite/opencfp/vendor/symfony/routing/Generator/UrlGenerator.php:163)"} []
[2016-06-28 18:13:15] opencfp.INFO: < 500 [] []
```

# Temporary Solution to Fix (if you encounter this)

Simply delete all rows from the `talks` table for the user you deleted. This may not be easy if you have a lot of speakers and talks and you don't remember the user ID for the speaker you deleted.

[...]

PR #399, opened 2016-06-29
Fix 398

mdwheele:
This PR resolves the issue reported in #398. 

There may be better ways of doing this (a migration that sets up foreign key relations to automatically cascade deletes when a talk is removed), but this seems to get the job done. I need someone to double-check the work for readability as the test is kinda gnarly with the amount of arranging we have to do.

### DIFF
--- a/classes/Http/Controller/Admin/SpeakersController.php
+++ b/classes/Http/Controller/Admin/SpeakersController.php
@@ -160,7 +160,9 @@ class SpeakersController extends BaseController
         $mapper = $spot->mapper(\OpenCFP\Domain\Entity\User::class);
         $speaker = $mapper->get($req->get('id'));
 
-        $spot->config()->connection()->beginTransaction();
+        $connection = $spot->config()->connection();
+
+        $connection->beginTransaction();
 
         try {
             $this->removeSpeakerTalks($speaker);
@@ -170,14 +172,14 @@ class SpeakersController extends BaseController
         }
 
         if ($response === false) {
-            $spot->config()->connection()->rollBack();
+            $connection->rollBack();
 
             $ext = "Unable to delete the requested user";
             $type = 'error';
             $short = 'Error';
         } else {
-            $spot->config()->connection()->commit();
-            
+            $connection->commit();
+
             $ext = "Successfully deleted the requested user";
             $type = 'success';
             $short = 'Success';
--- a/classes/Http/Controller/Admin/SpeakersController.php
+++ b/classes/Http/Controller/Admin/SpeakersController.php
@@ -2,6 +2,7 @@
 
 namespace OpenCFP\Http\Controller\Admin;
 
+use OpenCFP\Domain\Entity\User;
 use OpenCFP\Domain\Services\AirportInformationDatabase;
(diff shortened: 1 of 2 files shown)


mdwheele:
Going to grab dinner. I will fix any failures on Travis if they come up. Tests pass locally 😬

classes/Http/Controller/Admin/SpeakersController.php
```diff
@@ -157,6 +159,9 @@ public function deleteAction(Request $req)
 
         $mapper = $spot->mapper(\OpenCFP\Domain\Entity\User::class);
         $speaker = $mapper->get($req->get('id'));
+
+        $this->removeSpeakerTalks($speaker);
+
         $response = $mapper->delete($speaker);
```
ramsey: I'm not familiar with the tools OpenCFP uses, but I think it would be good to wrap these lines in a transaction block that will do a rollback if there are any problems at all (either removing talks or deleting the speaker). This way, if the speaker deletion process fails for any reason, then all their talks won't be deleted.

Then again, maybe the Spot\Mapper tools are already doing this for you?

ramsey:
Aside from my code comment about the use of database transactions, everything else here looks pretty good. This seems like the proper way to handle this, short of database cascading deletes on the foreign keys.

mdwheele:
Okie-doke, I've wrapped everything in a transaction via `Spot\Locator::config()->connection()`. I'm not intimately familiar with Spot2, but I think this is reasonable. Tests are passing, just need @chartjes or someone else to review + merge.

classes/Http/Controller/Admin/SpeakersController.php
```diff
@@ -157,7 +159,17 @@ public function deleteAction(Request $req)
...
+        $spot->config()->connection()->beginTransaction();
+
+        try {
+            $this->removeSpeakerTalks($speaker);
+            $response = $mapper->delete($speaker);
+        } catch (\Exception $e) {
+            $spot->config()->connection()->rollBack();
+        }
+
+        $spot->config()->connection()->commit();
```
ramsey: The `commit()` call should go inside the `try` block. Otherwise, it will try to commit after a rollback is performed.

ramsey:
Does Spot throw an exception if the `DELETE` fails? A few lines after your code, there is a check to see if `$response` is boolean `false`. If an exception is being thrown, I don't think `$response` will be set.

mdwheele:
Ew, looks like it can do both. I'll fix-up. Thanks for the catch (no pun intended). I shouldn't have been working on this late last night haha. 😄

mdwheele:
Aight, should be better now. 

Since it looks like Spot itself sometimes throws booleans out (on failure) and Doctrine's DBAL (and PDO) can/will throw exceptions on failure cases, we have to handle both. It's not pretty, but I set the current `$response` to false if any exception occurs AND if Spot fails, but is not exceptional. After the `try..catch` block, we test `$response` is false, we rollback. Otherwise, we commit; one or the other.

I also rolled the flash messages into that conditional.

chartjes:
Thanks @mdwheele and you can probably save yourself work by not doing it so late at night next time ;)

PR #403, opened 2016-08-21
Upgrading OpenCFP to Silex 2.x

beryllium:
This PR upgrades OpenCFP to work with Silex 2, per a twitter discussion I had with Chris a few months ago.

Numerous changes were made to accommodate the changes in Silex 2 and Pimple 3, such as updating method signatures to reference Pimple 3's Container when appropriate. Also, as Silex 2 services are "shared" by default, several service definitions were able to be simplified.

The tests pass for me, but this change is rather substantial, so I don't know how much trust can be placed on that.

I debated whether I should open the PR at all - until finally deciding that my progress needs feedback before I can spend more time on it.

### **Important Note:**
The original ConfigServiceProvider dependency does not support Silex 2, so I modified composer.json to use a different source (my own fork). 

Additionally, due to the way the container freezes services once instantiated, several places in the tests needed to be updated to "unset" the service before overriding it with a new object.


mdwheele:
Hey @beryllium!

First, thanks a ton for all this work. We really appreciate it!

I intend to merge this, but need a little time to look at it with a closer eye. Passing tests are awesome, but as I'm sure you've noticed, there are plenty of tests that do weird things with the container and mocks. 

In the meantime, some thoughts:

**ConfigServiceProvider**

I've been eye-ing Silex 2.0 and the issue it would cause for the `ConfigServiceProvider` for a while. Namely that the service provider repository looks unmaintained. That said, it's definitely a problem that once-solved, shouldn't need maintenance. My concern is that there are PRs/issues from 2 years ago on this topic with little sign of attention. I will reach out to Igor and see if he's interested in transferring maintenance responsibilities. If so, we can move the project here or wherever it makes sense. Alternatively, if we don't hear back, we have options:

* This might be an opportunity to kill a dependency and just pull in the code required to do this. It's not much if we don't need formats other than YAML. 
* We could continue as-is and rely on your fork. I don't like this option (not because of the work) but because it complicates Composer dependencies. I'd rather see Silex-2.x compatible code go somewhere authoritative, if we don't bring it in. Perfectly fine relying on a fork for now though.

**Remaining Tasks**

You mentioned spending more time on this. What do you have in mind that needs addressing? If it's longer-running, I'd like to move this PR to another branch and work on getting everything straight there. That way, we can review smaller focused work rather than big-bang merge. It's also an opportunity to enumerate things that need to get done so that I can help you with it. That said, after I give this a closer look, I'd like to merge it somewhere rather than continuing to grow it.

Thanks again. Really awesome stuff.

beryllium:
I'm right there with you on the fork approach being non-ideal. That was one of the main reasons I held off submitting the PR. If it could go pure YAML, that would certainly simplify things - and eliminating code is generally a good thing. :)

As for spending more time on it - partly I am worried that the changes to the framework & components might have caused unexpected side effects that might not be surfaced from casual testing. It might help for someone with more experience than myself at OpenCFP administration  to take a look to make sure the switch hasn't borked it up.

On the flip side, I suppose the move-fast-and-break-things ethos might apply here, but I'd suggest making some kind of "stable" release prior to that.

mdwheele:
@beryllium This works for me. I've exercised every feature (lightly) in addition to having tests pass, as expected. I am changing the PR target to `opencfp:silex-2.x` to do some follow-up work on config and I'll plan on tagging you when I open the `silex-2.x` -> `master` PR. If you have any other stuff to send for this, send it to that branch and we'll get it in.

PR #404, opened 2016-08-23
Remove `igorw/config-service-provider`

mdwheele:
This PR removes a dependency on the `ConfigServiceProvider` we use from https://github.com/igorw/ConfigServiceProvider. The service provider's main feature is its ability to load prefixed configuration from multiple formats given a configuration path to load. It also enabled self-referencing replacements using `%replacement%` syntax. We currently only support one format, YAML, and only one configuration file is ever loaded at a time. We have not had a reason to use replacements thus far.

We're removing this because the provider is *barely* incompatible with Silex 2.0 (some typehints need to be changed, basically) and PRs to the repository seem unmanaged.

```
/**
 * Replace reference to `ChainConfigDriver` with `null` when PR below is merged.
 * Symfony's YAML package deprecated allowing a path to be passed to `parse` method.
 * This was causing our test-suite to fail. When this is merged, we can upgrade and
 * all will be well again.
 *
 * @see https://github.com/igorw/ConfigServiceProvider/pull/46
 */
```

That PR was opened a year ago. I still plan to reach out to Igor to see if he's interested in at least merging Silex 2 support, but after thinking about it a bit more, it really doesn't make a whole lot of sense for us right now. We can keep it simpler by just loading the format we support and being done with it.

I'm sending this to `opencfp:silex-2.x` as one of the last remaining things before creating the PR from that branch to `master`. 

- [x] Remove dependency on ConfigServiceProvider
- [x] Update a few Composer packages that have moved

Follows #403.
/cc @beryllium

### DIFF
--- a/classes/Provider/YamlConfigDriver.php
+++ b/classes/Provider/YamlConfigDriver.php
@@ -2,10 +2,9 @@
 
 namespace OpenCFP\Provider;
 
-use Igorw\Silex\YamlConfigDriver as IgorYamlConfigDriver;
 use Symfony\Component\Yaml\Yaml;
 
-class YamlConfigDriver extends IgorYamlConfigDriver
+class YamlConfigDriver
 {
     public function load($filename)
     {
@@ -14,5 +13,10 @@ class YamlConfigDriver extends IgorYamlConfigDriver
         }
         $config = Yaml::parse(file_get_contents($filename));
         return $config ?: [];
+    }
+
+    public function supports($filename)
+    {
+        return (bool) preg_match('#\.ya?ml(\.dist)?$#', $filename);
     }
 }
--- a/classes/Application.php
+++ b/classes/Application.php
@@ -142,20 +142,17 @@ class Application extends SilexApplication
      */
     protected function bindConfiguration()
     {
-        /**
-         * Replace reference to `ChainConfigDriver` with `null` when PR below is merged.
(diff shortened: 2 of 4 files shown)


classes/Provider/YamlConfigDriver.php
```diff
@@ -15,4 +14,9 @@ public function load($filename)
         $config = Yaml::parse(file_get_contents($filename));
         return $config ?: [];
     }
+
+    public function supports($filename)
+    {
+        return (bool) preg_match('#\.ya?ml(\.dist)?$#', $filename);
```
beryllium: This might not be an issue, but maybe you'd want to confirm that the file actually has a name, rather than allowing a file named ".yaml"/".yaml.dist"? I suppose it could be argued that those are both perfectly cromulent filenames.

mdwheele: Will probably leave this as-is. The error will get caught by php `file_get_contents` if isn't an actual file. The filename is derived from the current environment which is a value object ([`Environment`](https://github.com/opencfp/opencfp/blob/master/classes/Environment.php)).

mdwheele:
Leave this unmerged for a bit to allow folks time to pull in 2f747fc (security update). Currently, 2f747fc will go in without a need for `composer install` or any other added work. Adding these on top requires more work on end-user and since this isn't really all that urgent, I think it's worth waiting a few more days to make lives easier!

PR #410, opened 2016-09-15
user:create command 

### LINKED ISSUE #274: [Enhancement] Create admin user through bin/opencfp
## Current Behavior
If one wants to use `bin/opencfp admin:promote` to give a user admin access, the account must already exist. This means a user must go to the CFP website and create an account before they can be made an admin in the system.

## Suggested Enhancement
CLI users should be able to promote an email address to admin even if it isn't known to the system yet.

There are two ways to add this to the script:
 1. Update `admin:promote` to create an account if the supplied email isn't found.
 2. Add `user:create` to allow users to be created on the command line, which could then be promoted with `admin:promote`.

I prefer option 2 because it has more utility. It's very possible that an admin could want to create a user on the command line that isn't an administrator (for example, during development). #1 is a shortcut for the scenario in this ticket, but it doesn't provide any flexibility.

### Considerations
The newly created account will have to have a password. The CLI user could create a password right there, but the most secure (and slickest) way would be to create a one-time password that had to be changed on first login. That, of course, is more work, and I don't personally think it needs to be a blocker in implementing this feature.

First Name and Last Name are currently required on the registration form. Either they'll have to be passed into the script, or someone will have to make sure it won't break something to leave them blank.

bretterer:
Add ability to create a user from the command line.  It also has a flag to auto promote to admin.

bin/opencfp user:create --first_name=Brian --last_name=Retterer --email="bretterer@gmail.com" --password="123456abcdef" --admin

Fixes #274

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ OpenCFP is a PHP-based conference talk submission system.
    * [Using the API](#json-api-usage)
  * [Command-line Utilities](#command-line-utilities)
    * [Admin Group Management](#admin-group-management)
+   * [User Management](#user-management)
    * [Clear Caches](#clear-caches)
    * [Scripts to Rule Them All](#scripts-rule-all)
  * [Testing](#testing)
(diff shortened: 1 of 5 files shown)


chartjes:
Thanks @bretterer

Issue #396, opened 2016-06-06
Better default talk categories


modernserf:
The default talk categories are kind of peculiar: are there really a lot of talks on the IBMi platform? Given that a lot of confs using OpenCFP leave the default talk categories, these could be updated to be more universally applicable.

carnage:
Hi, given the above pull request allows much easier modification of categories; would you consider this issue resolved?

PR #411, opened 2016-10-04
Moves talk categories into config to enable easier modification

carnage:
This PR

* [x] Moves the specification of categories into the config file to allow users to change them once, in one place.

Relates to: #396


chartjes:
Thanks @carnage

PR #412, opened 2016-10-13
Updated templates to use display name for categories

carnage:
This PR

* [x] Updates templates to use display name for categories


chartjes:
Thanks @carnage

PR #414, opened 2016-10-14
Fixes #391

### LINKED ISSUE #391: Default sort order on talk review page
Quick question / quirk:

In [/classes/Domain/Entity/Mapper/Talk.php](https://github.com/opencfp/opencfp/blob/master/classes/Domain/Entity/Mapper/Talk.php#L156) the default sort order for the Review Talks page (aka top rated) seems to be `ORDER BY rating DESC, total_rating DESC`. Is this intended? Wouldn't that make the first vote for any given talk be the primary sort condition?

We're seeing an unexpected order of talks on this page and I suspect it's because of this.

One caveat of our usage however, we modified the review page to include all votes (+ and - 1s) where normally that page would just have the `WHERE rating = 1` clause, we have `WHERE rating != 0`

1. Get Bunches of talk submissions
2. Get a bunch of votes on talks by admins
3. Go to the top rated page

# Expected Result

* Talks are listed from top to bottom depending on what their final vote totals are

# Actual Result

* Talks are listed out of order

carnage:
This PR

* [x] Changes the review menu item name to Top rated talks (to match page heading and function)
* [x] Includes all ratings in total, not just positive ones.

Previous behaviour only listed out positive votes on a talk; showing the total vote count as equal to the rating. This could lead to erroneous results. The new behaviour only filters our talks after the rating has been totalled, so that negative votes affect the result and ratings count as well.

Fixes #391.

### DIFF
--- a/classes/Domain/Entity/Mapper/Talk.php
+++ b/classes/Domain/Entity/Mapper/Talk.php
@@ -167,8 +167,8 @@ class Talk extends Mapper
         $talks = $this->query(
             "SELECT t.*, SUM(m.rating) AS total_rating, COUNT(m.rating) as review_count FROM talks t "
             . "LEFT JOIN talk_meta m ON t.id = m.talk_id "
-            . "WHERE rating > 0 "
             . "GROUP BY m.`talk_id` "
+            . "HAVING total_rating > 0 "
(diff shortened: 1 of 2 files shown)


chartjes:
Thanks @carnage

PR #419, opened 2016-10-20
Fix issue causing unit test failiures

carnage:
This PR

* [x] Updates testing.yml to fix the unit test failures


chartjes:
Thanks @carnage

PR #420, opened 2016-10-20
Revert "Revert "Fixes #391""

### LINKED ISSUE #391: Default sort order on talk review page

carnage:
Reverts opencfp/opencfp#417

Unit test failure had nothing to do with this PR - fixed separatly

### DIFF
--- a/classes/Domain/Entity/Mapper/Talk.php
+++ b/classes/Domain/Entity/Mapper/Talk.php
@@ -167,8 +167,8 @@ class Talk extends Mapper
         $talks = $this->query(
             "SELECT t.*, SUM(m.rating) AS total_rating, COUNT(m.rating) as review_count FROM talks t "
             . "LEFT JOIN talk_meta m ON t.id = m.talk_id "
-            . "WHERE rating > 0 "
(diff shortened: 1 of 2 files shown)


chartjes:
Thanks @carnage

PR #421, opened 2016-10-20
Revert "Revert "Move talk lengths to config""

carnage:
Reverts opencfp/opencfp#418

Unit test failure had nothing to do with this PR - fixed separatly


chartjes:
Thanks @carnage

Issue #424, opened 2016-11-08
CFP is shown as closed on homepage before end date with a different timezone


cgrandval:
# Steps required to reproduce the problem

1. Define a timezone like America/Recife (- 4 hours) for Brazilian event (not UTC)
2. Define today as the "enddate" in the config file
3. Modify your local time after 20h (with America/Recife timezone used)
4. Go to homepage

# Expected Result

* On the homepage, the "Register now" button must be available and the message must be "Call For Papers Now Open! Submissions accepted until XXX"

# Actual Result

* The CFP is not closed (and you can subscribe and submit a talk correctly) but on homepage, the button is not available and the message says CFP is closed.

# Problem
The timezone is defined in the Application.php after TwigServiceProvider is registered. But this provider defines the "cfp_open" variable from 2 dates comparaison (now and the end date). Without the timezone defined, it uses UTC.

PR #425, opened 2016-11-08
Move timezone set before register TwigServiceProvider

cgrandval:
This PR resolves the issue #424


mdwheele:
Looks good. Thanks!

Issue #426, opened 2016-11-14
Arrival and departure dates are displayed with one day before on homepage


cgrandval:
# Steps required to reproduce the problem

1. Define an arrival and departure dates in the configuration (development.yml or production.yml), for instance 17/12/2016 and 20/12/2016
2. Go to the homepage

# Expected Result

* The message displayed must be "OpenCFP is on 17/12/2016 to 20/12/2016 in Miami, FL"

# Actual Result

* The message is "OpenCFP is on 16/12/2016 to 19/12/2016 in Miami, FL"

PR #427, opened 2016-11-14
Fix arrival and departures displayed dates in the header

cgrandval:
This PR resolves the issue #426 

If the date is already a DateTime object, and if you want to keep its current timezone, pass false as the timezone value : http://twig.sensiolabs.org/doc/filters/date.html#timezone


mdwheele:
Thankss @cgrandval!

PR #429, opened 2016-12-16
Not Rated Talks Query is Incorrect

mwillbanks:
The not rated talks attempts to create a left join but in the where
statement attempts to look at the admin user id.  Unfortunately, this
will cause any talks that are not yet in talk_meta with the admin user
id to not be found.

The proper join query includes the admin user id as that is what you are
attempting to key off of rather than in the where portion of the query.
This will resolve where the talk does not yet exist in talk_meta and
will provide not yet rated talks properly in the admin area.


carnage:
hah, I've literally just encountered and fixed this myself.

The not viewed by seems to exhibit the same issue

chartjes:
Thanks @mwillbanks and @carnage

PR #432, opened 2016-12-26
allow users to tag sessions

jsundquist:
This pull request adds a way for users to add tags / keywords to their sessions. This provides a way for event organizers to group talks better then just in a single track.


classes/Domain/Entity/TalkTag.php
```diff
@@ -0,0 +1,26 @@
...
+use Spot\Entity;
+
+class TalkTag extends Entity
+{
+    protected static $table = 'talks_tags';
+
+    public static function fields()
+    {
+        return [
+            'id'        => ['type' => 'integer', 'primary' => true, 'autoincrement' => true],
```
localheinz: Can we consistently unalign `=>` here? Seems a bit off!

classes/Http/Controller/TalkController.php
```diff
@@ -170,12 +171,13 @@ public function editAction(Request $req)
...
 
         $talk_mapper = $spot->mapper(\OpenCFP\Domain\Entity\Talk::class);
-        $talk_info = $talk_mapper->get($talk_id)->toArray();
+        $talk_info = $talk_mapper->where(['id' => $talk_id])->with(['tags'])->execute()->first()->toArray();
 
         if ($talk_info['user_id'] !== (int) $user->getId()) {
             return $this->redirectTo('dashboard');
         }
 
+
```
localheinz: Can we remove this extra empty line here?

classes/Http/Controller/TalkController.php
```diff
@@ -191,6 +193,10 @@ public function editAction(Request $req)
             'slides' => $talk_info['slides'],
             'other' => $talk_info['other'],
             'sponsor' => $talk_info['sponsor'],
+            'tags' => array_reduce($talk_info['tags'], function ($response, $item) {
+
```
localheinz: Can we remove this extra empty line here?

classes/Http/Controller/TalkController.php
```diff
@@ -191,6 +193,10 @@ public function editAction(Request $req)
             'slides' => $talk_info['slides'],
             'other' => $talk_info['other'],
             'sponsor' => $talk_info['sponsor'],
+            'tags' => array_reduce($talk_info['tags'], function ($response, $item) {
```
localheinz: Can we add a type-hint for `$item`? Looks like it is either an `array` or an object implementing `ArrayAccess`, doesn't it?

classes/Http/Controller/TalkController.php
```diff
@@ -291,6 +299,17 @@ public function processCreateAction(Request $req)
 
         if ($isValid) {
             $sanitized_data = $form->getCleanData();
+            $tags = explode(',', $sanitized_data['tags']);
+
+            $tag_entity = [];
```
localheinz: How about using `array_map()` here instead, and also, maybe importing `OpenCFP\Domain\Entity`, as well as using plural for the variable name?

```php
$tag_entities = array_map(function ($tag) {
    return new Entity\Tag([
        'tag' => trim($tag),
    ]);
}, $tags);
```

Alternatively, maybe just inline `$tags` and rename `$tag_entity` to `$tags`?

classes/Http/Controller/TalkController.php
```diff
@@ -304,20 +323,25 @@ public function processCreateAction(Request $req)
...
 
-            $talk_mapper = $spot->mapper(\OpenCFP\Domain\Entity\Talk::class);
-            $talk = $talk_mapper->create($data);
+            $talk->relation('tags', new Collection($tag_entity));
+
+            try {
+                $talk_data = $talk_mapper->save($talk, ['relations' => true]);
+            } catch (\Exception $e) {
+                echo $e->getMessage();
+                die();
```
localheinz: Ha, should we handle an exception differently here?

classes/Http/Controller/TalkController.php
```diff
@@ -380,7 +406,17 @@ public function updateAction(Request $req)
 
         if ($isValid) {
             $sanitized_data = $form->getCleanData();
-            $updated_at = new \DateTime();
+            $tags = explode(',', $sanitized_data['tags']);
+
+            $tag_entity = [];
```
localheinz: Similar to above, maybe turn this into static method on `Entity\Tag`?

```php
/**
 * @param string[] $tags
 * 
 * @throws \InvalidArgumentException
 *
 * @return static[]
 */   
public static function fromTags(array $tags)
{
    return array_map(function ($tag) {
        if (!is_string($tag)) {
             throw new \InvalidArgumentException('Tag needs to be a string');
        }

        $tag = trim($tag);

        if ($tag === '') {
             throw new \InvalidArgumentException('Tag needs to be a non-blank string');
        }

        return new Entity\Tag([
            'tag' => $tag,
        ]);
    }, $tags);
}
```

classes/Http/View/TalkHelper.php
```diff
@@ -74,4 +74,15 @@ public function getLevelDisplayName($level)
...
+
+    public function getTags($input)
+    {
+        $tags = [];
+
+        foreach ($input as $tag) {
+            $tags[] = $tag->tag;
+        }
+
+        return implode(', ', $tags);
```
localheinz: How about using `array_map()` here?

```php
return implode(', ', array_map(function (Entity\Tag $tag) {
    return $tag->tag;
}, $tags));
```

classes/Http/Controller/TalkController.php
jsundquist: Not quiet sure. Interestingly enough anywhere an exception is thrown, they are always just echoed out and died.

classes/Http/Controller/TalkController.php
jsundquist: Actually already working on refactoring a it a bit to convert it from an inline code to a method.  I don't know if an exception really needs to be thrown if a user does not provide tags for there talks.  It's not a required field at this point in time.

chartjes:
Thank you @jsundquist for the PR and @localheinz for the review

PR #433, opened 2016-12-26
Enhancement: Update friendsofphp/php-cs-fixer

localheinz:
This PR

* [x] removes `fabpot/php-cs-fixer` and requires `friendsofphp/php-cs-fixer`
* [x] updates `.travis.yml` and `.php_cs`
* [x] runs `php-cs-fixer`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/compare/v1.11.2...v2.0.0.

PR #434, opened 2016-12-26
Fix: Keep require and require-dev sections close to each other

localheinz:
This PR

* [x] keeps the `require` and `require-dev` sections in `composer.json` close to one another

Not exactly related to #433.

PR #435, opened 2016-12-26
Fix: Keep packages sorted

localheinz:
This PR

* [x] keeps packages in `composer.json` sorted without having to specify `--sort-packages` when requiring packages

Also not directly related to #433.

💁‍♂️ For reference, see https://getcomposer.org/doc/06-config.md#sort-packages.


composer.json
```diff
@@ -1,41 +1,44 @@
...
     "license": "MIT",
+    "config": {
+        "sort-packages": true
+    },
     "require": {
         "php": ">=5.5",
-        "twig/twig": "1.*",
-        "pagerfanta/pagerfanta": "1.0.*@dev",
+        "aptoma/twig-markdown": "0.2.*",
         "cartalyst/sentry": "2.1.*",
```
localheinz: I added the `config` section above, then I ran

```
$ composer require "cartalyst/sentry:2.1.*"
```

to trigger the sorting, as this is the only package that isn't outdated.

```
$ composer info --direct | grep cartalyst/sentry
cartalyst/sentry              v2.1.6             PHP 5.3+ Fully-featured Authentication & Authorization System

$ composer outdated | grep cartalyst/sentry

Package fabpot/php-cs-fixer is abandoned, you should avoid using it. Use friendsofphp/php-cs-fixer instead.
Package guzzle/guzzle is abandoned, you should avoid using it. Use guzzlehttp/guzzle instead.
```

chartjes:
@localheinz There's a conflict with the composer.lock file that needs to be resolved...

localheinz:
@chartjes 

Let me sort this out once #434 has been merged!

localheinz:
Rebased!

localheinz:
Thanks for reviewing all the PRs, @chartjes!

PR #436, opened 2016-12-26
Fix: Use assertCount() instead of assertTrue()

localheinz:
This PR

* [x] uses more appropriate `assertCount()` instead of `assertTrue()`


chartjes:
Thanks @localheinz

PR #437, opened 2016-12-26
Fix: Class should probably be abstract

localheinz:
This PR

* [x] turns `BaseCommand` into an `abstract` class


chartjes:
Asking @mdwheele for a review

PR #438, opened 2016-12-26
Fix: Remove accessor only used in test

localheinz:
This PR

* [x] removes an accessor which is only used in a test

Follows #437.


chartjes:
Thanks @localheinz

PR #440, opened 2016-12-26
Enhancement: Reference XML schema for phpunit.xml

localheinz:
This PR

* [x] references the corresponding schema for `phpunit.xml`

💁‍♂️ This helps with auto-completion when modifying the configuration file!


chartjes:
Thanks @localheinz

PR #445, opened 2017-01-26
Remove rating when repeatedly clicking thumbs up/down

nb:
This PR

* [x] If I click thumbs up on an already thumbs-upped talk removes the rating.
* [x] If I click thumbs down on an already thumbs-downed talk removes the rating.

Rationale: for [our event](http://apply.wp-europe.org/) we would like to split scoring the talks – some people only rate development talks, some people only rate development ones, etc. Sometimes, however, we rate the wrong talk, one that we shouldn't have rated at all and being able to remove the rating would be super nice.


chartjes:
Thanks @nb!

PR #447, opened 2017-01-28
Fixes double encoding of &, < and >

### LINKED ISSUE #444: In the usual implementations an ampersand is output as "&amp;" and not as "&"
## Steps required to reproduce the problem

1. Create a speaker-profile
2. Enter "wdv Gesellschaft für Medien & Kommunikation mbH & Co. OHG" in the company-field
3. Save

## Expected Result

* In the profile I see "Company: wdv Gesellschaft für Medien & Kommunikation mbH & Co. OHG"

## Actual Result

* The Profile shows "Company: wdv Gesellschaft für Medien &amp; Kommunikation mbH &amp; Co. OHG"

bildschirmfoto 2017-01-25 um 09 45 11

The Screenshot has been taken on the PHPKonfs OpenCFP-Implementation but I experienced that up to now on all OpenCFP-implementations I've used. So it seems to be a common issue.

heiglandreas:
This PR 

- [x] fixes an issue where &, < and > are encoded to ```&amp;```, ```&lt;``` and ```&gt;``` which will be encoded to ```&amp;mp;```, ```&amp;lt;``` and ```&amp;gt;``` resp.

The introduced code revokes the encoding of the three characters after sanitization.

Fixes #444.

### DIFF
--- a/classes/Provider/Gateways/RequestCleaner.php
+++ b/classes/Provider/Gateways/RequestCleaner.php
@@ -32,8 +32,11 @@ class RequestCleaner
             if (is_array($value)) {
                 $sanitized[$key] = $this->clean($value);
             } else {
-                $sanitized[$key] = $this->purifier->purify($value);
-                ;
+                $sanitized[$key] = preg_replace(
+                    ['/&amp;/', '/&lt;\b/', '/\b&gt;/'],
+                    ['&', '<', '>'],
+                    $this->purifier->purify($value)
+                );
             }
         }
 


classes/Provider/Gateways/RequestCleaner.php
```diff
@@ -32,8 +32,11 @@ private function clean(array $data)
             if (is_array($value)) {
                 $sanitized[$key] = $this->clean($value);
             } else {
-                $sanitized[$key] = $this->purifier->purify($value);
-                ;
+                $sanitized[$key] = preg_replace(
+                    ['/&amp;/', '/&lt;\b/', '/\b&gt;/'],
```
nb: What cases does the `\b` usage cover? And why do we have it only on `<` and `>`, but not `&`?

chartjes:
Thanks @heiglandreas for the PR

PR #448, opened 2017-01-31
Fix missing `BadMethodCallException` by updating Spot 2 dependency

nb:
Context: there was a missing property in a twig template and Spot was trying to throw `BadMethodCallException`, however the class wasn’t declared. It was [fixed some time ago](https://github.com/vlucas/spot2/commit/0c5baa035f90f34434e10cac0e1cf53f90d1408e#diff-d1cd5c83ba49d43232655b056fad3936R4), so we should update to a newer version.

Since we were depending on `dev-master` the easiest way to update Spot 2
was to update all dependencies via `composer update`.

Tests pass and I tested the main flows – registration, posting a new talk, voting, they all seem to work.

This PR

* [x] Updated dependencies.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3568,22 +3568,22 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/vlucas/spot2.git",
-                "reference": "2e9acb2f4115bd386d9f2f9446c799128f283d92"
+                "reference": "4abb1bd1ef13741c38fc56f7f3d634b779b19eef"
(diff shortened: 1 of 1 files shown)


nb:
Hm, I will look at the test failures in a bit.

nb:
I switched to updating only Spot 2, not all of the packages, let’s see how it goes.

chartjes:
Thanks @nb

PR #449, opened 2017-01-31
Account for possibly missing `photo_path`

nb:
If a user doesn’t have a photo we are getting a Twig exception when trying to view their comments below a talk:

```
An exception has been thrown during the rendering of a template
("Method Spot\Relation\BelongsTo::photo_path does not exist")
```

This PR

* [x] Switches to `.get` in the Twig template.


chartjes:
Thanks @nb!

PR #451, opened 2017-02-10
fix enddate

### LINKED ISSUE #450: Clean up end date code - add support for time & timezone

1. Be a developer who cares about dates and times
2. Grep the codebase for how `enddate` is used (https://github.com/opencfp/opencfp/search?utf8=%E2%9C%93&q=enddate)
3. Hang head in shame

# Expected Result

* consistently use the enddate config value and ideally manage it in a central place
* support a timezone in enddate
* support a time as well if someone wants to end at 5pm or something

# Actual Result

* in some places the code uses strtotime with 11:59 appended, other places new DateTime objects are created without time
* there is no way to specify a timezone without breaking the code
* no way to specify time without breaking code

heiglandreas:
This PR moves calls from ```strtotime``` to ```DateTime```.

Additionally the configuration entry ```enddate``` can now be used to also add a time when the CfP ends. You should then also adapt the ```date_format``` to include the time. When no time is given in the config-value the current behaviour is used to let the CfP run until 23:59 of the given day. 

> Note: The CFP will still end 59 seconds before midnight!

* [x] replace ```strtotime``` with ```DateTime```

Fixes #450.

### DIFF
--- a/classes/Http/Controller/DashboardController.php
+++ b/classes/Http/Controller/DashboardController.php
@@ -35,20 +35,24 @@ class DashboardController extends BaseController
     /**
      * Check to see if the CfP for this app is still open
      *
-     * @param integer $currentTime
+     * @param integer|\DateTimeInterface $currentTime
      *
      * @return boolean
      */
     public function isCfpOpen($currentTime = null)
     {
-        if (!$currentTime) {
-            $currentTime = strtotime('now');
+        if (! $currentTime) {
+            $currentTime = new \Datetime();
         }
 
-        if ($currentTime < strtotime($this->app->config('application.enddate') . ' 11:59 PM')) {
-            return true;
+        if (! $currentTime instanceof \DateTimeInterface) {
+            $currentTime = new \DateTime('@' . $currentTime);
         }
 
-        return false;
+        if ($currentTime > new \DateTime($this->app->config('application.enddate') . ' 11:59 PM')) {
+            return false;
+        }
+
+        return true;
     }
 }
--- a/classes/Http/Controller/DashboardController.php
+++ b/classes/Http/Controller/DashboardController.php
@@ -49,7 +49,12 @@ class DashboardController extends BaseController
             $currentTime = new \DateTime('@' . $currentTime);
         }
 
-        if ($currentTime > new \DateTime($this->app->config('application.enddate') . ' 11:59 PM')) {
+        $enddate = new \DateTime($this->app->config('application.enddate'));
+        if ($enddate->format('H:i:s') == '00:00:00') {
+            $enddate->add(new \DateInterval('PT23H59M'));
+        }
+
+        if ($currentTime > $enddate) {
             return false;
         }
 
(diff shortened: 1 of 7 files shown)


heiglandreas:
I tested it and it looked good but please - someone should give it a second go!

classes/Http/Controller/DashboardController.php
```diff
@@ -35,20 +35,29 @@ public function showSpeakerProfile()
...
+            $currentTime = new \DateTime('@' . $currentTime);
         }
 
-        return false;
+        $enddate = new \DateTime($this->app->config('application.enddate'));
+        if ($enddate->format('H:i:s') == '00:00:00') {
+            $enddate->add(new \DateInterval('PT23H59M'));
+        }
+
+        if ($currentTime > $enddate) {
```
pmeth-delvia: how about just
```
return $currentTime <= $enddate;
```

classes/Http/Controller/DashboardController.php
```diff
@@ -35,20 +35,29 @@ public function showSpeakerProfile()
     /**
      * Check to see if the CfP for this app is still open
      *
-     * @param integer $currentTime
+     * @param integer|\DateTimeInterface $currentTime
```
heiglandreas: It's a leftover from the current implementation. I wanted to keep that to not break stuff I didn't find. But you're right. I should remove that and actually break the code to find currently unchanged implementations.

classes/Http/Controller/DashboardController.php
```diff
@@ -35,20 +35,29 @@ public function showSpeakerProfile()
...
 
-        if ($currentTime < strtotime($this->app->config('application.enddate') . ' 11:59 PM')) {
-            return true;
+        if (! $currentTime instanceof \DateTimeInterface) {
+            $currentTime = new \DateTime('@' . $currentTime);
         }
 
-        return false;
+        $enddate = new \DateTime($this->app->config('application.enddate'));
+        if ($enddate->format('H:i:s') == '00:00:00') {
```
heiglandreas: Yesterday evening I couldn't find a different solution. Perhaps checking the microseconds might be a way… I'd rather not add an ```endtime```-field in the config. That would make parsing more complicated and scatter the same stuff over two config-entries while it can stay in one…

classes/Http/Controller/DashboardController.php
```diff
@@ -35,20 +35,29 @@ public function showSpeakerProfile()
...
+            $enddate->add(new \DateInterval('PT23H59M'));
```
heiglandreas: That's right. But as it was the current behavior I didn't want to break it as I couldn't see whether that's a wanted feature or not. If that was a bug all the time it's more than easy to change it here :)

classes/Http/Controller/DashboardController.php
heiglandreas: Good idea! Makes it much cleaner!

heiglandreas:
Using the CallForProposal-Class sounds good! I'll need to add a second method though that calculates the end-date from the config-value though.

chartjes:
Thanks @heiglandreas

PR #455, opened 2017-02-13
Update Requirements Section in README

zanderbaldwin:
Update the requirements section to specify installing the PHP7+ `intl` extension (same version as the minimum requirement for PHP).

`php7.0-intl` is the name of the package in *Debian Testing* and *Ubuntu 16.04+*, I assumed these were the target platforms, please correct me if I'm wrong.


chartjes:
Thanks for the PR @zanderbaldwin

PR #464, opened 2017-03-29
Fix: Run php-cs-fixer on Travis

localheinz:
This PR

* [x] sets `WITH_CS` to `true` so `php-cs-fixer` is run on Travis
* [x] runs `make cs`

Follows #453.


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #466, opened 2017-03-29
Fix: Use Elvis operator where possible

localheinz:
This PR

* [x] uses the Elvis operator where possible


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #467, opened 2017-03-29
Fix: Use null coalescing operator where possible

localheinz:
This PR

* [x] uses the null coalescing operator where possible


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #474, opened 2017-04-12
Get path to HTML Purifier cache from app

### LINKED ISSUE #472: [Cache] Configure path for HTMLPurifier cache
Without a configured path the HTMLPurifier wants to create a cache under `vendor/ezyang/htmlpurifier/library/HTMLPurifier/DefinitionCache/Serializer` and to chmod-it to `0777`. It is even mentioned in the [Troubleshooting](https://github.com/opencfp/opencfp#troubleshooting).

The code actually supports defining a custom path for the HTMLPurifier cache: https://github.com/opencfp/opencfp/blob/44bad694bbb492eef4b545019324761d5712f381/classes/Provider/HtmlPurifierServiceProvider.php#L21

I propose to update the `production.yml.dist` and `development.yml.dist` files adding in the `paths.cache.purifier` section:

``` yaml
paths:
  cache:
    purifier: /install/path/of/opencfp/cache
```

as well as adding a small section to the README file explaining that this path should be set.

This way when installing a new OpenCFP instances users won't have to look into the code how to setup the cache, nor will there be a need for troubleshooting the cache any more.

robertbasic:
The app has a reusable method that tells us where
HTML Purifier should put the cache files.

Use it instead of trying to get the cache path from the config.

Fixes #472

### DIFF
--- a/classes/Provider/HtmlPurifierServiceProvider.php
+++ b/classes/Provider/HtmlPurifierServiceProvider.php
@@ -18,7 +18,7 @@ class HtmlPurifierServiceProvider implements ServiceProviderInterface
             if ($app->config('cache.enabled')) {
                 $cachePermissions = 0755;
                 $config->set('Cache.SerializerPermissions', $cachePermissions);
-                $cacheDirectory = $app->config('paths.cache.purifier');
+                $cacheDirectory = $app->cachePurifierPath();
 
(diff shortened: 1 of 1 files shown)


chartjes:
Thanks for the PR @robertbasic!

PR #486, opened 2017-09-23
print correct not-found filename on error

tobru:
This shows the correct path of the configuration file which couldn't be found.


mdwheele:
👍 Thanks!

PR #488, opened 2017-09-25
Add tests

BackEndTea:
Added unit tests for the ProfileController and the TalkController.


chartjes:
@BackEndTea Thanks so much for your contribution!

PR #489, opened 2017-09-26
Added joind.in url to satisfy issue #350.

### LINKED ISSUE #350: Dedicated field for speakers to enter their joind.in url
Lots of speakers exclude it or bury it or only include links in their 'more info' sections of talks which they've done before and include a link to a specific talk which is rather annoying when you want to see it in the context of all their talks.

Having a dedicated field for this would help encourage people to provide an easy to use joind.in link to their speaker profile (e.g. https://joind.in/user/chartjes)

funzie19:
This PR

Adds the ability for a speaker to enter their joind.in page when they sign up or edit their profile.

Fixes #350


chartjes:
@funzie19 Thanks so much for the addition

PR #492, opened 2017-09-28
Allow sign up without joind.in url

### LINKED ISSUE #491: Should joind.in URL be required to create Speaker Profile?

1. Visit the Signup Page
2. Fill out email, password, name; leaving joind.in empty
3. Agree to CoC
4. Submit

# Expected Result

* My profile should be created and I should be able to submit talks

# Actual Result

image

This was likely unintentional. If someone else doesn't grab it, I'll send a PR as soon as I finish some work I'm doing on the test suite to hopefully make it less stressful.

BackEndTea:
Changed the check for the joind.in url so users are now allowed to sign up with either a valid url or no url at all.

Fixes #491.

### DIFF
--- a/classes/Http/Form/SignupForm.php
+++ b/classes/Http/Form/SignupForm.php
@@ -249,7 +249,10 @@ class SignupForm extends Form
 
     public function validateUrl()
     {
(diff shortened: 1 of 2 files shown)


mdwheele:
Thanks!

PR #493, opened 2017-09-28
[WIP] Improve readability of test-suite

mdwheele:
This PR represents a direction we **could** take our tests in to clean things up. Personally, I have a hard time reading / understanding what behaviours some of the tests are guaranteeing. Controller tests are about the worst of them because there is a lot of setup / teardown that has to happen. So, that said, this is work in progress to give an idea. 

I updated `DashboardControllerTest` to use the helpers as much as possible. Question is: Did it improve anything?

*Full disclosure, these helpers are inspired or straight-up ripped from Laravel Framework. I focused mostly on the request / response helpers, but I also think improvements could be made to test factories, swapping authenticated users (`$this->actingAs('user', 'role')`) and asserting changes made to database.* 

**Other Examples**

```php
/**
 * @test
 */
public function the_homepage_shows_cfp_is_open()
{
    $this->callForPapersIsOpen()
        ->get('/')
        ->assertSee('Call For Papers Now Open!');
}
```

```php
/**
 * @test
 */
public function the_homepage_shows_cfp_is_closed()
{
    $this->callForPapersIsClosed()
        ->get('/')
        ->assertSee('Call for Papers has ended!');
}
```

```php
/**
 * @test
 */
public function profiles_are_only_editable_by_the_owner()
{
    $speaker = $this->make('speaker');
    $otherSpeaker = $this->make('speaker');
    
    $this->actingAs($speaker)
        ->get("/profile/edit/{$otherSpeaker->id}")
        ->assertSee('Unauthorized');
}
```

I also toyed with implementing something like the following to receive the route name instead of URI. I dropped it for now. Could look like:

```php
/**
 * @test
 */
public function profiles_are_only_editable_by_the_owner()
{
    $speaker = $this->make('speaker');
    $otherSpeaker = $this->make('speaker');
    
    $this->actingAs($speaker)
        ->route("user_edit", $otherSpeaker->id)
        ->assertSee('Unauthorized');
}
```


mdwheele:
The other thing I'm looking to cut down on is the amount of setup we have to do per test method. Specifically, any time we mess around with changing the authenticated user, that's like 20 lines blocking what behaviour we're actually testing.

PR #494, opened 2017-10-01
Allow first names containing non-english chars

dunglas:
My first name is Kévin (notice the `é`), and currently I cannot register on OpenCFP. This PR handles first names like last names, and will allow people with non-english characters in their names to submit CFPs with their real names.


chartjes:
Thanks @dunglas

PR #497, opened 2017-10-02
Add speaker notes to the speaker profile

### LINKED ISSUE #496: Additional speaker info isn't displayed

1. Create a speaker account
2. Input something in additional notes section
3. Look at the speaker profile as an admin

# Expected Result

* When the profile is displayed I can see the additional notes
image

# Actual Result

image

image

It gets saved in the database just fine, its just that there is no part in the speakers view for the admin that actually displays it.

BackEndTea:
This will allow an admin to see the additional notes from a speaker, if they are set.

Fixes #496 .


chartjes:
Thanks @BackEndTea

PR #499, opened 2017-10-05
Fixes for Sentinel branch

mdwheele:
I was unable to login when on the `ch-symfony-security-refactor` branch for a number of reasons:

1. Couldn't install dependencies because `composer.lock` was corrupt / invalid somehow
1. For existing users (from Sentry), Sentinel did not think they were activated because there was not a "completed" entry in the `activations` table. I wrote a migration that does that for all current users.
1. Application failed to login because `persistences` table did not exist. I created a migration for that.

I'm sending this and moving forward.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -357,16 +357,16 @@
         },
         {
             "name": "doctrine/cache",
-            "version": "v1.6.1",
+            "version": "v1.6.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/cache.git",
-                "reference": "b6f544a20f4807e81f7044d31e679ccbb1866dc3"
+                "reference": "eb152c5100571c7a45470ff2a35095ab3f3b900b"
             },
             "dist": {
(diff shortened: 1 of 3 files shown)


mdwheele:
Worth noting that apparently, a good number of Doctrine projects updated minimum version of PHP to 7.1 so if running Composer on 7.1 to commit the lockfile (with updated versions requiring 7.1), TravisCI complains it can't install those packages because we use 7. 

For now, I switched back to PHP 7.0 locally. Another option is bumping our minimum support version of PHP to 7.1 as well.

mdwheele:
lol... github is confused. closing this to re-open clean.

chartjes:
@mdwheele Travis is complaining cuz some of the tests are busted. Honestly, this branch should just be blown away and started from scratch because I was adding too many things into it outside of the stated goal of refactoring away from Sentry

mdwheele:
I can take a stab at that tonight.

PR #500, opened 2017-10-05
Allow searching for specific speakers

BackEndTea:
As the system gets more and more users it can become a bit annoying to find a specific speaker to either promote them to admin or check out their talks.

Right now it just searches through their first and last name, i'm not sure if it should be expanded to also include things like their bio/email/twitter etc.

### DIFF
--- a/classes/Http/Controller/Admin/SpeakersController.php
+++ b/classes/Http/Controller/Admin/SpeakersController.php
@@ -29,11 +29,23 @@ class SpeakersController extends BaseController
         /* @var Locator $spot */
         $spot = $this->service('spot');
 
+        $search = $req->get('search');
+
+
+
         $rawSpeakers = $spot
             ->mapper(\OpenCFP\Domain\Entity\User::class)
-            ->all()
-            ->order(['first_name' => 'ASC'])
-            ->toArray();
+            ->all();
+
+        if ($search != '' && $search != null) {
+            $rawSpeakers = $rawSpeakers
+                ->where(['first_name :like' => $search])
+                ->orWhere(['last_name :like' => $search]);
+        }
+
+        $rawSpeakers = $rawSpeakers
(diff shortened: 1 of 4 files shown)


classes/Http/Controller/Admin/SpeakersController.php
```diff
@@ -32,11 +32,21 @@ public function indexAction(Request $req)
...
+        $search = $req->get('search');
+
         $rawSpeakers = $spot
             ->mapper(\OpenCFP\Domain\Entity\User::class)
-            ->all()
-            ->order(['first_name' => 'ASC'])
-            ->toArray();
+            ->all();
+
+        if (''!= $search && null != $search) {
```
chartjes: Please don't use "Yoda conditionals" in the code.

BackEndTea:
I have fixed the Yoda conditionals. I'm not entirely sure how to go about testing this though.

chartjes:
@BackEndTea If I were writing a test to go with this code, I think it would be something like this:

* create a DB table with the list of speakers in it
* add the 'search' variable to the request
* execute that controller method (should be lots of examples of doing this in our code base)
* verify in the output that you see only the names of the speakers you are expecting

BackEndTea:
I'll wait for #501 since that makes some changes to the database tests before adding the tests. So i'm closing this for now

BackEndTea:
@chartjes I changed it around a little bit and moved this functionality to the User Mapper. It makes more sense for it to be there.

Plus i also wrote the unit tests for it.

mdwheele:
@BackEndTea I can help resolve this, if you need. I just did you a dirty and sent a pretty rediculously-large PR.

BackEndTea:
@mdwheele I don't mind fixing it myself, but i probably won't get around to that until monday

chartjes:
Thanks for making the changes @BackEndTea

Issue #498, opened 2017-10-04
[RFC] Updating from Cartalyst/Sentry to Cartalyst/Sentinel


chartjes:
Sentry has been deprecated, and using a deprecated library worries me. I did attempt to upgrade it but I got very unfocused and decided to also add test coverage while I did it.

The work did not require a lot of code changes but extensive DB schema changes were needed. These make me uncomfortable too.

The branch in question is ch-symfony-security-refactor for people who want to look at what I had been doing. I am unsure on how to proceed, looking for some advice and structure.

mdwheele:
Mobile at the moment, so I can't comment in-depth. When I made an attempt at this, one of the most problematic issues was that resolving identity happens through multiple interfaces. Some controllers reference Sentry directly to grab the user and others use the IdentityProvider interface. 

Stubbing an IdentityProvider for tests seems much lighter than all the setup required to get Sentry/Sentinel to behave correctly. 

I think the following would be useful to continue:

1. Get all controllers using IdentityProvider to gate access to the User rather than grabbing Sentry/Sentinel directly from container. 
2. Create factories for test Users that can be used to simplify tests. 
3. Drop a lot of the setup from tests where we can. 

I check out the branch later and take a deeper look.

chartjes:
@mdwheele 

I think an IdentityProvider would be a great first step -- creating a double for the provider rather than all the Sentry stuff seems like a much better testing approach

mdwheele:
I'm checking this out; running through the upgrade now. I'll send some PRs to the branch shortly. Immediately ran into:

```
vagrant@homestead:~/Code/opencfp$ composer install


  [Seld\JsonLint\ParsingException]
  "./composer.lock" does not contain valid JSON
  Parse error on line 921:

  ---------------------^
  Expected: 'STRING' - It appears you have an extra trailing comma


install [--prefer-source] [--prefer-dist] [--dry-run] [--dev] [--no-dev] [--no-custom-installers] [--no-autoloader] [--no-scripts] [--no-progress] [--no-suggest] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--ignore-platform-reqs] [--] [<packages>]...
```

mdwheele:
I've opened #499 to fix a few issues I had getting the branch running for a pre-existing user. I've gone so far as to log in and that's it. I'll dig in a bit more now.

mdwheele:
[...]

PR #501, opened 2017-10-06
Refactor all clients of Sentry to rely on interfaces we own.

mdwheele:
Related; #498

This PR adds three interfaces related to authentication and account management:

- `Authentication`: Used to handle authenticated sessions, focused on the web (not OAuth). Should be used anywhere we need to check if someone is authenticated or not. Should also be used if you need reference to a Sentry/Sentinel `UserInterface` rather than the Spot entity. 

  This abstraction is leaky in that it returns interfaces we don't own. That's okay for now because at least we have a centralized target for future refactors now and I didn't see much value in wrapping this. Personally, I would prefer to collapse user models down to one and potentially dump Spot... but outside scope.

- `IdentityProvider`: Used to get the current Spot user. If we were using relationships off of Spot, this would be useful, but currently, most callers were looking for a Sentry User and using the Talk mapper to get related talks.

- `AccountManagement`: Used to create accounts, activate them, promote/demote admins, find users by login, identifier or role. Note that creation of users, by default, should activate them.

The need for both `IdentityProvider` and `Authentication` goes away if we collapse User models. I think this is fine for now, just worth a thought later.

There is a `Sentry` implementation for each of these to maintain this PR as a "refactor" only. When this is merged, an alternative implementation using Sentinel can be created with required database migrations. I've got a branch with most of this work done that just needs to be ported to these interfaces. Ultimately, we need to be diligent in using these interfaces and making sure they're used in future PRs to prevent inconsistency.

I broke my commits into individual refactor steps as best as I could; trying to make sure the tests passed at all points. There were several adjustments required to keep tests passing. Most of these were oriented about mocks of Sentry needing to be updated to the appropriate interfaces. Some tests got very cleaned up. Others stayed about the same. I don't think any got *harder* to reason about but now there is a lot of opportunity for folks having an easier time testing. Just make a double for the interface and forget about all the underlying detail.

After patching everything, I ran through the application manually testing as many features as I could. I found a few issues not covered by tests:

- The interface for Intervention 2.x was different than 1.x. Made several changes
- Discovered a few typos in different classes that weren't covered by tests

Anywho, this is ready for review as far as I'm concerned.


mdwheele:
I modified `DatabaseTestcase` to use MySQL instead of sqlite3. When TravisCI makes a relevant version of sqlite3 available, we can take advantage of that. Unfortunately, the version available does not like the airports migration because of the bulk insert. This really doesn't affect runtime all that much, TBH. And it certainly wasn't worth the headache in rolling special Docker containers to get xenial.

Issue #485, opened 2017-09-14
Request for feedback: Improve review mechanism


BackEndTea:
We would like to use OpenCFP for the Dutch PHP Conference, but are considering developing features for reviewing/rating talks.

- Adding a reviewer role, who would have limited access to the dashboard
- Adding a rating system based on a 1-10 score.
- Adding an overview of the talk scores for the admin.

Some of these ideas were mentioned earlier in #415. This isn’t quite the same idea as #484, as this would be a ‘private voting’ so to speak. This also won’t require any hooks with external ticketing systems.

Would such a feature be of any value to OpenCFP? And if so, any feedback on how to approach the goal of making rating/reviewing talks in OpenCFP easier is very welcome!

chartjes:
@BackEndTea I do have a long-running refactor where I am moving the app to Sentinel from Sentry for authentication purposes and had implemented the idea of having users who were reviewers but not admins.

I am not sure of the scope of work to add a rating system to replace or sit next to the existing one. I certainly don't object to seeing such a thing built on top of the existing system, but I seem to have gotten bogged down in the refactor of the auth system and have no end date in sight.

Issue #415, opened 2016-10-14
[RFC] Overhaul review mechanism


carnage:
I've been looking at the review mechanism in opencfp and talking to several organisers about how they do their selection - most have opted for an external tool for reviews. 

I'd like to gather opinions on how (if) this should be done before ploughing headlong into it.

Some initial ideas:

[ ] A score based rating system of talks instead (as well as?) of a +/- rating system.
[ ] A review panel role which allows access to the review talks pages (perhaps only once cfp has closed)
[ ] An anon review option (Users with only review panel role would not see speaker info)
[ ] Admin tool to blacklist speaker/talk preventing it showing to on the review pages.
[ ] Ability to create a shortlist of talks (either automatically based on rating or manually)

PR #503, opened 2017-10-07
Initial refactor to use a strategy class for handling ratings

carnage:
Playing with some options around talk ratings for #415 and #485 this is an initial refactor which removes the rating logic into a service class.

This piece of functionality is self contained and I'm going to suggest merging it, as the work going on around the migration away from spot my break the code.

This does introduce a minor bc break; the authentication interface gained a new method userId as a shortcut for retrieving the logged in user's id. This only presents a problem to people who've built their own authentication layer as a replacement to sentinel.

Further work required: (Future PR's)

- service locate the constructed strategy instead of calling new in the controller
- produce a similar service pattern for rendering the rating widget + rating score
- might also require some work around the logic for fetching unrated talks 

Once the above is completed, it should be possible to introduce new rating strategies controlled via config.

### DIFF
--- a/classes/Domain/Services/Authentication.php
+++ b/classes/Domain/Services/Authentication.php
@@ -25,6 +25,14 @@ interface Authentication
     public function user(): UserInterface;
 
     /**
+     * Returns current authenticated User Id.
+     *
+     * @return int
+     * @throws NotAuthenticatedException
+     */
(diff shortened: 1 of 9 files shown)


chartjes:
Thanks @carnage

PR #505, opened 2017-10-08
Added new option to select +1 rated talks

carnage:
A left over one from last year rebased against latest master.

Allows admin users to pull back a list of the talks that they +1'd


chartjes:
Thanks @carnage

Issue #506, opened 2017-10-08
[RFC] Refactor from using Spot to Illuminate


chartjes:
As I look at the upcoming refactor from Sentry to Sentinel, I notice that it natively supports working with illuminate/database. It could be easier to integrate other libraries and tools into OpenCFP if we support illuminate/database

mdwheele:
👍 for Eloquent. Behind that 100%.

BackEndTea:
I'm playing around a bit with this, and i'm wondering, would it be better to simply call the static models like this:
```php
$users = User::all();
```
Or if its better to load the models from the container so we can do this:
```php
$users = $this->service(User::class)->all();
```
And have some kind of Model provider that looks a bit like this
```php
public function register(Container $app)
    {
        $app[Favorite::class] = function ($app) {
            return new Favorite;
        };

        $app[Group::class] = function ($app) {
            return new Group;
        };

        $app[Talk::class] = function ($app) {
            return new Talk;
        };

        $app[TalkComment::class] = function ($app) {
            return new TalkComment;
        };

        $app[TalkMeta::class] = function ($app) {
            return new TalkMeta;
        };

        $app[User::class] = function ($app) {
            return new User;
        };
    }
```

And on another note, this is probably a good chance to up code coverage as this and #498 will require a decent amount of refactoring.

mdwheele:
I wouldn't do service lookup on Eloquent models. For such a relational model, calling them directly would be fine, I think. That said, it wouldn't be unreasonable to continue the use of things like `SpeakerRepository`.

mdwheele:
Left this on a related PR but it's probably more relevant here:

> Another thing to consider is "Why do we 'own' models for things related to a specific authentication library?" We should probably delegate to Sentinel through Authentication or AccountManagement rather than keeping our own. I'd say the same for Phinxlog, but I'm not entirely sure of callers right now.
> 
> As far as the model goes, I would like to see a lot of the functionality related to speakers behind a Speaker or User model (e.g. submitTalk, updateProfile, etc.) to avoid things like are in the current TalksController where we kinda hit the model from the side to create a Talk from thin air. This would also encapsulate a lot of the permissions checks we currently do in controllers (e.g. all the checks for User id matching to make sure you are authorized to edit Talk entities)

BackEndTea:
@mdwheele I changed one controller to work with Models, and the nice thing about working with the provider is that you can easily mock them out. Which becomes a bit more difficult with static calls.

[...]

PR #507, opened 2017-10-09
Initial refactor from spot to Illuminate

BackEndTea:
Related to #506.

This PR adds Illuminate Models to eventually replace the spot entities.

Since the airport functionality was already done via Illuminate i have replaced that to work with the Airport model.

While making the models i noticed that some tables or entities currently aren't used, so i decided to not make Models for those right now. These are:

* Phinxlog
* Tag
* TalkTag
* Throttle
* UserGroup

UserGroup is a pivot table, so thats handeld behind the scenes by Illuminate.

The next step would be to start replacing all usages of spot in the code base.


chartjes:
Thanks for the new structure and making things ready for Illuminate to replace Spot

mdwheele:
Another thing to consider is "Why do we 'own' models for things related to a specific authentication library?" We should probably delegate to `Sentinel` through `Authentication` or `AccountManagement` rather than keeping our own. I'd say the same for `Phinxlog`, but I'm not entirely sure of callers right now.

As far as the model goes, I would like to see a lot of the functionality related to speakers behind a `Speaker` or `User` model (e.g. `submitTalk`, `updateProfile`, etc.) to avoid things like are in the current `TalksController` where we kinda hit the model from the side to create a `Talk` from thin air.

chartjes:
PhinxLog is used to track database migrations, so it's not used by the app itself.

👍 to delegate things where we can so when the inevitable happens (Sentinel becoming deprecated) then refactoring to switch to something else is a little easier

PR #508, opened 2017-10-09
Use Illuminate Database for the admins dashboard controller

BackEndTea:
Related to #506.

This PR replaces all usage of spot in the admin dashboard controller with Illuminate Database models.
It also adds tests for this controller, and for the home made functions in the models.

Some small things besides that:
* Removed the Request object from the index function, since that was not used.
* Added an empty ```$guarded``` array to some models, so we can use the create() functionality.

To use the beginTransaction and rollBack functions you need to do so on the Capsule object in order for it to work

### DIFF
--- a/classes/Domain/Model/Talk.php
+++ b/classes/Domain/Model/Talk.php
@@ -25,4 +25,86 @@ class Talk extends Model
     {
         return $this->hasMany(TalkMeta::class, 'talk_id');
     }
+
+    /**
+     * Return a collection of recent talks
+     *
+     * @param $admin_id
+     * @param int $limit
+     *
+     * @return array|Talk[]
+     */
+    public function getRecent($admin_id, $limit = 10)
+    {
+        $talks = $this
+            ->orderBy('created_at')
+            ->with(['favorites', 'meta'])
+            ->get()
+            ->take($limit);
+        $formatted = [];
+        foreach ($talks as $talk) {
+            $formatted[] = $this->createdFormattedOutput($talk, $admin_id);
+        }
+        return $formatted;
+    }
+    /**
+     * Iterates over DBAL objects and returns a formatted result set
+     *
+     * @param  mixed   $talk
+     * @param  integer $admin_user_id
+     * @return array
+     */
+    public function createdFormattedOutput($talk, $admin_user_id, $userData = true)
+    {
+        if ($talk->favorites) {
+            foreach ($talk->favorites as $favorite) {
+                if ($favorite->admin_user_id == $admin_user_id) {
+                    $talk->favorite = 1;
+                }
+            }
+        }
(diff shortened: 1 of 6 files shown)


mdwheele:
Looks like an excellent start. I want to take a look at this branch a little bit and suggest some changes but am mobile at the moment. Most simple stylistic changes.

classes/Domain/Model/Talk.php
```diff
@@ -25,4 +32,86 @@ public function meta()
...
+
+    /**
+     * Return a collection of recent talks
+     *
+     * @param $admin_id
+     * @param int $limit
+     *
+     * @return array|Talk[]
+     */
+    public function getRecent($admin_id, $limit = 10)
```
mdwheele: For methods like this, use query scopes. Not doing so requires instantiation of an active record when, query scopes are the "factory method" we more or less need. https://laravel.com/docs/5.5/eloquent#local-scopes

classes/Domain/Model/TalkMeta.php
```diff
@@ -8,6 +8,12 @@ class TalkMeta extends Model
 {
     protected $table = 'talk_meta';
     public $timestamps = false;
+    /**
+     * The attributes that aren't mass assignable.
+     *
+     * @var array
+     */
+    protected $guarded = [];
```
mdwheele: Anytime I use Eloquent, the first thing I do is make a base model that does exactly this. We could go ahead and do that: `OpenCFP\Domain\Model\Model`.

mdwheele:
Also, nevermind the codeclimate failure. That'll go away once we finish the refactor.

mdwheele:
@BackEndTea I've got a patch based on this branch that removes the need to manually begin and rollback transactions. Since we're no longer executing `DatabaseTest` using sqlite, there was opportunity to simply things. 

https://github.com/ibuildingsnl/opencfp/compare/illuminate-admin-dashboard...mdwheele:patch/ibuildingsnl/illuminate-admin-dashboard

mdwheele:
Thanks @BackEndTea. This gets us started in a positive direction. Tremendous help!

There were a few adjustments note-worthy for future usage of Eloquent:

- Any time `$model->all()->count()` is used, two calls to the database are incurred: `all()` actually generates a result set and `count()` sends an aggregate query. Using `Model::count()` is enough.
- `getRecent` was simplified as:
```php
public function scopeRecent(Builder $query, $admin_id, $limit = 10)
{
    return $query
        ->orderBy('created_at')
        ->with(['favorites', 'meta'])
        ->take($limit)
        ->get()
        ->map(function ($talk) use ($admin_id) {
            return $this->createdFormattedOutput($talk, $admin_id);
        });
}
```

PR #509, opened 2017-10-10
Add Eloquent model factories to simplify test setup

mdwheele:
This PR

* [x] Introduces [Eloquent model factories](https://laravel.com/docs/5.5/database-testing#using-factories) to the test suite.
* [x] Adds sample "common" factories for our new `User` and `Talk` Eloquent models.
* [x] Simplifies a test method in `TalkTest` by using the factory. Serves as an example for further work.

Follows #508.
Related to #506.

It is worth noting that these factories are not only useful for tests that hit the database. It is equally useful to call `make()` instead of `create()` to spin-up relational models that are returned from mocks of things like `Authentication` or `AccountManagement`.

Example (assuming work continues and appropriate factories are added):

```php
public function mocking_the_speaker_repository_for_some_reason()
{
    $repository = Mockery::mock(SpeakerRepository::class);

    # This will make repo return User, with pre-defined set of generated Talks
    # completely in-memory and using the *real* classes. No need to mock
    # all the way down to the metal.
    $repository->shouldReceive('findById')->with(32)->andReturn(
        factory(User::class)->states('with-talks')->make(['id' => 32])
    );

    $this->assertInstanceOf(User::class, $repository->findById(32));
}
```


chartjes:
Thanks @mdwheele

PR #510, opened 2017-10-14
Fix active navigation class

mdwheele:
This PR

* [x] Corrects an issue where `current_page` returned a closure that when compared against the current `path($route)`, always evaluated as false, leaving the active CSS class for navigation items never applied.
* [x] Removes duplicate global registrations from service providers (Twig and WebGateway)

I'm working on a Sass-based restyling of the application and in the process of simplifying our templates / markup, came across this bug present on `master`.

**Before**

image

**After**

image


chartjes:
Thanks @mdwheele

PR #511, opened 2017-10-15
Failed login throws 500 error

carnage:
A failed login was not catching all the exceptions correctly and thus throwing a 500 error instead of a nice error message; this fixes that problem


chartjes:
Thanks @carnage

PR #512, opened 2017-10-16
Fix csv issues

BackEndTea:
**Steps required to reproduce the problem**

1. Create a talk
2. Make the description an excel executable formula like =2+3
3. Export csv of talks

**Expected Result**

* The cell with the talk description displays =2+3

**Actual Result**

* The cell displays 5

This particular instance isn't anything bad, but as detailed here: https://www.contextis.com/blog/comma-separated-vulnerabilities this could be used for all sorts of malpractice.

This fixes the issue by adding a ' in front of all cells that start with any of  ```= - + @``` to stop the formulas from executing

cc: @chartjes


chartjes:
Thanks @BackEndTea

PR #513, opened 2017-10-20
Move talk formatting to a new class

BackEndTea:
This PR moves the createdFormattedOutput function of the model to another class. The version in the talk mapper still exists, but is just used to point to the new function.

I moved the function to a new location in order to keep the models clean, and a bit of separation of concerns. I'm not 100% sure on the names of the class and interface, so feedback is welcome.

I also changed the scopeRecent a little bit to now just return a builder object instead of the formatted array. This is in line with how scopes are intended, and would allow us to chain multiple scopes if so desired.

Related to #506 .


chartjes:
Thanks @BackEndTea

PR #516, opened 2017-10-24
Fix: Update composer itself once only

localheinz:
This PR

* [x] removes a run of `composer self-update` from `before_install` section

💁‍♂️ For reference, see

* https://travis-ci.org/opencfp/opencfp/builds/288589424#L459-L462
* https://travis-ci.org/opencfp/opencfp/builds/288589424#L493-L495

❗️ There were times when the run of `composer self-update` on Travis CI would **not** update to the latest stable version, but these times are over, hehe.


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

Issue #395, opened 2016-06-03
Exception occurs when running migrations on MySQL 5.7.12


ramsey:
# Steps required to reproduce the problem
1. Set up a new database user and an empty database
2. Configure database settings in phinx.yml and config/production.yml
3. Run `vendor/bin/phinx migrate --environment=production`
# Expected Result
- Migrations succeed with no errors.
# Actual Result
- PDOException thrown when migrating `20150702132854_fix_defaults_for_users_table.php`:

```
[PDOException]
SQLSTATE[42000]: Syntax error or access violation: 1067 Invalid default value for 'activated_at'
```

It seems this might be the case because the value "0000-00-00 00:00:00" is not valid when `NO_ZERO_DATE` is enabled, and as of MySQL 5.7.8, it is enabled by default. See http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_zero_date

Setting the following in `my.cnf` fixes the problem, but I'm worried that my server mode is not strict enough now:

```
[mysqld]
sql_mode = 
```

mdwheele:
Thanks @ramsey. Related to this issue is https://github.com/opencfp/opencfp/blob/master/migrations/20160106061843_fix_zero_dates.php which was put in place to fix this exact issue for servers that had the users table (and more) and did an I place upgrade of MySQL. 

We can likely apply a similar fix up to the offending migration or (might actually be a better solution) set the SQL mode less permissive at runtime in order to get through the migrations (allowing the current existing one to do its job). SQL mode should not persist past the session required to do migrations but if it does we can revert to original configured value. 

Thoughts anyone else?

chartjes:
I'm in favour of any solution that does not require users to alter their mysqld settings...

chartjes:
There are no longer any migrations that set dates to 0000-00-00 00:00:00 so I am closing this issue.

PR #517, opened 2017-10-24
Database migration changes

chartjes:
* made them database-agnostic
* removed method that checks if the database isn't working that was MySQL specific

### DIFF
--- a/migrations/20141121173412_add_talk_meta.php
+++ b/migrations/20141121173412_add_talk_meta.php
@@ -15,7 +15,7 @@ class AddTalkMeta extends AbstractMigration
             ->addColumn('rating', 'integer', ['default' => 0])
             ->addColumn('viewed', 'boolean', ['default' => false])
             ->addColumn('created', 'datetime')
-            ->addIndex(['admin_user_id', 'talk_id'], ['name' => 'talk_meta--admin_user_id__talk_id', 'unique' => true])
+            ->addIndex(['admin_user_id', 'talk_id'], ['name' => 'talk_meta__admin_user_id__talk_id', 'unique' => true])
             ->create();
     }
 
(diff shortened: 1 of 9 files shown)


migrations/20141121173412_add_talk_meta.php
```diff
@@ -15,8 +15,8 @@ public function change()
             ->addColumn('rating', 'integer', ['default' => 0])
             ->addColumn('viewed', 'boolean', ['default' => false])
             ->addColumn('created', 'datetime')
-            ->addIndex(['admin_user_id', 'talk_id'], ['name' => 'talk_meta--admin_user_id__talk_id', 'unique' => true])
+            ->addIndex(['admin_user_id', 'talk_id'], ['name' => 'talk_meta__admin_user_id__talk_id', 'unique' => true])
```
mdwheele: For databases that supported the `--` characters, do we want to fix the index names or leave it alone?

migrations/20160106061843_fix_zero_dates.php
```diff
@@ -19,7 +19,7 @@ class FixZeroDates extends AbstractMigration
 
     public function up()
     {
-        $this->execute("SET SESSION sql_mode = ''");
+        //$this->execute("SET SESSION sql_mode = ''");
```
mdwheele: These changes were put in place [because of this issue](https://github.com/opencfp/opencfp/issues/395). If we support MySQL versions < 5.7.8, we'll need some special sauce for this. From your comment in that issue, I also don't like the idea of having to change `mysqld` settings to get the migrations to run so this might cause a regression, but I don't have an old version of MySQL to test on *right now*.

migrations/20141121173412_add_talk_meta.php
chartjes: I changed them because Postgres doesn't like them

migrations/20160106061843_fix_zero_dates.php
chartjes: Same thing with this one.

There is no big urgency on this PR but it is laying the groundwork for allowing the app to be deployed to Heroku

mdwheele:
👍

PR #519, opened 2017-11-03
Fix twig errors on non existent route

BackEndTea:
Before this, if the route the user requested wasn't defined, it would
cause an error in the 404 page, as it didn't go through the before
functions.

This commit fixes that by first of all binding it to app, instead of web
so it checks in all routes, even if it doesn't match the web, or any
route we know. It also makes it an Early event, which means that it will
fire off, even if its a 404.

**Old version**
image


chartjes:
Thanks @BackEndTea

PR #520, opened 2017-11-06
Refactor of Talk Rating

BackEndTea:
This PR does a few things:

 1. Create a TalkRating base class which has some standard functionality like saving the new rating.
 2. Create a TalkRatingContext, which defines what strategy to use
 3. Create a Provider which calls the TalkRatingContext
 4. Override some standard Illuminate functionality in relation to created at and updated at functions.

Follows #503 .

### DIFF
--- a/classes/Domain/Services/TalkRating/TalkRatingContext.php
+++ b/classes/Domain/Services/TalkRating/TalkRatingContext.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace OpenCFP\Domain\Services\TalkRating;
+
+use OpenCFP\Domain\Model\TalkMeta;
+use OpenCFP\Domain\Services\Authentication;
+
+class TalkRatingContext
+{
+    public static function getTalkStrategy(string $strategy, Authentication $auth) : TalkRatingStrategy
+    {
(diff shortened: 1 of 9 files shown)


classes/Domain/Services/TalkRating/TalkRatingContext.php
```diff
@@ -0,0 +1,20 @@
...
+    public static function getTalkStrategy(string $strategy, Authentication $auth) : TalkRatingStrategy
+    {
+        $strategy = strtolower($strategy);
+        switch ($strategy) {
+            case 'yesno':
+                return new YesNoRating(new TalkMeta, $auth);
+            default:
+                return new YesNoRating(new TalkMeta, $auth);
+        }
+    }
```
chartjes: Given the default case is the same as 'yesno', can't this just be

return new YesNoRating(new TalkMeta, $auth);

BackEndTea: That would also do the trick, but i made it this way as a set-up for adding more rating options.

chartjes:
Thanks @BackEndTea

PR #522, opened 2017-11-07
Refactor talk view action to Illuminate

BackEndTea:
This PR makes the talk view action in the admins controller use Illuminate instead of spot.

This was a bit more intensive then initially expected, so this PR includes all the following changes as well

1. New functions in the TalkMeta and User model
2. Admin Talks Controller Test now uses our own test case as a base class, plus some changes to make this work
3. Put the Database Test Case in a trait, so it can be used in combination with our normal test case. Plus changes to the classes that originally extended the Database Test case to make them work again.
4. A few small changes to templates to accommodate the changes from spot to illuminate.
5. Added an asAdmin function to the base testcase which does the sentry mocking for us. 

Using the database in our tests does slow it down a bit, but if we keep the amount of DB interaction to a minimum it might still be faster than using the overload: functionality in Mockery. Plus it makes the tests a lot smaller and a lot more readable.

Related to #506 .


chartjes:
Thanks @BackEndTea

PR #523, opened 2017-11-07
Enhancement: Return types and general clean up

BackEndTea:
This PR adds return types to a lot of functions. Functions that could also return null/ other types of responses haven't been updated.

It also forces some types on parameters where possible. And updates some docblocks

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -18,6 +18,7 @@ $config = PhpCsFixer\Config::create()
         'psr0' => false,
         'single_blank_line_before_namespace' => true,
         'return_type_declaration' => true,
+        'single_blank_line_before_namespace' => true,
         'single_quote' => true,
         'trailing_comma_in_multiline_array' => true,
         'yoda_style' => [
--- a/classes/Application/Speakers.php
+++ b/classes/Application/Speakers.php
@@ -58,7 +58,7 @@ class Speakers
     /**
      * Retrieves a talk owned by a speaker.
      *
-     * @param $talkId
+     * @param int $talkId
      *
      * @return Talk
      * @throws NotAuthorizedException
--- a/classes/Application/Speakers.php
+++ b/classes/Application/Speakers.php
@@ -48,7 +48,7 @@ class Speakers
      *
      * @return SpeakerProfile
      */
-    public function findProfile() : SpeakerProfile
+    public function findProfile(): SpeakerProfile
     {
         $speaker = $this->identityProvider->getCurrentUser();
 
@@ -96,7 +96,7 @@ class Speakers
      * @return Talk
      * @throws \Exception
      */
-    public function submitTalk(TalkSubmission $submission) : Talk
+    public function submitTalk(TalkSubmission $submission): Talk
     {
         if (!$this->callForProposal->isOpen()) {
             throw new \Exception('You cannot create talks once the call for papers has ended.');
--- a/classes/Application/Speakers.php
+++ b/classes/Application/Speakers.php
@@ -48,7 +48,7 @@ class Speakers
      *
      * @return SpeakerProfile
      */
-    public function findProfile()
+    public function findProfile() : SpeakerProfile
     {
         $speaker = $this->identityProvider->getCurrentUser();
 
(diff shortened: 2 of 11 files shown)


.php_cs
```diff
@@ -15,6 +15,7 @@ $config = PhpCsFixer\Config::create()
         'no_unused_imports' => true,
         'ordered_imports' => true,
         'psr0' => false,
+        'single_blank_line_before_namespace' => true,
```
localheinz: @BackEndTea 

What do you think about enabling this fixer in a separate PR (similar to - for example - https://github.com/opencfp/opencfp/pull/524)? 

Personally, I like to do this in two steps:

* enable the fixer and commit (see the build fail)
* run `make cs` and commit (make the build pass)

A side effect of this is that I wait with enabling a fixer until I see a PR opened that indicates that a project could benefit from enabling a specific rule (for example, I spotted here that - while consistent in context of this PR - spacing is applied before `:` in return type declarations, but both the proposed [PSR-12](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md) as well as the default configuration for the `return_type_declaration` fixer indicate that no space should be used, so it's a good case for enabling the fixer on an existing project). 🤓

classes/Application/Speakers.php
```diff
@@ -63,7 +63,7 @@ public function findProfile()
      * @return Talk
      * @throws NotAuthorizedException
      */
-    public function getTalk($talkId)
+    public function getTalk(int $talkId)
```
localheinz: I'm not entirely sure whether we have filtering in place that sorts out passing in a possibly *integerish* `string` value via the UI, which if we didn't would fail with a `TypeError`. What do you think, do we need to add missing tests here?

Also, what do think about adding/updating the corresponding `@param`s in the docblocks?

classes/Domain/Model/Airport.php
```diff
@@ -15,7 +15,7 @@ class Airport extends Eloquent implements AirportInformationDatabase
      * @return Airport
```
localheinz: Should we update docblocks here and in other places as well?

I would like to refer to https://github.com/zendframework/zend-diactoros/pull/37#issuecomment-107195929. 

What do you think, @BackEndTea?

classes/Domain/Model/TalkComment.php
```diff
@@ -1,6 +1,5 @@
 <?php
 
-
```
localheinz: Looks like we definitely have a good case for enabling the `single_blank_line_before_namespace` fixer!

👍

classes/Util/Immutable.php
```diff
@@ -1,21 +0,0 @@
-<?php
-
-namespace OpenCFP\Util;
-
-use Exception;
-
-trait Immutable
```
localheinz: What do you think about cherry-picking cf6254927a406c3f77f1159a8ec6cfbcbbf0ef73 and consequently removing

* `Immutable`
* `ImmutableObject`

- both of which are in fact unused - as well as

* `ImmutableTest`

in a separate PR?

While I think all of the changes proposed in this PR here are good, I tend to prefer smaller PRs addressing single issues, as they are easier to review (and thus make it faster into `master`). Hope you don't mind me expressing this concern.

BackEndTea: Have to admit i had to figure out how to cherry pick a commit, but this is now done in #528

localheinz:
@BackEndTea 

This needs a good rebase after all of your other outstanding PRs have been merged!

BackEndTea:
@localheinz I'll get to that once @chartjes has finished most PR's.

classes/Domain/Model/Airport.php
BackEndTea: Yeah i'll get around to updating that.

I guess this is also a time to wonder how much the docblocks are needed, if we can paint the whole picture with type declarations.

classes/Application/Speakers.php
BackEndTea: I checked where this function is called, and its only used in one place (outside of automated tests), which is the talk view action. And there it gets filtered to an int anyways.

So if we would call this from another place in the future it should enforce some validation as well.

PR #526, opened 2017-11-07
Enhancement: Add PHP 7.1 to build matrix

localheinz:
This PR

* [x] adds PHP 7.1 to the Travis CI build matrix

💁‍♂️ Also moving collecting of code coverage to PHP 7.1.

PR #527, opened 2017-11-07
Fix: Conditions for collection coverage

localheinz:
This PR

* [x] fixes conditions for collection code coverage 

Somewhat related to #526.

💁‍♂️ The environment variable `WITH_COVERAGE` (previously used for indicating the desire to collect code coverage) is actually never used, but a variable `COLLECT_COVERAGE` is set in the `before_script` section, so I remove the former and rename the latter here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
   fast_finish: true
   include:
     - php: 7.0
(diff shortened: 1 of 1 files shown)


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #528, opened 2017-11-07
Remove Immutable

BackEndTea:
The only place immutable was used was in the old airport information

So this PR removes it from the project, including its tests


chartjes:
Thanks @BackEndTea (and @localheinz for the review)

PR #529, opened 2017-11-07
Enhancement/single blank line before namespace

BackEndTea:
This PR

* [x] enables the ```single_blank_line_before_namespace``` fixer
* [x] runs ```make cs``` 

For reference see: https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.7.1#usage

>**single_blank_line_before_namespace** [@Symfony]
>There should be exactly one blank line before a namespace declaration.


BackEndTea:
I did notice that this didn't fix some namespaces. The ones that are on the same line as the ```<?php``` tag don't get fixed by this.

localheinz:
@BackEndTea 

We can sort this out in https://github.com/opencfp/opencfp/pull/530!

chartjes:
Thanks @localheinz and @BackEndTea

PR #531, opened 2017-11-07
Fix: Do not import native PHP classes

localheinz:
This PR

* [x] removes imports for native PHP classes

💁‍♂️ It might be a matter of personal taste, but by never importing native PHP classes (and instead referencing them from relative to the root namespace), a) the list of imports is reduced and b) it's immediately clear by looking at a line of code whether it is referencing a class from the root namespace or not.


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #532, opened 2017-11-07
Fix: Use class keyword

localheinz:
This PR

* [x] makes use of the `class` keyword where possible


tests/Console/AdminPromoteCommandTest.php
```diff
@@ -85,7 +85,7 @@ public function promoteExistingNonAdminAccount()
 
     protected function createInputInterfaceWithEmail($email)
     {
-        $input = Mockery::mock('\Symfony\Component\Console\Input\InputInterface');
+        $input = Mockery::mock(\Symfony\Component\Console\Input\InputInterface::class);
```
localheinz: There are quite a bunch of references to namespaced classes in this fashion, but I figured to ease review, it's better to sort out imports in a separate PR.

localheinz:
Thank you, @chartjes!

PR #533, opened 2017-11-07
Fix merging errors

BackEndTea:
This PR fixes some errors that came from all these PR's.

- Fixed the view action function, since it was a mix of spot and illuminate for a bit.
- Renamed the second sanitize function to internalSanitize to stop the errors.


chartjes:
Thanks @BackEndTea for saving me from myself

PR #534, opened 2017-11-07
Fix: Use constants for HTTP response status codes

localheinz:
This PR

* [x] uses constants for HTTP response status codes (instead of magic numbers)


chartjes:
I need a macro for thanking @localheinz

localheinz:
I usually use saved replies, @chartjes!

PR #538, opened 2017-11-07
Fix: Remove unused private method

localheinz:
This PR

* [x] removes an unused `private` method

```
$ git grep -i generateTalks

tests/Domain/Model/TalkTest.php:114:    private function generateTalks()
```


localheinz:
Thank you, @chartjes!

PR #539, opened 2017-11-07
Fix: Reference phpunit.xsd using relative path

localheinz:
This PR

* [x] references the path to the `phpunit.xsd` schema relative to `phpunit.xml.dist`


localheinz:
Thank you, @chartjes!

PR #540, opened 2017-11-07
Fix: Remove duplicate font-awesome fonts

localheinz:
This PR

* [x] removes duplicate fonts

💁‍♂️ They are required in [`package.json`](https://github.com/opencfp/opencfp/blob/7854645744883cf854502b10a891f79f414f4ece/package.json#L13), and running

```
$ yarn install
$ yarn run production
```

will install them the `node_modules/`, and then copy them to `/web/assets/fonts/` (where they are still checked in, probably for those folks who cannot run either of these commands during deployment).

❓ Do we really want to check in build artefacts?


localheinz:
Thank you, @chartjes!

PR #541, opened 2017-11-07
Fix: Suggest to use yarn (rather than npm)

localheinz:
This PR

* [x] removes suggestions to use `npm` instead of `yarn`

💁‍♂️ It's painfully slow, and running `npm install` rather than running `yarn install` with a checked in `yarn.lock` will eventually produce different results (as `npm` doesn't take into account locked dependencies in `yarn.lock`).


localheinz:
Thank you, @chartjes!

PR #542, opened 2017-11-07
Enhancement: Use stages on Travis

localheinz:
This PR

* [x] makes use of the stages feature on Travis

💁‍♂️ For reference, see https://docs.travis-ci.com/user/build-stages. The advantage of using stages is this:

* builds are executed stage by stage, but subsequent stages are only executed if builds for prior stages passed (if coding standard checks failed, we don't need to even worry about tests)
* we can fine-tune each stage for the desired steps
* we can add more stages (for example, for compiling assets to see if these even compile)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,19 +2,6 @@ language: php
 
 sudo: false
 
-addons:
-  apt:
-    sources:
-      - travis-ci/sqlite3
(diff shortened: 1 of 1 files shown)


localheinz:
Thank you, @chartjes!

PR #546, opened 2017-11-08
Fix: Extend from WebTestCase

localheinz:
This PR

* [x] fixes the failing build by extending from `WebTestCase`

Follows #545.


localheinz:
Thank you, @chartjes!

PR #547, opened 2017-11-08
Enhancement: Add asset stage

localheinz:
This PR

* [x]  adds an `asset` stage for building client-side assets

Follows #542.


localheinz:
Thank you, @chartjes!

PR #550, opened 2017-11-08
Fix: Remove duplicate key from php-cs-fixer configuration

localheinz:
This PR

* [x] removes a duplicate key from the configuration for `php-cs-fixer`


localheinz:
Thank you, @chartjes!

PR #549, opened 2017-11-08
Enhancement: Enable php_unit_expectation and php_unit_no_expectation_annotation fixers

localheinz:
This PR

* [x] enables the `php_unit_expectation` and `php_unit_no_expectation_annotation` fixers
* [ ] runs `make cs`
* [ ] moves expectations to desired location

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.8.0#usage:

>**php_unit_expectation** [@PHPUnit52Migration:risky, @PHPUnit54Migration:risky, @PHPUnit56Migration:risky, @PHPUnit57Migration:risky, @PHPUnit60Migration:risky]
>
>Usages of `->setExpectedException*` methods MUST be replaced by `->expectException* methods`.
>
>Risky rule: risky when PHPUnit classes are overridden or not accessible, or when project has PHPUnit incompatibilities.
>
>Configuration options:
>
>* `target` (`'5.2'`, `'5.6'`, `'newest'`): target version of PHPUnit; defaults to `'newest'`

and

>**php_unit_no_expectation_annotation** [@PHPUnit32Migration:risky, @PHPUnit35Migration:risky, `@PHPUnit43Migration:risky, @PHPUnit48Migration:risky, @PHPUnit50Migration:risky, @PHPUnit52Migration:risky, @PHPUnit54Migration:risky, @PHPUnit56Migration:risky, @PHPUnit57Migration:risky, @PHPUnit60Migration:risky]
>
>Usages of `@expectedException*` annotations MUST be replaced by `->setExpectedException*` methods.
>
>Risky rule: risky when PHPUnit classes are overridden or not accessible, or when project has PHPUnit incompatibilities.
>
>Configuration options:
>
>* `target` (`'3.2'`, `'4.3'`, `'newest'`): target version of PHPUnit; defaults to `'newest'`
`use_class_const` (`bool`): use `::class` notation; defaults to `true`

Follows #535.

PR #551, opened 2017-11-08
Fix: Move expectations in regard to exceptions

localheinz:
This PR

* [x] moves expectations in regard to exceptions right before the *act* phase

Follows #549.

💁‍♂️ For reference, see https://thephp.cc/news/2016/02/questioning-phpunit-best-practices.

### DIFF
--- a/tests/Http/API/ApiControllerTest.php
+++ b/tests/Http/API/ApiControllerTest.php
@@ -39,9 +39,11 @@ class ApiControllerTest extends \PHPUnit\Framework\TestCase
      */
     public function it_warns_when_successful_status_code_is_used_for_error()
     {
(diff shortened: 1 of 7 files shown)


tests/Http/API/ApiControllerTest.php
```diff
@@ -39,9 +39,11 @@ public function it_can_send_a_simple_json_response()
      */
     public function it_warns_when_successful_status_code_is_used_for_error()
     {
+        $this->sut->setStatusCode(HttpFoundation\Response::HTTP_OK);
+
         $this->expectException(\PHPUnit\Framework\Exception::class);
```
localheinz: Seems like this should be a different exception.

localheinz:
Thank you, @chartjes!

PR #552, opened 2017-11-08
Fix: Remove useless annotation

localheinz:
This PR

* [x] removes useless `@test` annotations

💁‍♂️ Test methods are already prefixed with `test`.


localheinz:
Thank you, @chartjes!

PR #553, opened 2017-11-09
Fix: Use more appropriate assertions

localheinz:
This PR

* [x] makes use of more appropriate assertions


tests/Domain/Entity/Mapper/UserTest.php
```diff
@@ -42,7 +42,8 @@ public function findAllUsersWithNoSearch()
     {
         $this->createFiveUsers();
         $result = $this->mapper->search()->toArray();
-        $this->assertEquals(5, count($result));
+
+        $this->assertCount(5, $result);
```
localheinz: Also adding a bit of whitespace as can be seen here and below.

localheinz:
Thank you, @chartjes!

PR #554, opened 2017-11-09
Fix: Undefined classes

localheinz:
This PR

* [x] adds imports to otherwise undefined classes (alternatively references them relative to the root namespace)


localheinz:
Thank you, @chartjes!

PR #555, opened 2017-11-09
Enhancement: Enable phpdoc_separation fixer

localheinz:
This PR

* [x] enables the `phpdoc_separation` fixer
* [x] runs `make cs`

>**phpdoc_separation** [`@Symfony`]
>
>Annotations in phpdocs should be grouped together so that annotations of the same type immediately follow each other, and annotations of a different type are separated by a single blank line.


localheinz:
Thank you, @chartjes!

PR #561, opened 2017-11-10
Enhancement: PHP Mess Detector for Code Climate

BackEndTea:
This PR adds the PHPMD option to code climate. 

This should be useful to give more insight into where faults within the code base are, and give better insight in what future PR's do to the complexity of the code base.

For more info see: https://docs.codeclimate.com/docs/phpmd, and the PHPMD documentation https://phpmd.org/

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -10,6 +10,8 @@ engines:
     enabled: true
     config:
       standard: PSR2
+  phpmd:
+    enabled: true
 
 ratings:
   paths:


BackEndTea:
Only 395 'issues' to fix

localheinz:
@BackEndTea 

Ha, no worries - I would suggest to understand Code Climate failures as soft signals only (unless we are talking about decreasing test coverage). After all `phpmd`, unless configured otherwise, will also complain about too short variable names. We probably don’t want to rename `$id` everywhere, for example.

BackEndTea:
@localheinz 

Yeah i'd agree on that. But i don't think CodeClimate tells us about decreasing test coverage in PRs, so it might be good to find a tool, or a code climate setting that does that.

localheinz:
@BackEndTea 

Actually, Code Climate does provide that, it just needs to be enabled. See https://codeclimate.com/changelog/59baa212ced24c026c0017b8. 

@chartjes 

What do you think about enabling this, along with reporting coverage on pull requests as well?

localheinz:
@BackEndTea 

Otherwise, I can also highly recommend https://codecov.io for collecting coverage. Nice and easy setup, central configuration per organization, adds build statuses and comments on pull requests (can be configured). See https://github.com/localheinz/github-changelog/pull/217 for an example.

chartjes:
It used to be turned on, I would have no problem with it being used again.

PR #564, opened 2017-11-10
Enhancement: Enable phpdoc_align fixer

localheinz:
This PR

* [x] enables the `phpdoc_align` fixer
* [x] runs `make cs`

>**phpdoc_align** [`@Symfony`]
>
>All items of the given phpdoc tags must be aligned vertically.
>
>Configuration options:
>
>* `tags` (`array`): the tags that should be aligned; defaults to `['param', 'return', 'throws', 'type', 'var']`


localheinz:
Thank you, @chartjes!

PR #567, opened 2017-11-10
Enhancement: Enable phpdoc_scalar fixer

localheinz:
This PR

* [x] enables the `phpdoc_scalar` fixer
* [ ] runs `make cs`

>**phpdoc_scalar** [`@Symfony`]
>
>Scalar types should always be written in the same form. `int` not `integer`, `bool` not `boolean`, `float` not `real` or `double`.


localheinz:
Thank you, @chartjes!

PR #568, opened 2017-11-10
Enhancement: Enable phpdoc_no_empty_return fixer

localheinz:
This PR

* [x] enables the `phpdoc_no_empty_return` fixer
* [x] runs `make cs`
* [x] removes extra empty lines

>**phpdoc_no_empty_return** [`@Symfony`]
>
>@return void and @return null annotations should be omitted from phpdocs.


localheinz:
Thank you, @chartjes!

PR #572, opened 2017-11-10
Enhancement: Enable trim_array_spaces fixer

localheinz:
This PR

* [x] enables the `trim_array_spaces` fixer
* [x] runs `make cs`

> **trim_array_spaces** [`@Symfony`]
>
>Arrays should be formatted like function/method arguments, without leading or trailing single line space.


mdwheele:
👍

localheinz:
Thank you, @mdwheele!

PR #573, opened 2017-11-11
Refactor: Exports controller to Illuminate

BackEndTea:
No tests due to the exit() function being called. I can't quite figure out how to do this without html ending up in the csv

Related to #506.


mdwheele:
Thanks @BackEndTea!

PR #577, opened 2017-11-12
Refactor: Change Spot repositories to Illuminate repositories

BackEndTea:
This PR adds Illuminate talk and speaker repositories, which replace the spot ones.
It also updates a few other classes that relied on spot specific syntax.

Related to #506.

On another note, this broke one of the API tests, but the API doesn't work at all. It relies on the League\OAuth2\ system, which isn't even required in the composer.json anymore. 

If were still supporting the API i'll fix it up, but it looks like the API hasn't worked in quite some time.

### DIFF
--- a/classes/Application/Speakers.php
+++ b/classes/Application/Speakers.php
@@ -3,7 +3,7 @@
 namespace OpenCFP\Application;
 
 use OpenCFP\Domain\CallForProposal;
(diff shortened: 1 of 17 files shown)


mdwheele:
Don't worry about API in this PR. It can be fixed separately (and additional tests added to prevent this happening in the future).

PR #579, opened 2017-11-12
Use `random_bytes` instead of `ircmaxell/random-lib` to get tests passing on PHP 7.1

mdwheele:
This PR

* [x] Removes dependence on `ircmaxell/random-lib`, which was previously used to generate random strings of arbitrary length.
* [x] Sets target platform (PHP 7.0.25) in `composer.json` because I ran into issues with "Style" build step failing to install dependencies on PHP 7.0 because dependencies updated after removing `random-lib` (locally running PHP 7.1)

#526 fails because `random-lib` uses `mcrypt`, which is deprecated in PHP 7.1. This causes the test-suite to fail. For such a use-case as ours, `random_bytes` suffices.

Follows #526.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,6 @@ language: php
 
 sudo: false
 
-addons:
-  apt:
-    sources:
-      - travis-ci/sqlite3
-    packages:
-      - sqlite3
-
 matrix:
   fast_finish: true
   include:
--- a/classes/Infrastructure/Crypto/PseudoRandomStringGenerator.php
+++ b/classes/Infrastructure/Crypto/PseudoRandomStringGenerator.php
@@ -3,27 +3,11 @@
 namespace OpenCFP\Infrastructure\Crypto;
 
 use OpenCFP\Domain\Services\RandomStringGenerator;
-use RandomLib\Factory;
 
 class PseudoRandomStringGenerator implements RandomStringGenerator
 {
-    /**
-     * @var Generator
-     */
-    private $generator;
-
-    public function __construct(Factory $factory)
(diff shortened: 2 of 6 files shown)


mdwheele:
@localheinz any tips on why matrix expansion isn't happening for build stages in this PR? I'm following up on your introduction of PHP 7.1 to the matrix and have cleaned up the dependency issues. Just need to get Travis CI config straightened out.

localheinz:
@mdwheele

Apologies for the late reply, will take a look today!

localheinz:
@mdwheele 

See https://github.com/mdwheele/opencfp/pull/5.

localheinz:
@mdwheele 

Build is now failing because of undefined classes in migrations.

mdwheele:
Looks like that got it. Thanks for the help with Travis. Do you have a link to docs from them on how it works? The matrix expansion docs I found were at https://docs.travis-ci.com/user/build-stages/#Build-Stages-and-Build-Matrix-Expansion. However, I think that would have expanded every stage. I couldn't find anything with the syntax you provided (which obviously does exactly what we want).

localheinz:
@mdwheele 

Maybe http://blog.daemonl.com/2016/02/yaml.html helps?

mdwheele:
> Maybe http://blog.daemonl.com/2016/02/yaml.html helps?

Hahaha, yes! My usage of YAML is obviously rudimentary :stuck_out_tongue_closed_eyes: 

Everything else look decent here? I don't think we were necessarily getting anything super-special out of this dependency that this implementation couldn't equally provide.

PR #580, opened 2017-11-13
Fix: Order of sections

localheinz:
This PR

* [x] fixes the order of sections

💁‍♂️ While the order in `travis.yml` doesn't matter to Travis, it does do the interested reader.


localheinz:
Thank you, @chartjes!

PR #562, opened 2017-11-10
Update test coverage settings

chartjes:
This PR

* updates the Travis settings to provide support for code coverage messages to be sent

PR #581, opened 2017-11-13
[WIP] Fix: Always collect coverage

localheinz:
❗️ Blocked by #579.

This PR

* [x] always collects coverage

Follows https://github.com/opencfp/opencfp/pull/561#issuecomment-343460931.

💁‍♂️ Note that we should probably modify in regard to #579, that is, when PHP 7.1 is added to the build matrix. We should probably collect coverage only on the latest stable PHP version we decide to add to the build matrix. 

As you can see below, coverage diffs are now displayed in the build statuses:

screen shot 2017-11-13 at 11 54 56

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,6 @@ jobs:
             - sqlite3
 
       before_install:
-        - if [[ "$TRAVIS_PULL_REQUEST" == "false" && "$TRAVIS_BRANCH" == "master" ]]; then WITH_COVERAGE=true; else WITH_COVERAGE=false; fi
         - if [[ "$WITH_COVERAGE" == "false" ]]; then phpenv config-rm xdebug.ini; fi
         - composer validate --no-check-publish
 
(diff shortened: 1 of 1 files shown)


localheinz:
@mdwheele 

Let's put this on hold until #579 is sorted out, ok?

mdwheele:
No problem

BackEndTea:
How does this work in relation to sending the new coverage/ issues to Code climate?

localheinz:
@BackEndTea 

As pointed out in https://github.com/opencfp/opencfp/pull/562#issuecomment-343502363, the build will then be marked as failed when the coverage drops under a configured value.

Does that answer your question?

BackEndTea:
@localheinz 

Yeah i got it, i was just wondering how code climate knew when to 'update' their page of this repo, since it now always runs ``` ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT``` in the after script.

PR #582, opened 2017-11-13
Enhancement: Enable no_alias_functions fixer

localheinz:
This PR

* [x] enables the `no_alias_functions` fixer
* [x] runs `make cs`

>**no_alias_functions** [`@Symfony:risky`]
>
>Master functions shall be used instead of aliases.
>
>Risky rule: risky when any of the alias functions are overridden.


localheinz:
Thank you, @chartjes!

PR #584, opened 2017-11-13
Enhancement: Enable cast_spaces fixer

localheinz:
This PR

* [x] enables the `cast_spaces` fixer
* [x] runs `make cs`

>**cast_spaces** [`@Symfony`]
>
>A single space or none should be between cast and variable.
>
>Configuration options:
>
>* `space` (`'none'`, `'single'`): spacing to apply between cast and variable; defaults to `'single'`


localheinz:
Thank you, @chartjes!

Quick question, though, why the revert?

chartjes:
The style checker wanted some changes so the build was failing. I reverted, manually merged, ran phpcs, committed the new fixes, and we're all good now!

PR #588, opened 2017-11-13
Fix: Do not attempt to report coverage using old reporter

localheinz:
This PR

* [x] removes an invocation of the non-existent test reporter from `.travis.yml`

Follows #562.


localheinz:
Thank you, @chartjes!

PR #590, opened 2017-11-13
Enhancement: Collect coverage on a single PHP version only

localheinz:
This PR

* [x] collects coverage on a single PHP version only

Follows #581.

💁‍♂️ The idea is to speed up the build by collecting coverage on a single PHP version build only, and using the latest version of it (faster).

### Before

screen shot 2017-11-13 at 18 29 05

### After

screen shot 2017-11-13 at 18 28 29

### DIFF
--- a/.travis/code-climate.sh
+++ b/.travis/code-climate.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+# See
+#
+# - https://github.com/codeclimate/test-reporter#installation--usage
+# - https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables
+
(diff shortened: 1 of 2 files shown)


.travis/code-climate.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
localheinz: I have noticed that there are files related to setting up / configuring things on Travis CI in [`/tools/travis`](https://github.com/opencfp/opencfp/tree/master/tools/travis) - shall we move this file there, or maybe move the other files here?

localheinz:
Thank you, @chartjes!

PR #591, opened 2017-11-13
Fix: Move scripts related to Travis to .travis/

localheinz:
This PR

* [x] moves scripts related to Travis CI to `.travis/`

Related to https://github.com/opencfp/opencfp/pull/590#discussion_r150603781.

💁‍♂️ Could not find any usages of

* `setup-webserver.sh`
* `travis-ci-apache`

anywhere, should we remove these two files?

PR #594, opened 2017-11-13
Fix: Remove unused script

localheinz:
This PR

* [x] removes an unused script

Somewhat related to #591.

💁‍♂️ Running

```
$ git grep -i first_time_migrations
```

yields no results.


localheinz:
Thank you, @chartjes!

PR #595, opened 2017-11-13
Enhancement: Disable xdebug and enable only when needed

localheinz:
❗️ Blocked by #596.

This PR

* [x] disables xdebug and enables it only when needed


localheinz:
Thank you, @chartjes!

PR #599, opened 2017-11-13
Updated Symfony packages

chartjes:
This PR

* Updated some Symfony packages to be ~3.3 instead of ~3.2
* Adds Infection (mutation testing tool) as a require-dev option

I'm hoping Infection helps us get some more focus on our existing tests to be more specific and a little less loose.

Once this is merged I will tag master as v1.4.0


composer.json
```diff
@@ -29,24 +29,25 @@
...
-        "symfony/css-selector": "~3.2",
-        "symfony/form": "~3.2",
-        "symfony/intl": "~3.2",
-        "symfony/security": "~3.2",
-        "symfony/security-csrf": "~3.2",
-        "symfony/translation": "~3.2",
-        "symfony/twig-bridge": "~3.2",
-        "symfony/validator": "~3.2",
-        "symfony/yaml": "~2.5",
+        "symfony/config": "~3.3",
```
localheinz: @chartjes 

Since `~major.minor` is the equivalent of `^major.minor`, what do you think about using the `^` operator instead?

chartjes: No objections, let me make that change

localheinz: 👍

PR #600, opened 2017-11-13
Enhancement: Enable binary_operator_spaces fixer

localheinz:
This PR

* [x] enables and configures the `binary_operator_spaces` fixer
* [x] runs `make cs`

>**binary_operator_spaces** [`@Symfony`]
>
>Binary operators should be surrounded by space as configured.
>
>Configuration options:
>
>* `align_double_arrow` (`false`, `null`, `true`): (deprecated) Whether to apply, remove or ignore double arrows alignment; defaults to `false`
>* `align_equals` (`false`, `null`, `true`): (deprecated) Whether to apply, remove or ignore equals alignment; defaults to `false`
> * `default` (`'align'`, `'align_single_space'`, `'align_single_space_minimal'`, `'single_space'`, `null`): default fix strategy; defaults to `'single_space'`
> * `operators` (`array`): dictionary of binary operator => fix strategy values that differ from the default strategy; defaults to `[]`

### DIFF
--- a/classes/Domain/Entity/TalkTag.php
+++ b/classes/Domain/Entity/TalkTag.php
@@ -11,9 +11,9 @@ class TalkTag extends Entity
     public static function fields()
     {
         return [
-            'id' => ['type' => 'integer', 'primary' => true, 'autoincrement' => true],
+            'id'      => ['type' => 'integer', 'primary' => true, 'autoincrement' => true],
             'talk_id' => ['type' => 'integer', 'required' => true, 'unique' => 'talk_tag'],
-            'tag_id' => ['type' => 'integer', 'required' => true, 'unique' => 'talk_tag'],
+            'tag_id'  => ['type' => 'integer', 'required' => true, 'unique' => 'talk_tag'],
         ];
     }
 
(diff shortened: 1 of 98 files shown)


classes/Domain/Entity/TalkTag.php
```diff
@@ -21,7 +21,7 @@ public static function relations(\Spot\MapperInterface $mapper, \Spot\EntityInte
     {
         return [
             'talk' => $mapper->belongsTo($entity, \OpenCFP\Domain\Entity\Talk::class, 'talk_id'),
-            'tag'  => $mapper->belongsTo($entity, \OpenCFP\Domain\Entity\Tag::class, 'tag_id'),
```
localheinz: @chartjes 

Depending on what the preference is, we can also configure to align `=>` and `=` - what would you prefer?

chartjes: I think I like it all being lined up. 

I also think we should add some notes to the README for developers to let them know they should be running phpcs before every PR they submit.

localheinz:
@chartjes 

Not sure, is that it?

classes/Domain/Entity/TalkTag.php
localheinz: We have a note right here: https://github.com/opencfp/opencfp/blob/master/CONTRIBUTING.md#fixing-code-style-issues.

localheinz:
Thank you, @chartjes!

PR #602, opened 2017-11-14
Enhancement: Add new response asserts

BackEndTea:
This PR adds two new test asserts to improve the test suite

* [x] Assert No Flash Set, which checks that there is no flash set
* [x] Assert Target Url Contains, which checks if the target url of the redirect matches the given (sub) string

The assert target url contains method makes sure it is a redirect response first, since the getTargetUrl method isn't part of the normal response object.

Somewhat related to #599, since this should help improve tests

### DIFF
--- a/tests/TestResponse.php
+++ b/tests/TestResponse.php
@@ -4,6 +4,7 @@ namespace OpenCFP\Test;
 
 use OpenCFP\Application;
 use PHPUnit\Framework\Assert;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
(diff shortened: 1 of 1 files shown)


tests/TestResponse.php
```diff
@@ -94,6 +95,17 @@ public function assertFlashContains($flash)
         return $this;
     }
 
+    public function assertNoFlashSet()
+    {
+        Assert::assertNull($this->app['session']->get('flash'));
+    }
+
+    public function assertTargetURLContains($targetUrl)
```
localheinz: What do you think about adding a type declaration here?

BackEndTea: Good catch, didn't think about that, i'll update it in a second

localheinz:
@BackEndTea 

Needs a quick run of 

```
$ make cs
```

to get the build to pass, then I think this is good to go.

BackEndTea:
@localheinz I thought i could be quick and push it before heading to lunch, guess i overlooked that.

localheinz:
@BackEndTea 

Ha, no worries, hope you had a great lunch!

PR #606, opened 2017-11-14
Enhancement: Set up infection testing

BackEndTea:
This PR sets up infection testing. Run ```vendor/bin/infection``` to start up the tests or ```make infection```.
These tests take quite some time to run, so i haven't added them to travis. I've added the results of my run for those interested

Follows #599.

## Results
858 mutations were generated:
     403 mutants were killed
     370 mutants were not covered by tests
      72 covered mutants were not detected
      13 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 48%
         Mutation Code Coverage: 57%
         Covered Code MSI: 85%

[infection-log.txt](https://github.com/opencfp/opencfp/files/1470975/infection-log.txt)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ deployment.xml
 /web/.htaccess
 phinx.yml
 /phpunit.xml
+infection-log.txt
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,6 +32,12 @@ $ make test
 
 to run all the tests.
 
+Run 
+```
+$ make infection
+```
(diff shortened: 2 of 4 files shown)


chartjes:
I'm not sure we should be running these tests on a regular basis, mainly because I am not sure what actions we should be taking based on the results.

BackEndTea:
I think it will be a lot more useful to see what it does with code coverage above 80% or so. 
Running it should give more insight in what failures get caught by the test suite and what not. It basically gives an indication of how strong the regression testing is.

PR #617, opened 2017-11-16
Refactor: Move the UserInterface up to our own namespace

BackEndTea:
This PR moves the UserInterface up to our own namespace, which allows us to make use of it when moving from sentry to sentinel

Follows #501 .
Related to #498.


chartjes:
@BackEndTea looks like Codeclimate is not happy with you...

BackEndTea:
@chartjes Ive added tests for the new Class, and increased the coverage of the Authentication class while i was at it.

PR #620, opened 2017-11-17
Enhancement: code coverage

BackEndTea:
This PR adds coverage annotations to all tests to show what they are actually testing.

### DIFF
--- a/tests/Application/SpeakersTest.php
+++ b/tests/Application/SpeakersTest.php
@@ -14,6 +14,9 @@ use OpenCFP\Domain\Speaker\SpeakerRepository;
 use OpenCFP\Domain\Talk\TalkRepository;
 use OpenCFP\Domain\Talk\TalkSubmission;
 
+/**
+ * @covers \OpenCFP\Application\Speakers
+ */
 class SpeakersTest extends \PHPUnit\Framework\TestCase
 {
     const SPEAKER_ID = '1';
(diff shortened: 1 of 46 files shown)


localheinz:
@BackEndTea 

How about this instead:

* we explicitly split the tests into unit and integration tests

```
tests
├── Integration
└── Unit
```

* we ensure a 1:1 mapping between system under test and test class for unit tests
* we collect coverage only when running unit tests

Then

* we don't need to add any `@covers` or `$coversNothing` annotations
* we never collect accidental coverage
* we speed up the tests (as suggested earlier)

What do you think?

Alternatively, we can also enable the `php_unit_test_class_requires_cover` fixer, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.8.2#usage>

>**php_unit_test_class_requires_covers**
>
>Adds a default `@coversNothing` annotation to PHPUnit test classes that have no `@covers*` annotation.

localheinz:
@BackEndTea 

See #629.

BackEndTea:
@localheinz 
I like the idea of splitting up the tests between unit and integration tests, though
> we ensure a 1:1 mapping between system under test and test class for unit tests

Is a lot easier said then done. Therefore id like to split them up like you are doing in #629, but still keep the ```@covers``` annotations. And enabling ```php_unit_test_class_requires_covers``` looks like a good idea.

localheinz:
@BackEndTea 

Sounds good!

BackEndTea:
Closing this in favor of #629

PR #623, opened 2017-11-20
Enhancement: Update phpunit settings to ignore the API

BackEndTea:
This PR will ignore the API classes for code coverage, and exclude the remaining API tests from the standard testsuite.

This should give a better idea of the code coverage, and speed up the tests just a little bit.

Since phpunit will use ```phpunit.xml``` over ```phpunit.xml.dist``` 
```bash
$ cp phpunit.xml.dist phpunit.xml
```
Should be ran locally to update the file.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -10,6 +10,7 @@
          convertErrorsToExceptions="true"
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
+         defaultTestSuite="all"
          processIsolation="false"
          stopOnFailure="false">
     <php>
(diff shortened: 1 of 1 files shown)


localheinz:
@BackEndTea 

What do you think about splitting tests into

* integration tests
* unit tests

Then we could speed up the builds significantly by only collecting coverage for unit tests. I think we have a lot of mixed tests here, that’s why the build where we collect coverage is rather slow.

BackEndTea:
@localheinz 

Our real integration tests are currently the controller tests. But those are also the only thing covering the controllers right now.
Due to the way the controllers are set up, if we want to make true unit tests of those we need to use the mockery 'overload' feature. This slows the test down by a lot. When i played around with it they took twice as long as a test that builds up the database, interacts with it, and tears it down again.

I'm all for properly splitting up the tests, having unit tests for every inch, and then integration tests which do a full run of the app, it would require a different setup for the controllers to be any faster.

If the goal is speeding up the travis build we could also run the assets and style stages paralel, which would save a minute or so.

PR #625, opened 2017-11-21
Fix: Add infection as phony target

localheinz:
This PR

* [x] adds `infection` as a phony target to `Makefile`

Follows #606.


localheinz:
Thank you, @chartjes!

PR #626, opened 2017-11-21
Fix: Keep targets sorted in Makefile

localheinz:
This PR

* [x] keeps targets sorted in `Makefile`

Follows #606.


localheinz:
Thank you, @chartjes!

PR #627, opened 2017-11-21
Fix: Explicitly mark abstract test cases as abstract

localheinz:
This PR

* [x] explicitly marks abstract test cases as `abstract`


localheinz:
Thank you, @chartjes!

PR #628, opened 2017-11-21
Fix: Reduce visibility

localheinz:
This PR

* [x] reduces the visibility of `setUp()` and `tearDown()` to `protected`

💁‍♂️ This is the visibility as declared on the parent test cases, and there's no need to increase it.


localheinz:
Thank you, @chartjes!

PR #629, opened 2017-11-21
Enhancement: Split tests into integration and unit tests

localheinz:
This PR

* [x] moves tests into dedicated `Integration` and `Unit` namespaces
* [x] splits `phpunit.xml.dist` into dedicated configurations for integration and unit tests

💁‍♂️ The idea is to 

* make it easier (and faster) to run dedicated tests
* collect code coverage only when running unit tests.

For example, we can now run

```
$ make integration
```

and 

```
$ make unit
```

to run integration and unit tests, respectively.

Running

```
$ make unit
```

yields

```
composer install
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Nothing to install or update
Package guzzle/guzzle is abandoned, you should avoid using it. Use guzzlehttp/guzzle instead.
Generating autoload files
vendor/bin/phpunit --configuration tests/Unit/phpunit.xml.dist
PHPUnit 6.4.4 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.0.25
Configuration: /Users/am/Sites/opencfp/opencfp/tests/Unit/phpunit.xml.dist

........................................................................................................................................................                                                                                                                                                      152 / 152 (100%)

Time: 466 ms, Memory: 58.00MB

OK (152 tests, 218 assertions)
```

so, it's a lot faster to just run unit tests.

### Before

See https://travis-ci.org/opencfp/opencfp/builds/305249739:

screen shot 2017-11-21 at 13 34 59

### After

See https://travis-ci.org/opencfp/opencfp/builds/305251753:

screen shot 2017-11-21 at 13 35 15


BackEndTea:
Usually when i'm making something and writing tests for it, i just want to run those tests to see if i broke something before running the entire suite. So i do something like ```$ vendor/bin/phpunit tests/Domain/Model/TalkTest.php```. 

If we remove the ```phpunit.xml``` from the root will that still work?

localheinz:
@BackEndTea 

No, that won't work.

But how about running

```
$ make unit
```

instead?

BackEndTea:
@localheinz 

That works, but in that case someone still has to run all the unit tests, just to check if the one change they made in a specific class didn't break anything. 

If we had 0 database interaction in the unit tests it would be okay, but right now it takes roughly 30 seconds to run the tests locally. And when you just want to quickly check if the one string you changed is okay thats a bit much in my opinion.

localheinz:
@BackEndTea 

That's the idea, if we have database interactions then it's clearly not a unit test. The idea is to bring down running unit tests to hundreds or at most thousands of milliseconds, but not half a minute or more!

👍

localheinz:
@BackEndTea @chartjes 

If we want to merge this, here's what I suggest we should do afterwards:

* inspect test classes in `tests/Integration` and extract test methods therein into unit tests (if appropriate)
* inspect `BaseTestCase` and `WebTestCase` and see if they should be split as well (maybe some of the functionality is useful in the context of a unit test, but probably not all of it)

localheinz:
@chartjes 

Let me rebase real quick!

localheinz:
Thank you, @chartjes!

PR #630, opened 2017-11-21
Fix: Move test helper traits into OpenCFP\Test\Helper namespace

localheinz:
This PR

* [x] moves test helper traits into `OpenCFP\Test\Helper` namespace

Somewhat related to #629.

💁‍♂️ Now they can all hang out in the same namespace!

### DIFF
--- a/tests/BaseTestCase.php
+++ b/tests/BaseTestCase.php
@@ -5,6 +5,8 @@ namespace OpenCFP\Test;
 use Mockery;
 use OpenCFP\Application;
 use OpenCFP\Environment;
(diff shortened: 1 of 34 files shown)


BackEndTea:
I'd argue against naming it Util. Maybe something a bit more descriptive like ```TestHelper``` or just ```Helper```. Otherwise a 👍 for me

localheinz:
@BackEndTea 

`Helper` sounds good, happy to make the change!

localheinz:
Thank you, @chartjes!

PR #631, opened 2017-11-21
Enhancement: Separate path and environment from application

BackEndTea:
This PR separates the functionality some functionality out of the Application class

* [x] Create a Path class to handle paths
* [x] Update the Environment class to hold the is<environment> checks
* [x] Update all references to the old functions
* [x] Add unit tests for the moved functions

### DIFF
--- a/classes/Path.php
+++ b/classes/Path.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace OpenCFP;
+
+class Path
+{
+    private $path;
+    private $env;
+
+    public function __construct($path, $env)
+    {
+        $this->path = $path;
+        $this->env  = $env;
+    }
+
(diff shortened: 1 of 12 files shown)


BackEndTea:
After feedback from @localheinz in #635 i closed that and applied it to this branch

classes/Path.php
```diff
@@ -80,8 +75,6 @@ public function cacheTwigPath(): string
 
     /**
      * Get the HTML Purifier cache path.
```
localheinz: Maybe we can remove the entire docblock here (and elsewhere) as well?

They just repeat what is already obvious from the method name, and now that we have the return type declarations, they don't really add any value (only cost, because they need to be maintained). What do you think?

PR #634, opened 2017-11-21
Fix: Remove unused class

localheinz:
This PR

* [x] removes an unused class

Spotted in #629.


localheinz:
Thank you, @chartjes!

PR #635, opened 2017-11-21
Feature: different production path for uploaded files.

BackEndTea:
This PR is an extension of #631, as it splits up the uploads path getter for production.

On some production environments its not possible to have files persist on the disk.

By separating the production version of the uploads path this allows developers to still use the dev version locally, but not have to edit things every time a deployment happens.

Another possibility would be to move the uploads location to the config file. This would make it impossible to grab files from remote locations if the basePath function is used.

### DIFF
--- a/classes/Path.php
+++ b/classes/Path.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace OpenCFP;
+
+class Path
+{
+    private $path;
+    private $env;
+
+    public function __construct($path, $env)
+    {
+        $this->path = $path;
+        $this->env  = $env;
+    }
+
+    public function basePath()
+    {
+        return $this->path;
+    }
+
+    /**
+     * Get the configuration path.
+     *
+     * @return string
+     */
+    public function configPath(): string
+    {
+        return $this->basePath() . "/config/{$this->env}.yml";
+    }
+
+    /**
+     * Get the uploads path.
+     *
+     * @return string
+     */
+    public function uploadPath(): string
+    {
+        return $this->basePath() . '/web/uploads';
+    }
+
+    /**
+     * Get the templates path.
+     *
+     * @return string
+     */
+    public function templatesPath(): string
+    {
+        return $this->basePath() . '/resources/views';
+    }
+
+    /**
+     * Get the public path.
+     *
+     * @return string
+     */
+    public function publicPath(): string
+    {
+        return $this->basePath() . '/web';
+    }
+
+    /**
+     * Get the assets path.
+     *
+     * @return string
+     */
+    public function assetsPath(): string
+    {
+        return $this->basePath() . '/web/assets';
+    }
+
+    /**
+     * Get the Twig cache path.
+     *
+     * @return string
+     */
+    public function cacheTwigPath(): string
+    {
+        return $this->basePath() . '/cache/twig';
+    }
+
+    /**
+     * Get the HTML Purifier cache path.
+     *
+     * @return string
+     */
+    public function cachePurifierPath(): string
+    {
+        return $this->basePath() . '/cache/htmlpurifier';
+    }
+}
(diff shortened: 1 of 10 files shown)


classes/Path.php
```diff
@@ -0,0 +1,100 @@
+<?php
+
+namespace OpenCFP;
+
+class Path
```
localheinz: What do you think about marking this class as `final`?

classes/Path.php
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * The production path is separated to make it easier for prod environments where it is
+     * not possible to store files on the disk.
+     *
+     * @return string
+     */
+    public function uploadPath(): string
+    {
+        if ($this->env->isProduction()) {
+            return $this->basePath() . '/web/uploads';
```
localheinz: The way it looks right now, there's no difference between the results returned from this method, how about only adding the condition when we actually return a different value?

tests/PathTest.php
```diff
@@ -0,0 +1,137 @@
...
+
+namespace OpenCFP\Test;
+
+use OpenCFP\Environment;
+use OpenCFP\Path;
+
+/**
+ * @covers \OpenCFP\Path
+ */
+class PathTest extends BaseTestCase
```
localheinz: Do we really need to extend the `BaseTestCase` here? That is, are we using any of the functionality provided by it?

Also, what do you think about marking the test as `final`?

tests/PathTest.php
```diff
@@ -0,0 +1,137 @@
...
+{
+    /**
+     * @var Path
+     */
+    private $path;
+
+    public function setUp()
```
localheinz: Note #628.

tests/PathTest.php
```diff
@@ -0,0 +1,137 @@
...
+    {
+        parent::setUp();
+        $this->path = new Path('/home/folder/base', Environment::testing());
+    }
+
+    /**
+     * @test
```
localheinz: Not sure what the consensus is, but as of the moment it seems we have three different variations of test methods

* prefixed with `test` and using `camelCase`
* annotated with `@test` and using `camelCase`
* annotated with `@test` and using `snake_case`

Maybe we should make it consistent across the code base?

What do you think, @chartjes and @mdwheele (since you are the maintainers)?

tests/PathTest.php
```diff
@@ -0,0 +1,137 @@
...
+     */
+    public function configPathWorksForProdAndDev()
+    {
+        $prod = new Path('/home/folder/base', Environment::production());
+        $dev  = new Path('/home/folder/base', Environment::development());
+        $this->assertSame(
+            '/home/folder/base/config/production.yml',
+            $prod->configPath()
+        );
+        $this->assertSame(
```
localheinz: How about splitting this into two tests?

tests/PathTest.php
```diff
@@ -0,0 +1,137 @@
...
+            $this->path->uploadPath()
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function uploadPathReturnsUploadPathWhenProduction()
+    {
+        $path = new Path('/home/folder/base', Environment::production());
```
localheinz: Just food for thought, as I *used* to like setting up the system under test in `setUp()` - I stopped doing it because:

* I found tests became hard to understand (can't see everything in one place, need to go back to `setUp()`)
* when I end up with different scenarios, then I need to arrange individually anyway
* repetition in tests doesn't do any harm

Maybe this helps: https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests.

classes/Path.php
BackEndTea: I have no problem with that. I didn't do it since we have no ```final``` classes anywhere else in the project.

BackEndTea:
I'm closing this and applying the feedback to #631

PR #636, opened 2017-11-21
Enhancement: Reduce visibility

BackEndTea:
This PR changes a lot of protected variables to private

There were a bunch of protected variables that had no reason to be protected. 

The public variables in the ```BaseController``` are now protected, since there is no need for outside classes to call those.

### DIFF
--- a/classes/Application/Speakers.php
+++ b/classes/Application/Speakers.php
@@ -15,16 +15,16 @@ use OpenCFP\Domain\Talk\TalkWasSubmitted;
 class Speakers
 {
     /** @var CallForProposal */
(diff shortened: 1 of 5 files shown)


classes/Application/Speakers.php
```diff
@@ -15,16 +15,16 @@
 class Speakers
 {
     /** @var CallForProposal */
-    protected $callForProposal;
+    private $callForProposal;
```
localheinz: All of these pose potential BC breaks if any of the users of `opencfp/opencfp` extend any of these classes (which I don't know, of course).

Otherwise, totally 👍!

PR #637, opened 2017-11-21
Fix: Add type declarations

localheinz:
This PR

* [x] adds type and return type declarations

Somewhat related to #635.

/cc @BackEndTea


BackEndTea:
Looks good to me, was about to do this myself

localheinz:
Thank you, @chartjes!

PR #638, opened 2017-11-21
Fix: Cover equals()

localheinz:
This PR

* [x] adds missing tests for `Environment::equals()`


localheinz:
Thank you, @chartjes!

PR #639, opened 2017-11-21
Fix: No need to invoke method

localheinz:
This PR

* [x] removes an unnecessary method call

Blocked by #638.


localheinz:
Thank you, @chartjes!

PR #643, opened 2017-11-21
Fix: Remove comments that state the obvious

localheinz:
This PR

* [x] removes comments that state the obvious


localheinz:
Thank you, @chartjes!

PR #644, opened 2017-11-21
Enhancement: Enable phpdoc_order fixer

localheinz:
This PR

* [x] enables the `phpdoc_order` fixer
* [x] runs `make cs`

>**phpdoc_order**
>
>Annotations in phpdocs should be ordered so that param annotations come first, then throws annotations, then return annotations.


localheinz:
Thank you, @chartjes!

PR #645, opened 2017-11-21
Enhancement: Enable phpdoc_trim fixer

localheinz:
This PR

* [x] enables the `phpdoc_trim` fixer
* [x] runs `make cs`

>**phpdoc_trim** [`@Symfony`]
>
>Phpdocs should start and end with content, excluding the very first and last line of the docblocks.


localheinz:
Thank you, @chartjes!

PR #650, opened 2017-11-21
Fix: Specify language per stage

localheinz:
This PR

* [x] attempts to specify the build language per stage

💁‍♂️ Not sure if this works, let's find out.


localheinz:
Thank you, @chartjes!

PR #651, opened 2017-11-21
Fix: Run 'make cs'

localheinz:
This PR

* [x] runs `make cs` to fix the build

Follows #645.


localheinz:
@chartjes 

Looks like Code Climate is holding up, that's it/

localheinz:
Thank you, @chartjes!

PR #641, opened 2017-11-21
Enhancement: Enable php_unit_test_class_requires_covers fixer

localheinz:
This PR

* [x] enables the `php_unit_test_class_requires_covers` fixer
* [x] runs `make cs`

Somewhat related to #620.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.8.2#usage:

>**php_unit_test_class_requires_covers**
>
>Adds a default `@coversNothing` annotation to PHPUnit test classes that have no `@covers*` annotation.

❗️ Merging this probably only makes sense after #620 and #629.

PR #653, opened 2017-11-21
Fix: Replace coversNothing with appropriate covers annotations

localheinz:
❗️ Blocked by #651.

This PR

* [x] replaces `@coversNothing` with appropriate `@covers` annotations

Follows #641.


localheinz:
Thank you, @chartjes!

PR #655, opened 2017-11-21
Enhancement: Use MockeryPHPUnitIntegration trait

localheinz:
This PR

* [x] makes use of the `MockeryPHPUnitIntegration` trait instead of closing mockery ourselves


localheinz:
Thank you, @chartjes!

PR #656, opened 2017-11-21
Fix: Cover yet uncovered exceptions

localheinz:
This PR

* [x] adds tests for yet uncovered exceptions


localheinz:
Thank you, @chartjes!

PR #657, opened 2017-11-21
Fix: Move test into appropriate namespace

localheinz:
This PR

* [x] moves a test into the appropriate namespace

### DIFF
--- a/tests/Unit/Console/Command/AdminPromoteCommandTest.php
+++ b/tests/Unit/Console/Command/AdminPromoteCommandTest.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace OpenCFP\Test\Unit\Console;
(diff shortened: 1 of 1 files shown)


tests/Unit/Console/Command/AdminPromoteCommandTest.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace OpenCFP\Test\Unit\Console;
+namespace OpenCFP\Test\Unit\Console\Command;
```
localheinz: It covers `OpenCFP\Console\Command\AdminPromoteCommand`, so the corresponding unit test should reside in `OpenCFP\Test\Unit\Console\Command\AdminPromoteCommand`.

localheinz:
Thank you, @chartjes!

PR #658, opened 2017-11-21
Fix: Do not extend BaseTestCase if unit test

localheinz:
This PR

* [x] has a unit test extend from `PHPUnit\Framework\TestCase` instead of `OpenCFP\Test\BaseTestCase`

Follows https://github.com/opencfp/opencfp/pull/629#issuecomment-346014095.

💁‍♂️ Next up, moving `BaseTestCase` into the `OpenCFP\Test\Integration` namespace.


localheinz:
Thank you, @chartjes!

PR #659, opened 2017-11-22
Refactor the global error handler

BackEndTea:
This PR

* [x] Makes the ```registerGlobalErrorHandler``` function use $this to call the app instead of handing it to the function to use.


localheinz:
The build failure appears to be unrelated.

BackEndTea:
Yeah it looks like travis couldn't connect to mysql on the php 7 build. The other 2 passed just normally.

PR #663, opened 2017-11-22
Fix: Add test for AuthenticationException

localheinz:
This PR

* [x] adds a test for `OpenCFP\Domain\Services\AuthenticationException`

Follows #656.


localheinz:
Thank you, @chartjes!

PR #664, opened 2017-11-22
Fix: Simplify comparison

localheinz:
This PR

* [x] simplifies comparisons

Follows #631.


localheinz:
Thank you, @chartjes!

PR #665, opened 2017-11-22
Enhancement: Extract constants for environment types

localheinz:
This PR

* [x] extracts constants for environment types


localheinz:
Thank you, @chartjes!

PR #667, opened 2017-11-22
Fix: Rename field, parameter, variable

localheinz:
This PR

* [x] renames fields, parameters, and variables


localheinz:
Thank you, @chartjes!

PR #668, opened 2017-11-22
Enhancement: Use data provider for testing against different environment types

localheinz:
This PR

* [x] uses a data provider to test against different environment types


localheinz:
Thank you, @chartjes!

PR #670, opened 2017-11-22
Fix: Split and merge tests

localheinz:
This PR

* [x] splits and merges tests


localheinz:
Thank you, @chartjes!

PR #671, opened 2017-11-22
Fix: Add missing docblocks

localheinz:
This PR

* [x] adds missing docblocks

💁‍♂️ Docblocks add value as the exceptions thrown are now visible.

### DIFF
--- a/classes/Environment.php
+++ b/classes/Environment.php
@@ -15,6 +15,11 @@ class Environment
      */
     protected $type;
 
+    /**
+     * @param string $type
+     *
+     * @throws \InvalidArgumentException
+     */
     private function __construct(string $type)
     {
         if (! in_array($type, [self::TYPE_PRODUCTION, self::TYPE_DEVELOPMENT, self::TYPE_TESTING])) {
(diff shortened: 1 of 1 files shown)


classes/Environment.php
```diff
@@ -39,13 +44,25 @@ public static function testing(): self
         return new self(self::TYPE_TESTING);
     }
 
+    /**
+     * @throws \InvalidArgumentException
```
localheinz: Probably not quite appropriate to throw an `InvalidArgumentException` here, as the method itself doesn't accept any arguments.

classes/Environment.php
```diff
@@ -39,13 +44,25 @@ public static function testing(): self
...
+     *
+     * @return self
+     */
     public static function fromEnvironmentVariable(): self
```
localheinz: Maybe it makes sense to add an argument `$server` here?

Then throwing an `InvalidArgumentException` makes sense, and we don't need to access globals from within the method.

localheinz:
Thank you, @chartjes!

PR #672, opened 2017-11-22
Enhancement: Allow to pass in $_SERVER

localheinz:
This PR

* [x] allows to pass in `$_SERVER` to create an instance of `Environment` instead of accessing the superglobal from within

Follows https://github.com/opencfp/opencfp/pull/671#discussion_r152666225.


localheinz:
Thank you, @chartjes!

PR #673, opened 2017-11-22
Enhancement: Enable phpdoc_types_order fixer

localheinz:
This PR

* [x] enables the `phpdoc_types_order` fixer
* [x] runs `make cs`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.8.2#usage:

>**phpdoc_types_order**
>
>Sorts PHPDoc types.
>
>Configuration options:
>
>* `null_adjustment` (`'always_first'`, `'always_last'`, `'none'`): forces the position of null (overrides `sort_algorithm`); defaults to `'always_first'`
>* `sort_algorithm` (`'alpha'`, `'none'`): the sorting algorithm to apply; defaults to `'alpha'`

PR #674, opened 2017-11-22
Fix: True is not a type

localheinz:
This PR

* [x] adjusts a docblock that uses `true` as a type

Spotted in #673.


localheinz:
Thank you, @chartjes!

PR #675, opened 2017-11-22

localheinz:
This PR

* [x] uses return type declarations in test code where appropriate

### DIFF
--- a/tests/Helper/Faker/GeneratorTrait.php
+++ b/tests/Helper/Faker/GeneratorTrait.php
@@ -7,10 +7,7 @@ use Faker\Generator;
 
 trait GeneratorTrait
 {
-    /**
-     * @return Generator
-     */
-    protected function getFaker()
+    protected function getFaker(): Generator
     {
         static $faker;
 
--- a/tests/Helper/RefreshDatabase.php
+++ b/tests/Helper/RefreshDatabase.php
@@ -2,7 +2,7 @@
 
 namespace OpenCFP\Test\Helper;
 
-use Illuminate\Database\Capsule\Manager as Capsule;
+use Illuminate\Database\Capsule\Manager;
 
 trait RefreshDatabase
 {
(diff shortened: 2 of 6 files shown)


tests/Helper/RefreshDatabase.php
```diff
@@ -2,7 +2,7 @@
 
 namespace OpenCFP\Test\Helper;
 
-use Illuminate\Database\Capsule\Manager as Capsule;
```
localheinz: Removing an unnecessary alias here, too.

tests/Integration/Domain/Model/AirportTest.php
```diff
@@ -43,10 +43,7 @@ public function it_squawks_when_airport_is_not_found()
         $this->airports->withCode('foobarbaz');
     }
 
-    /**
-     * @return Airport
```
localheinz: Docblock doesn't add any value anymore.

tests/Helper/Faker/GeneratorTrait.php
```diff
@@ -7,10 +7,7 @@
 
 trait GeneratorTrait
 {
-    /**
-     * @return Generator
```
localheinz: Docblock doesn't add any value anymore.

tests/Unit/Http/Form/TalkFormTest.php
```diff
@@ -238,9 +238,9 @@ public function typeValidatesCorrectly($type, $expectedResponse)
     /**
      * Data provider for typeValidatesCorrectly
      *
-     * @return bool
```
localheinz: Type is invalid.

tests/Unit/Http/Form/TalkFormTest.php
```diff
@@ -281,9 +281,9 @@ public function categoryValidatesCorrectly($category, $expectedResponse)
...
```
localheinz: Type is invalid.

localheinz:
Thank you, @chartjes!

PR #677, opened 2017-11-22
Fix: Undefined class

localheinz:
This PR

* [x] fixes issues where classes are undefined


localheinz:
Thank you, @chartjes!

PR #678, opened 2017-11-22
Fix: Remove unneeded stub

localheinz:
This PR

* [x] removes an unneeded stub


localheinz:
Thank you, @chartjes!

Issue #618, opened 2017-11-16
Migration to Symfony 4


chartjes:
Silex is hitting EOL in June 2018, so I think it's time for us to figure out how to move OpenCFP more towards Symfony, perhaps targeting Symfony 4.

mdwheele:
Laravel! :grinning:

dbu:
i'd like to look into porting it to symfony 4 flex. no promises though, i'll try a bit and then report back if i think that i can seriously propose to do it, to check if that is whats wanted.

chartjes:
@dbu I definitely think Symfony 4 is the target to move to given how much Symfony stuff we are already relying on. Would greatly appreciate you taking a look

mdwheele:
Just as a heads up: I've started working on this. I've got the framework + some basic components (webpack-encore, twig, security) started. I'm working on pulling over our CI process and helping some things settle (the divergence between `.travis.yml`, `Makefile`, and `script/*`). I feel like TravisCI should be running (more or less) the same commands we are. If that happens to be `script/test`, cool. Currently, I've got to remember `make cs`, `script/setup` and `.travis.yml` is doing other things in `scripts`. Now's probably a good time to clean that up. 

Also pulling over all the `php-cs-fixer` / `.codeclimate.yml` stuff piece by piece.

I'm also including a simple LEMP stack on Docker that runs OpenCFP. You basically run `cfp shell` and it builds everything and drops you in a workspace to finish setting up the application.

I think that at *some point* it is going to be helpful to pump the breaks on PRs, but no need for that right now. I feel like the best option right now is to create a baseline, move over a few things, then we all agree to send PRs to a `symfony4` branch to move the rest of everything over. I think once I get the frontend working, complete with signup / login process handled, I'll put the branch up and we can all start migrating things in. 

This is also a GREAT time to create a decent service layer and inject dependencies into controllers rather than reaching out to the container constantly. Make it part of the "definition of done" for moving code into the new framework.

dbu:
sounds like a great plan! and glad you are on it. i don't think i would really find time to work on it anytime soon, after all... too much interesting stuff to do, too little time.

mdwheele:
That is exactly the problem I seem to find myself in pretty much all the time. Cheers!

chartjes:
@mdwheele The Symfony project has a page where they trying to put together a small guide to help people migrate. Anything you run into that is difficult should be reported here.

https://github.com/symfony/symfony-docs/issues/8678

[...]

PR #679, opened 2017-11-22

localheinz:
This PR

* [x] uses return type declarations in test code where appropriate

Follows #675.

### DIFF
--- a/tests/TestResponse.php
+++ b/tests/TestResponse.php
@@ -34,7 +34,7 @@ class TestResponse
         $this->app          = $app;
     }
 
-    public function assertSuccessful()
+    public function assertSuccessful(): self
     {
         Assert::assertTrue(
             $this->isSuccessful(),
@@ -44,7 +44,7 @@ class TestResponse
         return $this;
     }
 
-    public function assertStatus($status)
+    public function assertStatus($status): self
     {
         $actual = $this->getStatusCode();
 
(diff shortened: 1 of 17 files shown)


tests/TestResponse.php
```diff
@@ -34,7 +34,7 @@ public function __construct(Application $app, Response $response)
         $this->app          = $app;
     }
 
-    public function assertSuccessful()
+    public function assertSuccessful(): self
```
localheinz: Not entirely sure about these assertions in the first place, personally I would move the assertions to a trait and use the trait in tests instead, rather than decorating the response object.

tests/Unit/Console/Command/AdminPromoteCommandTest.php
```diff
@@ -80,15 +80,15 @@ public function promoteExistingNonAdminAccount()
...
+    protected function createInputInterfaceWithEmail($email): \Symfony\Component\Console\Input\InputInterface
     {
         $input = Mockery::mock(\Symfony\Component\Console\Input\InputInterface::class);
         $input->shouldReceive('getArgument')->with('email')->andReturn($email);
 
         return $input;
     }
 
-    protected function createOutputInterface()
+    protected function createOutputInterface(): \Symfony\Component\Console\Output\OutputInterface
```
localheinz: Instead of mocking `InputInterface` and `OutputInterface`, we could just use `Symfony\Component\Console\Tester\CommandTester`. I tried, but there's a lot of work required in regard to the application container being injected into the individual commands (rather than injecting dependencies into the commands and then pulling the commands from the container). Because of #618, I think it's not really worth the effort now, but suggest to sort this out anyway.

localheinz:
Thank you, @chartjes!

PR #680, opened 2017-11-22
Fix: Use type declarations

localheinz:
This PR

* [x] uses type declarations where possible


localheinz:
Thank you, @chartjes!

PR #681, opened 2017-11-22
Fix: Use constant

localheinz:
This PR

* [x] uses a constant instead of a magic number


localheinz:
Thank you, @chartjes!

PR #682, opened 2017-11-22
Enhancement: Enable and configure is_null fixer

localheinz:
This PR

* [x] enables and configures the `is_null` fixer
* [x] runs `make cs`

>**is_null** [`@Symfony:risky`]
>
>Replaces is_null(parameter) expression with `null === parameter`.
>
>Risky rule: risky when the function ``is_null()`` is overridden.
>
>Configuration options:
>
>* `use_yoda_style` (`bool`): whether Yoda style conditions should be used; defaults to `true`

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -22,6 +22,9 @@ $config = PhpCsFixer\Config::create()
             'spacing' => 'one',
         ],
         'increment_style' => true,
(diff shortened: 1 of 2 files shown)


localheinz:
Thank you, @chartjes!

PR #684, opened 2017-11-22
Fix: Cover named constructor

localheinz:
This PR

* [x] covers a named constructor

Follows #656.


localheinz:
Thank you, @chartjes!

PR #685, opened 2017-11-22

localheinz:
This PR

* [x] uses return type declarations

Follows #656.
Related to #684.


localheinz:
Thank you, @chartjes!

PR #686, opened 2017-11-22
Enhancement: Enable no_blank_lines_after_phpdoc fixer

localheinz:
This PR

* [x] enables the `no_blank_lines_after_phpdoc` fixer
* [x] runs `make cs`

>**no_blank_lines_after_phpdoc** [`@Symfony`]
>
>There should not be blank lines between docblock and the documented element.


localheinz:
Thank you, @chartjes!

PR #687, opened 2017-11-22
Enhancement: Enable no_useless_else fixer

localheinz:
This PR

* [x] enables the `no_useless_else` fixer
* [x] runs `make cs`

>**no_useless_else**
>
>There should not be useless else cases.


localheinz:
Thank you, @chartjes!

PR #692, opened 2017-11-23
Enhancement: Enable no_useless_return fixer

localheinz:
This PR

* [x] enables the `no_useless_return` fixer
* [x] runs `make cs`

>**no_useless_return**
>
>There should not be an empty return statement at the end of a function.


localheinz:
Thank you, @chartjes!

PR #695, opened 2017-11-23
Removed references to screenshots

chartjes:
This PR

* [x] updates README.md and removes references to screenshots

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,6 @@ Current release: v1.4.0
 ## README Contents
 
  * [Features](#features)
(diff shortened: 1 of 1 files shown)


BackEndTea:
Maybe we should add new screenshots of how it looks now. Though in that case i'd suggest adding those to a separate file. Takes a lot more scrolling than needed when you want to check something.

mdwheele:
👍 On an unrelated note... Firefox 57 is pretty amazing.

PR #700, opened 2017-11-24
DX: Rename config file for PHP CS Fixer

keradus:
same as for PHPUnit you use `phpunit.xml.dist` instead of `phpunit.xml`


chartjes:
Thank you @keradus

PR #702, opened 2017-11-24
Improve PHP CS Fixer <> Travis integration

keradus:
Depends on:
- [ ] #700

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ cache:
   yarn: true
   directories:
     - $HOME/.composer/cache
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -37,7 +37,9 @@ jobs:
         - mkdir -p "$HOME/.php-cs-fixer"
```
localheinz: @keradus 

We can probably also remove this then, and the corresponding caching directive?

chartjes:
Thanks @keradus

PR #703, opened 2017-11-24
CSRF additions

chartjes:
This PR

* [x] fixes some CSRF vulnerabilities

A helpful security audit from @enygma revealed a few spots where our forms and some GET actions with side effects were not protected using CSRF tokens. This PR fixes those spots and also adds CSRF protection to talk creation.

### DIFF
--- a/classes/Http/Controller/Admin/SpeakersController.php
+++ b/classes/Http/Controller/Admin/SpeakersController.php
@@ -12,6 +12,7 @@ use OpenCFP\Domain\Speaker\SpeakerProfile;
 use OpenCFP\Http\Controller\BaseController;
 use OpenCFP\Http\Controller\FlashableTrait;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Security\Csrf\CsrfToken;
 
 class SpeakersController extends BaseController
 {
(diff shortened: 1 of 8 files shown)


enygma:
secure-all-the-things

chartjes:
@localheinz @BackEndTea I think we need to change our tests runs on Travis to run all the tests because I am not getting proper code coverage values. I added tests to go with the new CSRF stuff I added in and being told coverage went down (which is not actually possible).

We should be running the ENTIRE test suite on Travis, and I don't think that is what is currently happening.

localheinz:
@chartjes 

We are only collecting test coverage for unit tests only to avoid collecting accidental coverage when running integration tests - that was the idea with #629.

classes/Http/Controller/Admin/SpeakersController.php
```diff
@@ -122,9 +123,15 @@ public function viewAction(Request $req)
 
     public function deleteAction(Request $req)
     {
+        $csrfTokenManager = $this->service('csrf.token_manager');
```
localheinz: What do you think about

* injecting services (rather than the container) so we can more easily write test
* sorting this out with a middleware instead

?

chartjes: No objections, can you point me at an example?

chartjes:
I admit I should've looked at #629 closer. I think that unit vs integration is okay but codeclimate doesn't let us do that.

classes/Http/Controller/Admin/SpeakersController.php
localheinz: I‘ll see if I can find out, but I remember @BackEndTea just recently moved authentication to a middleware, so I think it should be possible.

BackEndTea: I read some comment in i think the controller service provider that we cant really inject services the normal way because we override some of that. So fixing that would allow the service injection.

chartjes: So are we going to wait until we fix this or are you folks okay with pushing these commits because it fixes a security hole? We can definitely create an issue surrounding this

BackEndTea: I'm okay with pushing this to fix a security hole, creating an issue and then resolving that later.

PR #704, opened 2017-11-25
The contribution banner is now toggle-able

mdwheele:
This PR

* [x] Makes the contribution banner that shows in the top right toggle-able so that CFPs can decide whether or not to show.


chartjes:
Thanks @mdwheele

PR #705, opened 2017-11-25
Make the path for background configurable

mdwheele:
This PR

* [x] Allows a user to change the path used for the homepage hero section as well as the login / signup splash screens.

cats

### DIFF
--- a/config/development.dist.yml
+++ b/config/development.dist.yml
@@ -14,6 +14,7 @@ application:
   date_format: d/m/Y
   date_timezone: "UTC"
   coc_link: http://confcodeofconduct.com/
+  venue_image_path: /assets/img/venue.jpg
 
 api:
   enabled: true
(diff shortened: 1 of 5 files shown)


config/development.dist.yml
```diff
@@ -14,6 +14,7 @@ application:
   date_format: d/m/Y
   date_timezone: "UTC"
   coc_link: http://confcodeofconduct.com/
+  venue_image_path: /assets/img/venue.jpg
```
localheinz: Wondering, do you think it would be too difficult for users of OpenCFP to configure this on a SASS level?

Essentially, we could allow users to define corresponding variables in SASS, and if they aren’t defined, we just use defaults. This would then require them to recompile assets, though. What do you think?

mdwheele: That could be done. I'm not sure one is "better" than the other. In a perfect world, a lot of this would be part of a installation process supported by OpenCFP, not config files here and there. I hesitate to contribute that, considering upcoming plans.

chartjes:
Thanks @mdwheele

PR #706, opened 2017-11-25
Enhancement: Add asset target to Makefile

localheinz:
This PR

* [x] adds an `asset` target to `Makefile`

Somewhat related to https://github.com/opencfp/opencfp/pull/705#discussion_r153040895.


localheinz:
Interestingly, when I then run

```
$ make asset
```

the `NODE_ENV` variable does not get picked up properly.

PR #707, opened 2017-11-25
Fix: Remove unnecessary prefix

localheinz:
This PR

* [x] removes an unnecessary prefix

💁‍♂️ Not sure why it's there, but it works fine without it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,11 +2,11 @@
   "private": true,
   "scripts": {
     "dev": "npm run development",
-    "development": "node_modules/.bin/cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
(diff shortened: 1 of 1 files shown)


mdwheele:
As long as folks know to put `node_modules/.bin` in their path, then it's unnecessary. When I sent that, I was on a Windows machine that was all manner of broken. It's probably fine removing this and have reasonable assumptions about folks setting up their development environment correctly.

localheinz:
@mdwheele 

Just checked on https://docs.npmjs.com/cli/run-script:

>In addition to the shell's pre-existing `PATH`, `npm run` adds `node_modules/.bin` to the `PATH` provided to scripts. Any binaries provided by locally-installed dependencies can be used without the `node_modules/.bin` prefix. For example, if there is a `devDependency` on tap in your package, you should write:
>
>```
>"scripts": {"test": "tap test/\*.js"}
>```
>
>instead of
>
>```
>"scripts": {"test": "node_modules/.bin/tap test/\*.js"}  
>```
>
>to run your tests.

So unless everything is terribly broken on Windows, this should probably be good!

mdwheele:
Yeah, I think we're good. My honest advice to anybody doing development on Windows would be to use Vagrant or Docker and walk away. I would go check what version of Node/npm was on that machine, but it no longer exists.

PR #708, opened 2017-11-25
Fix: Run 'yarn run production'

localheinz:
This PR

* [x] runs `yarn run production`

💁‍♂️ When I run this on my machine, the permissions on the font files are changed from `755` to `644`


localheinz:
Thank you, @mdwheele!

PR #709, opened 2017-11-25
Enhancement: Allow to override hero image with SASS variable

localheinz:
This PR

* [x] allows to override the hero image with a SASS variable

An alternative to #705.


resources/views/home.twig
```diff
@@ -1,7 +1,7 @@
 {% extends "layouts/default.twig" %}
 
 {% block header %}
-    <header id="hero" class="bg-cover p-24" style="background:linear-gradient(rgba(0, 0, 0, 0.6), rgba(0, 0, 0, 0.6)), url(/assets/img/venue.jpg) no-repeat center -300px fixed;">
+    <header id="hero" class="opencfp-hero">
```
localheinz: Extracting a class `opencfp-hero` here so we don't need to apply multiple classes.

resources/assets/sass/app.scss
```diff
@@ -134,3 +143,14 @@ div:empty {
...
   }
 }
+
+//
+// Hero
+//
+
+.opencfp-hero {
+  @apply bg-cover;
+  @apply p-24;
```
localheinz: Applying the CSS classes found below.

resources/views/layouts/splash.twig
```diff
@@ -5,7 +5,7 @@
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <link href="/assets/css/app.css" rel="stylesheet" media="screen">
     </head>
-    <body class="h-full border-t-4 border-brand bg-cover flex items-center justify-center" style="background:linear-gradient(rgba(0, 0, 0, 0.7), rgba(0, 0, 0, 0.7)), url(/assets/img/venue.jpg) no-repeat center center fixed;">
+    <body class="opencfp-splash">
```
localheinz: Extracting a class `opencfp-splash` here so we don't need to apply multiple classes.

chartjes:
I propose closing this. As BFDL I prefer to not have this as a SASS variable.

mdwheele:
Like I mentioned in #705, I think it'd be nice to set some of this per-instance configuration through an interactive install process. That would include configuration of "spice" assets such as the splash and hero section. It could also include talk topics, reviewer/admin setup, etc. (i.e. pretty much everything in the current config files). That said, none of that exists and the closest we have are the config files.

What I *wouldn't* expect is an asset recompile as part of the install process, which would be required if the splash/hero sections are specified via Sass. 

I don't think the decision between #709 vs #705 matters so much, just throwing out some reasoning for attaching it to config vs asset compilation. I've been eye-ing better separation of instance-specific content and theming to not *force* people to maintain forks. We could get this to a point where that isn't necessary, I think.

PR #714, opened 2017-11-25
Fix: Mark ContainerAware trait as deprecated

localheinz:
This PR

* [x] marks the `ContainerAware` trait as deprecated

Somewhat related to #618.

💁‍♂️ In order to be able to properly test components currently using it, they should stop using it and have their dependencies injected. Marking the trait as `@deprecated` is a first step into that direction.

### DIFF
--- a/classes/ContainerAware.php
+++ b/classes/ContainerAware.php
@@ -2,6 +2,10 @@
 
 namespace OpenCFP;
 
+/**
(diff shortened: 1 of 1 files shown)


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #717, opened 2017-11-25
Enhancement: Decouple UserCreateCommand from application container

localheinz:
This PR

* [x] decouples the `UserCreateCommand` from the application container

Somewhat related to #714.
Somewhat related to #618.

### DIFF
--- a/bin/opencfp
+++ b/bin/opencfp
@@ -9,6 +9,8 @@ use OpenCFP\Environment;
 use OpenCFP\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
 use OpenCFP\Application as ApplicationContainer;
+use OpenCFP\Console\Command;
+use OpenCFP\Domain\Services;
 
 $basePath = realpath(dirname(__DIR__));
 $input = new ArgvInput();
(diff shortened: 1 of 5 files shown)


classes/Console/Command/UserCreateCommand.php
```diff
@@ -38,70 +49,63 @@ public function execute(InputInterface $input, OutputInterface $output)
...
-            return false;
+        if ($input->getOption('admin')) {
+            $roles[] = 'admin';
         }
-    }
 
-    private function promoteTo($user, $role = 'Admin')
-    {
-        if ($user->hasAccess(\strtolower($role))) {
-            $io->error(sprintf(
```
localheinz: Variable `$io` is undefined here.

classes/Console/Command/UserCreateCommand.php
```diff
@@ -38,70 +49,63 @@ public function execute(InputInterface $input, OutputInterface $output)
...
-                'Account with email %s already is in the Admin group.',
-                $email
```
localheinz: Variable `$email` is undefined here.

bin/opencfp
```diff
@@ -9,13 +9,19 @@ use OpenCFP\Environment;
...
 
 $basePath = realpath(dirname(__DIR__));
 $input = new ArgvInput();
 $environment = $input->getParameterOption(array('--env'), getenv('CFP_ENV') ?: 'development');
 
 $container = new ApplicationContainer($basePath, Environment::fromString($environment));
 $app = new Application($container);
+
+$app->addCommands([
+    new Command\UserCreateCommand($container[Services\AccountManagement::class]),
```
localheinz: Adding commands here for now, later we could also just fetch the application from the container, if we wanted to.

localheinz:
Thank you, @chartjes!

PR #718, opened 2017-11-26
Add excludes to code coverage

BackEndTea:
This PR

* [x] (Re-) adds excludes to the code coverage

This is how the set-up was before the split of Unit and Integration tests, and i see no reason not to exclude these from the total coverage.

I'm partial to adding Models to the excludes for coverage as well, since they would be impossible to test without using the database.

### DIFF
--- a/tests/Unit/phpunit.xml.dist
+++ b/tests/Unit/phpunit.xml.dist
@@ -19,6 +19,11 @@
     <filter>
         <whitelist>
             <directory>../../classes</directory>
+            <exclude>
+                <directory>../../classes/Provider</directory>
+                <directory>../../classes/Http/API</directory>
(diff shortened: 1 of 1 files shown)


localheinz:
@BackEndTea @mdwheele 

I'd suggest to remove dead code and not exclude anything from code coverage.

Excluding code from code coverage or static code analysis might give a false impression on the quality of code, and in the worst case we might miss opportunities to improve the quality.

mdwheele:
Hehe exactly. It will make it easier to *forget* that code.

BackEndTea:
I'm all for closing this and removing the dead code instead.

chartjes:
Closing and asking that the code we were excluding to be instead removed

PR #725, opened 2017-11-27
Fix: Check ratings a bit more strict

BackEndTea:
This PR

* [x] checks if the ratings given are numeric

Casting a string like 'blabla' to an int will make it 0 in php. Which isn't really what we want here.

### DIFF
--- a/classes/Http/Controller/Admin/TalksController.php
+++ b/classes/Http/Controller/Admin/TalksController.php
@@ -6,6 +6,7 @@ use OpenCFP\Domain\Services\Authentication;
 use OpenCFP\Domain\Services\Pagination;
 use OpenCFP\Domain\Talk\TalkFilter;
 use OpenCFP\Domain\Talk\TalkHandler;
+use OpenCFP\Domain\ValidationException;
 use OpenCFP\Http\Controller\BaseController;
 use OpenCFP\Http\Controller\FlashableTrait;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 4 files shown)


classes/Http/Controller/Admin/TalksController.php
```diff
@@ -74,6 +74,10 @@ public function viewAction(Request $req)
 
     public function rateAction(Request $req)
     {
+        if (!is_numeric($req->get('rating'))) {
```
localheinz: @BackEndTea 

I haven't looked into how filtering and validation is currently managed within the project, but if there's none, what do you think about using appropriate components sooner or later?

BackEndTea: The ```BaseController``` has a ```validate``` function, Which is used exactly once, in the ```processAction``` of the ```SignupController```.

localheinz: I was thinking more along the lines of creating specific validators for each use case.

BackEndTea: As in, some kind of abstract validation class/interface, and in this case a ```RatingValidation``` which checks if the request object fills the pre-defined criteria?

localheinz: Well, not exactly the entire request object, but the relevant request data?

BackEndTea: I wouldn't mind that, but its outside of the scope for this PR

chartjes:

PR #726, opened 2017-11-27
Enhancement: Enable  native_function_invocation fixer

BackEndTea:
This PR

* [x] enables the ```native_function_invocation``` fixer
* [x] runs ```make cs```

:tipping_hand_man: For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.8.2#usage:

> * ```native_function_invocation```
Add leading ```\``` before function invocation of internal function to speed up resolving.
*Risky rule: risky when any of the functions are overridden.*
Configuration options:
 ```exclude (array)```: list of functions to ignore; defaults to ```[]```

See also : https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -28,6 +28,7 @@ return PhpCsFixer\Config::create()
             'use_yoda_style' => false,
         ],
         'method_separation' => true,
+        'native_function_invocation' => true,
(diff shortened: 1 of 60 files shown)


chartjes:

PR #727, opened 2017-11-27
Bugfix: Allow talks to be deleted again

BackEndTea:
This PR

* [x] Fixes the bug that made a speaker unable to delete their own talk

The bug was introduced in https://github.com/opencfp/opencfp/commit/b0563ebef6acbac6a0ccb90750390a5a376c290b (by me)

### DIFF
--- a/classes/Http/Controller/TalkController.php
+++ b/classes/Http/Controller/TalkController.php
@@ -308,7 +308,7 @@ class TalkController extends BaseController
         }
 
         $userId = $this->service(Authentication::class)->userId();
-        $talk   = Talk::find($req->get('tid'), ['user_id']);
+        $talk   = Talk::find($req->get('tid'));
 
         if ((int) $talk->user_id !==  $userId) {
             return $this->app->json(['delete' => 'no']);


classes/Http/Controller/TalkController.php
```diff
@@ -308,7 +308,7 @@ public function deleteAction(Request $req)
         }
 
         $userId = $this->service(Authentication::class)->userId();
-        $talk   = Talk::find($req->get('tid'), ['user_id']);
+        $talk   = Talk::find($req->get('tid'), ['id', 'user_id']);
```
mdwheele: Is there any specific reason we're cherry-picking fields to be fetched for an Active Record? Seems error prone and possibly a source of bug/confusion for future contributors?

BackEndTea: I was under the impression it was slightly better for performance.

mdwheele: Hehe, sorry about that. I wasn't intending on making that a blocker for merge, but thanks for adjusting. I haven't looked around to see if we're doing this consistently, but would be willing to at least discuss performance merits of doing so. It's just something folks will have to keep in the back of their heads when doing things like exactly this or using model relationships. There's a cost to holding that in our heads that can be weighed against performance, for sure.

mdwheele:
Thanks @BackEndTea

BackEndTea:
Thanks @mdwheele

PR #728, opened 2017-11-27
Remove coveralls dev dependency

BackEndTea:
The dependency was added in the great silex 2.0 merge, yet i can't find
any place that it is actually implemented

This PR

* [x] Removes the ```php-coveralls``` dependency

As far as i could find it wasn't used anywhere.
```git grep coveralls``` didn't give any results besides the ```composer.json``` & ```composer.lock```.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,6 @@
         "fzaninotto/faker": "^1.5.0",
         "infection/infection": "^0.6.0",
         "mockery/mockery": "1.0.*@dev",
-        "php-coveralls/php-coveralls": "^1.0.2",
         "phpunit/phpunit": "^6.4.4",
         "phpunit/phpunit-mock-objects": "^4.0"
     },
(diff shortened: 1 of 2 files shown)


localheinz:
@BackEndTea 

Just a heads-up, the dependency was added in #701 to replace `satooshi/php-coveralls`.

BackEndTea:
@localheinz 
I wasn't able to locate ```satooshi/php-coveralls``` and it looks like it was renamed to ```php-coveralls/php-coveralls```. Since browsing to https://github.com/satooshi/php-coveralls just redirects you to the php-coveralls/php-coveralls repository.

localheinz:
@BackEndTea 

I think it can safely be removed since #562 introduced using https://github.com/codeclimate/test-reporter, rather than using the soon to be deprecated https://github.com/codeclimate/php-test-reporter.

chartjes:

PR #729, opened 2017-11-27
Enhancement: Enable combine_consecutive_unsets fixer

localheinz:
This PR

* [x] enables the `combine_consecutive_unsets` fixer
* [x] runs `make cs`

>**combine_consecutive_unsets**
>
>Calling `unset` on multiple items should be done in one call.


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #731, opened 2017-11-27
Enhancement: Enable phpdoc_no_alias_tag fixer

localheinz:
This PR

* [x] enables the `phpdoc_no_alias_tag` fixer
* [x] runs `make cs`

>**phpdoc_no_alias_tag** [`@Symfony`]
>
>No alias PHPDoc tags should be used.
>
>Configuration options:
>
>* `replacements` (`array`): mapping between replaced annotations with new ones; defaults to `['property-read' => 'property', 'property-write' => 'property', 'type' => 'var', 'link' => 'see']`


localheinz:
Thank you, @chartjes!

PR #732, opened 2017-11-27
Fix: Catch appropriate exception

localheinz:
This PR

* [x] catches the appropriate exception in `UserCreateCommand`

Follows #624.

PR #733, opened 2017-11-27
Fix: Use field directly

localheinz:
❗️ Blocked by #732.

This PR

* [x] uses a field directly


localheinz:
Thank you, @chartjes!

PR #735, opened 2017-11-27
Fix: Keep rules sorted

localheinz:
❗️ Blocked by #732.

This PR

* [x] keeps rules sorted in `.php_cs.dist`


localheinz:
Thank you, @chartjes!

PR #736, opened 2017-11-27
Enhancement: Enable phpdoc_annotation_without_dot fixer

localheinz:
This PR

* [x] enables the `phpdoc_annotation_without_dot` fixer
* [x] runs `make cs`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.8.2#usage:

>**phpdoc_annotation_without_dot** [`@Symfony`]
>
>Phpdocs annotation descriptions should not be a sentence.

PR #737, opened 2017-11-27
Fix: Docblock

localheinz:
This PR

* [x] adds a missing type to a docblock 

Spotted in #736.


classes/Domain/Speaker/SpeakerProfile.php
```diff
@@ -23,8 +23,8 @@ class SpeakerProfile
     private $hiddenProperties;
 
     /**
-     * @param $speaker
-     * @param array $hiddenProperties This is a blacklist, telling the view what fields it isn't allowed to show.
```
localheinz: I think we can also be a bit more specific here.

chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #738, opened 2017-11-27
Fix: Add return type declaration

localheinz:
This PR

* [x] adds return type declarations to `SpeakerRepository::findById()`

Spotted in #736.


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #739, opened 2017-11-27
Fix: Use || instead of in conditions

localheinz:
This PR

* [x] uses `||` instead of `or` in conditions

Spotted in #736.


localheinz:
Thank you, @chartjes!

PR #740, opened 2017-11-27
Reset password template was including a non-existant layout

chartjes:
This PR

* [x] fixes an error where the rest password template was including a non-existent layout

### DIFF
--- a/tests/Unit/Console/Command/UserCreateCommandTest.php
+++ b/tests/Unit/Console/Command/UserCreateCommandTest.php
@@ -141,6 +141,9 @@ final class UserCreateCommandTest extends Framework\TestCase
         $this->assertFalse($option->isArray());
     }
 
+    /**
+     * @expectedException OpenCFP\Infrastructure\Auth\UserExistsException
+     */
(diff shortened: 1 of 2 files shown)


tests/Unit/Console/Command/UserCreateCommandTest.php
```diff
@@ -141,6 +141,9 @@ public function testHasReviewerOption()
         $this->assertFalse($option->isArray());
     }
 
+    /**
+     * @expectedException OpenCFP\Infrastructure\Auth\UserExistsException
```
localheinz: @chartjes 

Can we remove this?

See #732.

chartjes: The test fails because it is not catching the exception. I am wondering why the test still fails

localheinz: How about rebasing on top of `master`?

BackEndTea: Won't this fail in the style stage since phpcs changes that to a ```$this->expectException```

PR #741, opened 2017-11-27
Fixed bad layout include

chartjes:
This PR

* [x] fixes wrong layout being included in the reset password form


chartjes:
@localheinz Does this look okay?

PR #742, opened 2017-11-27
Move Csrf check to middleware

BackEndTea:
This PR

* [x] Moves the Csrf Checks to middleware 

Follows: #703


chartjes:
This looks okay to me. @localheinz can you take a look?

classes/Infrastructure/Auth/CsrfCheck.php
```diff
@@ -0,0 +1,28 @@
...
+     * @var CsrfTokenManager
+     */
+    private $manager;
+
+    public function __construct(CsrfTokenManager $manager)
+    {
+        $this->manager = $manager;
+    }
+
+    public function checkCsrf($tokenId, $token)
```
localheinz: Can we add type declarations here, and if these are not sufficient, add a docblock as well?

classes/Provider/Gateways/WebGatewayProvider.php
```diff
@@ -67,12 +74,18 @@ public function boot(Application $app)
...
 
         // Talks
-        $web->get('/talk/edit/{id}', 'OpenCFP\Http\Controller\TalkController::editAction')->bind('talk_edit')->before($asSpeaker);
-        $web->get('/talk/create', 'OpenCFP\Http\Controller\TalkController::createAction')->bind('talk_new')->before($asSpeaker);
-        $web->post('/talk/create', 'OpenCFP\Http\Controller\TalkController::processCreateAction')->bind('talk_create')->before($asSpeaker);
-        $web->post('/talk/update', 'OpenCFP\Http\Controller\TalkController::updateAction')->bind('talk_update')->before($asSpeaker);
-        $web->post('/talk/delete', 'OpenCFP\Http\Controller\TalkController::deleteAction')->bind('talk_delete')->before($asSpeaker);
-        $web->get('/talk/{id}', 'OpenCFP\Http\Controller\TalkController::viewAction')->bind('talk_view')->before($asSpeaker);
+        $web->get('/talk/edit/{id}', 'OpenCFP\Http\Controller\TalkController::editAction')
+            ->bind('talk_edit')->before($asSpeaker)->before($csrfChecker);
```
localheinz: If we start wrapping, can we wrap all method calls?

localheinz:
@chartjes 

Taking a look now!

classes/Provider/ApplicationServiceProvider.php
```diff
@@ -48,6 +49,10 @@ public function register(Container $app)
             return new SentinelAuthentication($app[Sentinel::class], $app[AccountManagement::class]);
         };
 
+        $app[CsrfCheck::class] = function ($app) {
```
localheinz: Can we add a type declaration for the `$app` argument of the closure here?

classes/Infrastructure/Auth/CsrfCheck.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        $token = new CsrfToken($tokenId, $token);
+        if (!$this->manager->isTokenValid($token)) {
+            return new RedirectResponse('/dashboard');
```
localheinz: How about turning this entire class into a validator that has an `isValid()` method, which accepts a `Request` object, returns a `bool`?

The advantage would be that this class would have a single responsibility, which is to validate a request object, and not need to return a `RedirectResponse` here.

classes/Provider/Gateways/WebGatewayProvider.php
```diff
@@ -59,6 +60,12 @@ public function boot(Application $app)
             return SpeakerAccess::userHasAccess($app);
         };
 
+        $csrfChecker = function (Request $req) use ($app) {
```
localheinz: Can we rename `$req` to `$request` here? There's no real advantage of shortening it.

classes/Provider/Gateways/WebGatewayProvider.php
```diff
@@ -59,6 +60,12 @@ public function boot(Application $app)
...
+            $checker = $app[CsrfCheck::class];
+
+            return $checker->checkCsrf($req->get('token_id'), $req->get('token'));
```
localheinz: If we adjusted the `CsrfCheck` class as suggested above, we could modify the middleware to something like this:

```php
$csrfChecker = function(Request $request) use ($app) {
   $validator = $app[CsrfCheck::class];

    if (!$validator->isValid($request)) {
        return new RedirectResponse('/dashboard');
    }

    return $request;
}
```

This would have the advantage that the target of the redirection is not hard-coded into the validator, but hard-coded into the definition of the middleware, and it's more obvious that we either return a `RedirectResponse` here, or pass through the initial request.

classes/Provider/Gateways/WebGatewayProvider.php
BackEndTea: I wrapped them all

BackEndTea:
@localheinz 

I updated the CsrfValidation, could you give it another look?

classes/Domain/Services/Validator.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace OpenCFP\Domain\Services;
+
+use Symfony\Component\HttpFoundation\Request;
+
+interface Validator
+{
+    public function isValid(Request $request): bool;
```
localheinz: Not sure if this makes sense, are we going to implement this interface with that exact type declaration multiple times? Maybe rename it to `RequestValidator`, then?

What do you think?

classes/Provider/Gateways/WebGatewayProvider.php
localheinz: Ha, I meant more like wrap all of the *chained* method calls so they will always be on a new line, like this:

```php
$web->get('/talk/edit/{id}', 'OpenCFP\Http\Controller\TalkController::editAction')
    ->bind('talk_edit')
    ->before($asSpeaker)
    ->before($csrfChecker);
```

But, whatever @chartjes and @mdwheele prefer!

tests/Unit/Infrastructure/Auth/CsrfCheckTest.php
```diff
@@ -0,0 +1,41 @@
...
+    public function testIsFinal()
+    {
+        $reflection = new \ReflectionClass(CsrfValidator::class);
+        $this->assertTrue($reflection->isFinal());
+    }
+
+    public function testReturnsTrueWhenTokenMangerReturnsTrue()
+    {
+        $manager = Mockery::mock(CsrfTokenManager::class);
+        $manager->shouldReceive('isTokenValid')->andReturn(true);
```
localheinz: Can we also specify some expectations towards what is passed to `CsrfTokenManager::isValid()`?

More specifically, can we assert that the parameters

* `token`
* `token_id`

are fetched from the `Request` object, and that we pass an instance of `CsrfToken` to `isTokenValid()` which composes these values?

tests/WebTestCase.php
```diff
@@ -210,4 +212,13 @@ public function asReviewer(int $id = 1): self
 
         return $this;
     }
+
+    public function passCsrfValidator()
```
localheinz: Can we add a return type declaration as well?

Other than this, this entire thing looks great to me, @BackEndTea!

BackEndTea:
@localheinz 
Should be all good now.

BackEndTea:
Thanks @chartjes @localheinz

PR #744, opened 2017-11-27
Fix: Cover repository

localheinz:
This PR

* [x] adds tests for the `IlluminateSpeakerRepository`

### DIFF
--- a/tests/Unit/Infrastructure/Persistence/IlluminateSpeakerRepositoryTest.php
+++ b/tests/Unit/Infrastructure/Persistence/IlluminateSpeakerRepositoryTest.php
@@ -1,0 +1,96 @@
+<?php
+
+namespace OpenCFP\Test\Unit\Infrastructure\Persistence;
+
+use Illuminate\Database\Eloquent;
+use OpenCFP\Domain\EntityNotFoundException;
+use OpenCFP\Domain\Model;
+use OpenCFP\Domain\Speaker;
+use OpenCFP\Infrastructure\Persistence\IlluminateSpeakerRepository;
+use OpenCFP\Test\Helper\Faker\GeneratorTrait;
+use PHPUnit\Framework;
+
+/**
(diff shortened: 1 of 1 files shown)


tests/Unit/Infrastructure/Persistence/IlluminateSpeakerRepositoryTest.php
```diff
@@ -0,0 +1,96 @@
...
+            'save',
+        ]);
+
+        $user
+            ->expects($this->once())
+            ->method('save');
+
+        $repository = new IlluminateSpeakerRepository($this->createUserMock());
+
+        $repository->persist($user);
```
localheinz: I am not sure what [`IlluminateSpeakerRepository::persist()`](https://github.com/opencfp/opencfp/blob/70b5aff3fb766b38642a6f2ae3f350871d740fae/classes/Infrastructure/Persistence/IlluminateSpeakerRepository.php#L49-L52) is supposed to return, as the method `Model\User::save()` appears to be a magic method.

chartjes: If I recall the pattern correctly, the persist() method should return a boolean value based on whether or not it was able to create a record in the database for $user

BackEndTea: yes the [save()](https://laravel.com/api/5.5/Illuminate/Database/Eloquent/Model.html#method_save function returns a boolean, whether it was successful or not.

localheinz:
Thank you, @chartjes!

tests/Unit/Infrastructure/Persistence/IlluminateSpeakerRepositoryTest.php
localheinz: Thank you, @BackEndTea! 

I don't think we really need to return a `bool` here, what do you think?

PR #745, opened 2017-11-27
Fix: Cover repository

localheinz:
This PR

* [x] covers the `IlluminateTalkRepository`

Follows #744.

### DIFF
--- a/tests/Unit/Infrastructure/Persistence/IlluminateTalkRepositoryTest.php
+++ b/tests/Unit/Infrastructure/Persistence/IlluminateTalkRepositoryTest.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace OpenCFP\Test\Unit\Infrastructure\Persistence;
+
+use OpenCFP\Domain\Model;
+use OpenCFP\Domain\Talk;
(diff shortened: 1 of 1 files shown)


tests/Unit/Infrastructure/Persistence/IlluminateTalkRepositoryTest.php
```diff
@@ -0,0 +1,52 @@
...
+            'save',
+        ]);
+
+        $talk
+            ->expects($this->once())
+            ->method('save');
+
+        $repository = new IlluminateTalkRepository();
+
+        $repository->persist($talk);
```
localheinz: I am not sure what [`IlluminateTalkRepository::persist()`](https://github.com/opencfp/opencfp/blob/70b5aff3fb766b38642a6f2ae3f350871d740fae/classes/Infrastructure/Persistence/IlluminateTalkRepository.php#L17) is supposed to return, as the method `Model\Talk::save()` appears to be a magic method.

chartjes: Yeah, same issue as I indicated in #744

localheinz:
Thank you, @chartjes!

PR #746, opened 2017-11-27
Fix: Add type declarations

localheinz:
This PR

* [x] adds type declarations

Follows #744.
Follows #745.


localheinz:
Thank you, @chartjes!

PR #747, opened 2017-11-27
Fix: Do not return boolean from persist()

localheinz:
This PR

* [x] stops passing through what `save()` returns

Follows https://github.com/opencfp/opencfp/pull/744#discussion_r153315056.

### DIFF
--- a/classes/Domain/Talk/TalkRepository.php
+++ b/classes/Domain/Talk/TalkRepository.php
@@ -6,8 +6,6 @@ interface TalkRepository
 {
     /**
      * @param $talk
-     *
-     * @return mixed
      */
     public function persist($talk);
 }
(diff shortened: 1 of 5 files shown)


classes/Domain/Talk/TalkRepository.php
```diff
@@ -6,8 +6,6 @@
 {
     /**
      * @param $talk
```
localheinz: Together with #746, we can probably remove the entire docblock.

classes/Infrastructure/Persistence/IlluminateTalkRepository.php
```diff
@@ -9,11 +9,9 @@ class IlluminateTalkRepository implements TalkRepository
 {
     /**
      * @param Talk $talk
```
localheinz: Together with #746, we can probably remove the entire docblock.

localheinz:
Thank you, @chartjes!

PR #750, opened 2017-11-27
Fix: Path should be a string

localheinz:
This PR

* [x] uses a type declaration and removes a docblock from the constructor of `OpenCFP\Application`


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #751, opened 2017-11-27
Fix: Remove dependency on BASE_PATH constant

localheinz:
This PR

* [x] removes the dependency on the `BASE_PATH` constant in integration and unit tests

💁‍♂️ Currently cannot run a single unit tests from within PhpStorm without configuring the working directory. Also, PhpStorm doesn't provide support for constants declared within `phpunit.xml`, so this is much more helpful.

### DIFF
--- a/tests/BaseTestCase.php
+++ b/tests/BaseTestCase.php
@@ -40,7 +40,7 @@ abstract class BaseTestCase extends \PHPUnit\Framework\TestCase
 
     public function createApplication(): Application
     {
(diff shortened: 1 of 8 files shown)


chartjes:
@localheinz I understand why we should remove it (being more PhpStorm friendly is a good thing) but it just *looks* so ugly with __DIR__ and '../../../../..'

chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #752, opened 2017-11-28
Fix: capitalization error

BackEndTea:
This PR

* [x] Fixes a capitalisation error in the reviewer dashboard controller filename & class name
 
Some deployment environments would error because they couldn't find the class as described in the web gateway provider.

This also fixes the reviewers dashboard controller test which was checking the admin controller instead.


mdwheele:
Thanks @BackEndTea

BackEndTea:
Thanks @mdwheele

PR #753, opened 2017-11-28
Enhancement: Use sentinel cookie

BackEndTea:
This PR

* [x] Updates Sentinel to ```v2.0.17```
* [x] Uses Sentinels ```NullCookie``` instead of our own 
* [x] Deletes our own ```NullCookie``` and its tests

Follows #624 

The reason it was introduced in #624 was because Sentinel didn't have ```NullCookie``` in its core. But now it does, so there is no reason to use our own version.


BackEndTea:
@localheinz 

Done, should be explicit in the ```composer.json``` now.

chartjes:
Thank you @BackEndTea and @localheinz

BackEndTea:
Thanks @chartjes

localheinz:
@BackEndTea 

Looks like 

```
$ composer update
```

was run, effectively updating **all** dependencies. I would always suggest to only run 

```
$ composer update
```

in a Continuous Integration environment, when we want to test whether an application or librarystill works with the highest dependencies available.

In all other cases, I would only update a single dependency by running

```
$ composer require foo/bar:^x.y.z
```

or - if need be - by running

```
$ composer require foo/bar:^x.y.z --update-with-dependencies
```

The problem we have here is that we have a few unbound version constraints, see

```
$ composer validate

require.ezyang/htmlpurifier : unbound version constraints (dev-master) should be avoided
require.robmorgan/phinx : unbound version constraints (*) should be avoided
require.swiftmailer/swiftmailer : exact version constraints (5.0) should be avoided if the package follows semantic versioning
```

and even if we didn't, I can tell from experience that even updating a library that follows semantic versioning with insufficient test coverage might break something in production.

BackEndTea:
It looks like i messed up in #754, if needed that can be reverted, and i'll fix it.

localheinz:
@BackEndTea 

Ha, no worries, should be all good! Just noticed after rebasing #755 on top of `master` that a lot of dependencies changed.

PR #755, opened 2017-11-28
Fix: Cover SpeakerProfile

localheinz:
This PR

* [x] covers `SpeakerProfile` with unit tests


tests/Unit/Domain/Speaker/SpeakerProfileTest.php
```diff
@@ -0,0 +1,627 @@
...
+        $speaker = $this->createUserMock([
+            'has_made_profile' => $hasMadeProfile,
+        ]);
+
+        $profile = new SpeakerProfile($speaker);
+
+        $this->assertFalse($profile->needsProfile());
+    }
+
+    public function providerHasMadeProfile(): \Generator
```
localheinz: I'm using a data provider here (and in the test below) because we are using a non-strict comparison in [`SpeakerProfile::needsProfile()`](https://github.com/opencfp/opencfp/blob/40ce73ce014afec8faf92443401688ee8975d6b9/classes/Domain/Speaker/SpeakerProfile.php#L47-L50).

Ideally we would be using a strict comparison, and then the data provider would become obsolete.

tests/Unit/Domain/Speaker/SpeakerProfileTest.php
```diff
@@ -0,0 +1,627 @@
...
+        $speaker = $this->createUserMock([
+            'transportation' => $transportation,
+        ]);
+
+        $profile = new SpeakerProfile($speaker);
+
+        $this->assertFalse($profile->getTransportation());
+    }
+
+    public function providerDoesNotNeedTransportation(): \Generator
```
localheinz: I'm using a data provider here because we are using a non-strict comparison in [`SpeakerProfile::getTransportation()`](https://github.com/opencfp/opencfp/blob/40ce73ce014afec8faf92443401688ee8975d6b9/classes/Domain/Speaker/SpeakerProfile.php#L113-L118).

Ideally we would be using a strict comparison, and then the data provider would become obsolete.

chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #756, opened 2017-11-28
Enhancement: Enable header_comment fixer

localheinz:
This PR

* [x] enables the `header_comment` fixer
* [x] runs `make cs`

>**header_comment**
>
>Add, replace or remove header comment.
>
>Configuration options:
>
>* `commentType` (`'comment'`, `'PHPDoc'`): comment syntax type; defaults to `'comment'`
* `header` (`string`): proper header content; required
* `location` (`'after_declare_strict'`, `'after_open'`): the location of the inserted header; defaults to `'after_declare_strict'`
* `separate` (`'both'`, `'bottom'`, `'none'`, `'top'`): whether the header should be separated from the file content with a new line; defaults to `'both'`

@chartjes @mdwheele 

If you want to enable this, this mostly required input from your side what you would like the header comment to look like.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,5 +1,16 @@
 <?php
 
+$year = date('Y');
+
+$header = <<<TXT
+Copyright (c) 2013-$year OpenCFP
+
+For the full copyright and license information, please view
+the LICENSE file that was distributed with this source code.
+
+@see https://github.com/opencfp/opencfp
+TXT;
+
 $finder = PhpCsFixer\Finder::create()
     ->exclude('migrations')
     ->in(__DIR__);
@@ -24,6 +35,10 @@ return PhpCsFixer\Config::create()
         ],
         'function_to_constant' => true,
         'function_typehint_space' => true,
+        'header_comment' => [
+            'commentType' => 'PHPDoc',
+            'header' => $header,
+        ],
         'increment_style' => true,
         'is_null' => [
             'use_yoda_style' => false,
--- a/classes/Application.php
+++ b/classes/Application.php
@@ -1,5 +1,14 @@
 <?php
 
+/**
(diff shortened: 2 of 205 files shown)


chartjes:
Would this mean that it would enforce there being a header docblook in every single file?

localheinz:
@chartjes 

> Would this mean that it would enforce there being a header docblook in every single file?

Exactly!

chartjes:
@localheinz I'm not sure I want that. What is the normal justification for this?

localheinz:
@chartjes 

Ha, no problem, it was just a suggestion!I think most of the open source projects do have a file-level header, though, so one can at least see where the thing is coming from.

As an example, I have worked on projects before where dependencies had been checked in into version control, rather than using `composer`. Thanks to a file-level header it was a bit easier to find out which one it is and then require it via `composer`.

.php_cs.dist
```diff
@@ -1,5 +1,14 @@
 <?php
 
+$header = <<<TXT
+Copyright (c) TBD
```
localheinz: @chartjes 

We can easily specify the year here using `date('Y')`, if you like!

localheinz: Might just be a bit of an issue when people contribute after midnight in Europe, hehe, when you guys are still preparing to celebrate a new year.

🎉

chartjes: @localheinz If we add in `date('Y')` then I am okay with this

.php_cs.dist
```diff
@@ -1,5 +1,16 @@
 <?php
 
+$year = date('Y');
+
+$header = <<<TXT
+Copyright (c) 2013-$year OpenCFP
```
localheinz: @chartjes 

Would this be accurate?

localheinz:
Thank you, @chartjes!

PR #757, opened 2017-11-28
Enhancement: Render email in login form

localheinz:
This PR

* [x] renders the email address in the login form

💁‍♂️ This has the advantage that a) I don't need to type in my email again if a login failed, and b) that I can see whatever the email address is I typed in earlier.


mdwheele:
:heart:

localheinz:
Thank you, @mdwheele!

PR #759, opened 2017-11-28
Fix: Add docblocks

localheinz:
This PR

* [x] adds docblocks to `SpeakerProfile` where necessary

Follows #755.


localheinz:
Thank you, @chartjes!

PR #760, opened 2017-11-28
Fix: Add return type declarations

localheinz:
This PR

* [x] adds return type declarations to `SpeakerProfile` where possible

Follows #755.


localheinz:
Thank you, @mdwheele!

PR #761, opened 2017-11-28
Fix: Remove unnecessary ternary

localheinz:
This PR

* [x] removes an unnecessary ternary

Follows #755.


localheinz:
Thank you, @chartjes!

PR #766, opened 2017-11-28
Enhancement: Enable declare_strict_types fixer

localheinz:
This PR

* [x] enables the `declare_strict_types` fixer
* [x] runs `make cs`
* [x] fixes failing tests

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.8.3#usage:

>**declare_strict_types** [`@PHP70Migration:risky`, `@PHP71Migration:risky`]
>
>Force strict types declaration in all files. Requires PHP >= 7.0.
>
>*Risky rule: forcing strict types will stop non strict code from working.*

PR #767, opened 2017-11-28
Enhancement: Extract dedicated constructors

localheinz:
This PR

* [x] extracts dedicated named constructors for `UserNotFoundException`

Related to #766.


localheinz:
Thank you, @BackEndTea!

PR #768, opened 2017-11-28
Fix: Remove default value

localheinz:
This PR

* [x] removes a default value from `AccountManagement::promoteTo()` and `AccountManagement::demoteFrom()`

💁‍♂️ `Admin` doesn't really seem like a default value that makes sense.


localheinz:
@mdwheele 

Conflicts have been resolved!

PR #769, opened 2017-11-28
Fix: Remove cartalyst/sentry and code referencing it

localheinz:
This PR

* [x] removes `cartalyst/sentry` and code referencing it

Follows #624.


localheinz:
Thank you, @chartjes!

PR #772, opened 2017-11-29
Fix return type of speaker profile

BackEndTea:
This PR

* [x] Makes sure the ```getTalks``` function returns an ```array```

Without the toArray it would return a collection, and throw a fatal error.

Follows #760

### DIFF
--- a/classes/Domain/Speaker/SpeakerProfile.php
+++ b/classes/Domain/Speaker/SpeakerProfile.php
@@ -69,7 +69,7 @@ class SpeakerProfile
     {
         $this->assertAllowedToSee('talks');
 
-        return $this->speaker->talks;
+        return $this->speaker->talks->toArray();
     }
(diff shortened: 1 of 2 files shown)


classes/Domain/Speaker/SpeakerProfile.php
```diff
@@ -63,9 +64,9 @@ public function needsProfile(): bool
      *
      * @throws NotAllowedException
      *
-     * @return Talk[]
+     * @return Collection|Talk[]
      */
-    public function getTalks(): array
+    public function getTalks(): Collection
```
localheinz: How about we leave it to `array` and instead enforce that an array is returned?

This has the advantage that we don't expose the collection, and thus, it can't be modified.

BackEndTea:
Thank you @localheinz

PR #775, opened 2017-11-29
Fix: Use camelCase for fields and variables

localheinz:
This PR

* [x] uses `camelCase` for fields and variables

### DIFF
--- a/classes/Domain/Services/ResetEmailer.php
+++ b/classes/Domain/Services/ResetEmailer.php
@@ -18,7 +18,7 @@ class ResetEmailer
     /**
      * @var \Swift_Mailer
      */
-    private $swift_mailer;
+    private $swiftMailer;
 
     /**
      * @var \Twig_Template
@@ -28,12 +28,12 @@ class ResetEmailer
     /**
      * @var string
      */
-    private $config_email;
+    private $configEmail;
 
     /**
      * @var string
      */
-    private $config_title;
+    private $configTitle;
 
     /**
      * @param \Swift_Mailer  $swiftMailer
@@ -43,10 +43,10 @@ class ResetEmailer
      */
     public function __construct(\Swift_Mailer $swiftMailer, \Twig_Template $template, $configEmail, $configTitle)
     {
-        $this->swift_mailer = $swiftMailer;
+        $this->swiftMailer  = $swiftMailer;
         $this->template     = $template;
-        $this->config_email = $configEmail;
-        $this->config_title = $configTitle;
+        $this->configEmail  = $configEmail;
+        $this->configTitle  = $configTitle;
     }
 
     /**
@@ -63,7 +63,7 @@ class ResetEmailer
         try {
             $message = $this->preparedMessage($email, $parameters);
 
-            return $this->swift_mailer->send($message);
+            return $this->swiftMailer->send($message);
         } catch (\Exception $e) {
             echo $e;
             die();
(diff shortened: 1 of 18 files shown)


classes/Http/Form/Form.php
```diff
@@ -13,12 +13,12 @@
...
-    protected $_cleanData;
-    protected $_taintedData;
-    protected $_purifier;
-    protected $_fieldList = [];
+    protected $options;
+    protected $messages;
+    protected $cleanData;
+    protected $taintedData;
+    protected $purifier;
+    protected $fieldList = [];
```
localheinz: These changes are potential BC breaks, if anyone else extends `Form`

mdwheele: I think that's fine and arguably even a non-issue.

classes/Http/Form/Form.php
```diff
@@ -167,8 +167,8 @@ public function hasErrors(): bool
      */
     protected function _addErrorMessage($message)
```
mdwheele: Should we drop underscore here?

classes/Domain/Services/ResetEmailer.php
```diff
@@ -82,8 +82,8 @@ private function parameters($userId, $resetCode)
                 ? 'https' : 'http',
             'host'    => !empty($_SERVER['HTTP_HOST']) ? $_SERVER['HTTP_HOST'] : 'localhost',
```
mdwheele: `phpmd` complains about use of `$_SERVER`. We *could* inject an instance of `Request` here but... I don't know that I care. What do you think? And if we don't care, do we want to create an issue to adjust `phpmd` to slack-off?

classes/Http/Form/SignupForm.php
```diff
@@ -43,44 +43,44 @@ class SignupForm extends Form
...
+        $ValidSpeakerBio    = true;
+
+        if (!empty($this->taintedData['speaker_info'])) {
+            $validSpeakerInfo = $this->validateSpeakerInfo();
         }
 
-        if (!empty($this->_taintedData['speaker_bio'])) {
-            $valid_speaker_bio = $this->validateSpeakerBio();
+        if (!empty($this->taintedData['speaker_bio'])) {
+            $ValidSpeakerBio = $this->validateSpeakerBio();
```
mdwheele: Lowercase the `V` in `$ValidSpeakerBio`

classes/Http/Form/SignupForm.php
```diff
@@ -43,44 +43,44 @@ class SignupForm extends Form
...
-            $agree_coc;
+            $validEmail         &&
+            $validPasswords     &&
+            $validFirstName     &&
+            $validLastName      &&
+            $validCompany       &&
+            $validTwitter       &&
+            $validUrl           &&
+            $validSpeakerInfo   &&
+            $ValidSpeakerBio    &&
```
mdwheele: `$validSpeakerBio`

classes/Domain/Services/ResetEmailer.php
localheinz: How about in a separate PR?

BackEndTea: The silex Application object has the ```requireHttps()``` function we could use, and the host can be retrieved from the config file ( i think). So no need to use the ```Request``` object.

localheinz:
Thank you, @mdwheele!

classes/Domain/Services/ResetEmailer.php
localheinz: Sounds like a good plan to me, @BackEndTea!

PR #776, opened 2017-11-29
Enhancement: Decouple ReviewerDemoteCommand from application container

localheinz:
This PR

* [x] decouples the `ReviewerDemoteCommand` from the application container

Follows #717.


localheinz:
@BackEndTea 

Absolutely!

For now my aim is to decouple all of the commands from the  `BaseCommand` and the dependency on the application container, but I definitely want to follow up sorting out duplicated code.

localheinz:
Thank you, @BackEndTea!

PR #777, opened 2017-11-29
Enhancement: Do not exclude migrations from php-cs-fixer

localheinz:
This PR

* [x] removes `migrations/` from exclusions for `php-cs-fixer`
* [x] runs `make cs`

💁‍♂️ I can't think of a reason why the same rules should not apply to migrations.


localheinz:
Thank you, @chartjes!

PR #778, opened 2017-11-29
Fix: Name of test class

localheinz:
This PR

* [x] fixes the name of a test class

💁‍♂️ The system under test is the [`SpeakersController`](https://github.com/opencfp/opencfp/blob/b607dc45cb0c58edaec1605d4ef61e18207cd44c/classes/Http/Controller/Reviewer/SpeakersController.php).


localheinz:

PR #779, opened 2017-11-29
Fix: Hotel is actually a boolean

localheinz:
This PR

* [x] adjusts `SpeakerProfile::getHotel()` to return a `bool`


localheinz:
Thank you, @chartjes!

PR #758, opened 2017-11-28
Fix: Compare strictly

localheinz:
This PR

* [x] performs strict comparions in `SpeakerProfile`

Follows #755.

PR #780, opened 2017-11-29
Fix: Use non-nullable TINYINT for boolean fields

localheinz:
This PR

* [x] adds a migration to turn fields which represent `bool`s in the `users` table into non-nullable `TINYINT` fields

Follows #779.
Follows https://github.com/opencfp/opencfp/pull/758#discussion_r153525414.


localheinz:
Thank you, @chartjes!

PR #781, opened 2017-11-29
Fix: Allow login/signup forms to be responsive

localheinz:
This PR

* [x] allows login and signup forms to be responsive

💁‍♂️ For reference, see https://tailwindcss.com/docs/adding-new-utilities#making-custom-utilities-responsive.

### Before

screen shot 2017-11-29 at 16 44 16

### After

screen shot 2017-11-29 at 16 44 08


resources/assets/sass/app.scss
```diff
@@ -134,3 +134,14 @@ div:empty {
     @apply .font-bold;
   }
 }
+
+//
+// Login / SignUp
+//
+@responsive {
+  .opencfp-login,
+  .opencfp-signup {
```
localheinz: Extracting two different classes here, as we might want to address these forms separately at some point in time.

localheinz:
Thank you, @mdwheele!

PR #782, opened 2017-11-29
Fix: Rename CallForProposals to CallForPapers

localheinz:
This PR

* [x] renames `CallForProposals` to `CallForPapers`

💁‍♂️ Running

```
$ git grep -i callforproposal
```

after this change yields nothing now.


localheinz:
Thank you, @chartjes!

PR #783, opened 2017-11-29
Fix: Run 'make cs'

localheinz:
This PR

* [x] runs `make cs`

Follows #777.
Follows #780.


localheinz:
Thank you, @chartjes!

PR #784, opened 2017-11-29
Fix: Remove unnecessary assignment

localheinz:
This PR

* [x] removes an unnecessary assignment


localheinz:
Thank you, @chartjes!

PR #785, opened 2017-11-29
Fix: Extract template map from switch statement

localheinz:
This PR

* [x] extracts a map of HTTP status codes to template scripts from a `switch` statement


localheinz:
Thank you, @BackEndTea!

PR #786, opened 2017-11-29
Enhancement: Decouple AdminPromoteCommand from application container

localheinz:
This PR

* [x] decouples the `AdminPromoteCommand` from the application container

Follows #717.


localheinz:
Thank you, @BackEndTea!

PR #790, opened 2017-11-30
Enhancement: Decouple AdminDemoteCommand from application container

localheinz:
This PR

* [x] decouples the `AdminDemoteCommand` from the application container

Follows #717.

### DIFF
--- a/classes/Console/Command/AdminDemoteCommand.php
+++ b/classes/Console/Command/AdminDemoteCommand.php
@@ -13,44 +13,80 @@ declare(strict_types=1);
 
 namespace OpenCFP\Console\Command;
 
-use OpenCFP\Console\BaseCommand;
+use OpenCFP\Domain\Services;
+use OpenCFP\Infrastructure\Auth;
+use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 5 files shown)


classes/Console/Command/AdminDemoteCommand.php
```diff
@@ -13,44 +13,80 @@
...
+        } catch (Auth\UserNotFoundException $exception) {
+            $io->error(\sprintf(
+                'Could not find account with email "%s".',
+                $email
+            ));
+
+            return 1;
+        }
+
+        $this->accountManagement->demoteFrom(
```
BackEndTea: What about adding a check to see if the user is actually an admin before trying to demote the account. And if the account is not an admin, telling the user so, what do you think?

localheinz: @BackEndTea 

I have been thinking about that ever since I started decoupling these commands, however, since this affects all of them, I would like to do it all in one go. What do you think?

BackEndTea: In that case it would make more sense to do that all in one go.

localheinz:
Thank you, @BackEndTea!

PR #791, opened 2017-11-30
Enhancement: Extract PathInterface

localheinz:
This PR

* [x] extracts a `PathInterface` from `Path`

💁‍♂️ Need this for decoupling the `ClearCacheCommand` from the application container.

### DIFF
--- a/tests/Unit/PathTest.php
+++ b/tests/Unit/PathTest.php
@@ -15,12 +15,27 @@ namespace OpenCFP\Test\Unit;
 
 use OpenCFP\Environment;
 use OpenCFP\Path;
+use OpenCFP\PathInterface;
 
 /**
  * @covers \OpenCFP\Path
  */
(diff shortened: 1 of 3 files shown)


tests/Unit/PathTest.php
```diff
@@ -15,12 +15,27 @@
...
+        $reflection = new \ReflectionClass(Path::class);
+
+        $this->assertTrue($reflection->isFinal());
+    }
+
+    public function testImplementsPathInterface()
+    {
+        $reflection = new \ReflectionClass(Path::class);
+
+        $this->assertTrue($reflection->implementsInterface(PathInterface::class));
```
BackEndTea: It looks like we either test this with ``` $reflection->implementsInterface()``` or with ```assertInstanceOf()```, may be best to standardise it to either one.

localheinz:
Thank you, @BackEndTea!

PR #792, opened 2017-11-30
Enhancement: Decouple ClearCacheCommand from application container

localheinz:
This PR

* [x] requires `mikey179/vfsStream`
* [x] decouples the `ClearCacheCommand` from the application container

Follows #717.

### DIFF
--- a/classes/Console/Command/ClearCacheCommand.php
+++ b/classes/Console/Command/ClearCacheCommand.php
@@ -13,18 +13,29 @@ declare(strict_types=1);
 
 namespace OpenCFP\Console\Command;
 
-use OpenCFP\Console\BaseCommand;
+use OpenCFP\PathInterface;
+use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
-class ClearCacheCommand extends BaseCommand
+final class ClearCacheCommand extends Command
 {
+    /**
+     * @var PathInterface
+     */
+    private $path;
+
+    public function __construct(PathInterface $path)
+    {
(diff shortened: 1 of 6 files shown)


localheinz:
@chartjes 

Still needs more work, give me a moment!

classes/Console/Command/ClearCacheCommand.php
```diff
@@ -39,19 +50,50 @@ public function execute(InputInterface $input, OutputInterface $output)
...
 
         $paths = [
-            $this->app['path']->cachePurifierPath(),
-            $this->app['path']->cacheTwigPath(),
+            $this->path->cachePurifierPath(),
+            $this->path->cacheTwigPath(),
         ];
 
         \array_walk($paths, function ($path) use ($io) {
-            \passthru(\sprintf('rm -rf %s/*', $path));
```
localheinz: I fiddled around too long trying to get this to work with `mikey179/vfsStream`, but without success . . .

classes/Console/Command/ClearCacheCommand.php
```diff
@@ -39,19 +50,50 @@ public function execute(InputInterface $input, OutputInterface $output)
...
+                '  * %s (%d files and directories)',
+                $path,
+                $count
             ));
         });
 
         $io->success('Cleared caches.');
     }
+
+    private function deleteFilesAndDirectoriesIn(string $path): int
```
localheinz: . . . that's why I am deleting the content of a directory like this now.

localheinz:
@chartjes 

Sounds good, but unsure whether this makes sense for caching. I suggest that when we switch to using Symfony 4, that we use proper caching adapters, and then in the context of the `ClearCacheCommand`, we shouldn't need to access the file system directly, but rather have the cache adapters clear the caches (whatever the backends are).

localheinz:
Thank you, @chartjes!

PR #794, opened 2017-12-01
Fix: Display the airport information the speaker entered

BackEndTea:
This PR

* [x] Changes the airport display to display what the speaker entered, even if it doesn't match an airport code
For example: 
image


BackEndTea:
Thanks @mdwheele

PR #795, opened 2017-12-01
Enhancement: Update phpunit/phpunit

localheinz:
This PR

* [x] updates `phpunit/phpunit`

💁‍♂️ For reference, see https://github.com/sebastianbergmann/phpunit/compare/6.4.4...6.5.1.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,8 +46,7 @@
         "fzaninotto/faker": "^1.5.0",
         "infection/infection": "^0.6.0",
         "mockery/mockery": "1.0.*@dev",
-        "phpunit/phpunit": "^6.4.4",
-        "phpunit/phpunit-mock-objects": "^4.0"
+        "phpunit/phpunit": "^6.5.1"
     },
     "autoload": {
         "psr-4": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -46,8 +46,7 @@
         "fzaninotto/faker": "^1.5.0",
         "infection/infection": "^0.6.0",
         "mockery/mockery": "1.0.*@dev",
-        "phpunit/phpunit": "^6.4.4",
-        "phpunit/phpunit-mock-objects": "^4.0"
```
localheinz: Removing this requirement as `phpunit/phpunit:^6.5.0` depends on `phpunit/phpunit-mock-objects:^5.0.0`.

BackEndTea: Don't we use the ```phpunit/phpunit-mock-objects``` directly in our tests?
Because if so i'd rather have it explicitly in our own dependencies.

localheinz: It's a direct dependency of `phpunit/phpunit`, as can be seen in [`composer.json`](https://github.com/sebastianbergmann/phpunit/blob/6.5.1/composer.json#L38), so that should be fine.

BackEndTea: Since this is more of a personal taste thing, what do you think @chartjes @mdwheele ?

mdwheele: I'm fine with it. I'm pretty sure if `phpunit/phpunit-mock-objects` is every removed from `phpunit/phpunit`'s dependencies, it'll happen in a major. It's been a composer dependency since 4.0 (and likely before, but Composer didn't exist).

localheinz:
Thank you, @mdwheele!

PR #796, opened 2017-12-01
Enhancement: improve airport model

BackEndTea:
This PR

* [x] Specified the exception being thrown by the airport model
* [x] Improves the check to see if we got an airport
* [x] Updates tests

Related to #794


classes/Domain/Model/Airport.php
```diff
@@ -23,18 +24,18 @@ class Airport extends Eloquent implements AirportInformationDatabase
...
      *
      * @return self
      */
     public function withCode($code): self
     {
         $airport = $this->where('code', $code)->first(['code', 'name', 'country']);
 
-        if (!$airport) {
-            throw new \Exception("An airport matching '{$code}' was not found.");
+        if ($airport instanceof self) {
```
localheinz: What do you think about reverting this change to make it more obvious what the happy path is?

The way it looks right now it suggests that it is an exception if we have actually found an airport, rather than the other way around.

BackEndTea:
@localheinz Should be fixed now

localheinz:
@BackEndTea 

Needs a quick rebase on top of `master`, then this is good to go!

BackEndTea:
@localheinz Should be all good now.

BackEndTea:
Thank you @localheinz

PR #797, opened 2017-12-01
Fix: Inline method

localheinz:
This PR

* [x] inlines a method

Somewhat related to #796.


localheinz:
Thank you, @BackEndTea!

PR #798, opened 2017-12-01
Enhancement: Enable php_unit_strict fixer

BackEndTea:
This PR

* [x] Enables php_unit_strict fixer
* [x] Runs ```make cs```
* [x] Fixes tests

Follows #766 

:tipping_hand_man: For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.8.3#usage:
>* **php_unit_strict**
>
>  PHPUnit methods like ``assertSame`` should be used instead of
  ``assertEquals``.
>
>  *Risky rule: risky when any of the functions are overridden.*
>
>  Configuration options:
>
>  - ``assertions`` (``array``): list of assertion methods to fix; defaults to
    ``['assertAttributeEquals', 'assertAttributeNotEquals', 'assertEquals',
    'assertNotEquals']``

### DIFF
--- a/tests/Unit/EnvironmentTest.php
+++ b/tests/Unit/EnvironmentTest.php
@@ -35,7 +35,7 @@ class EnvironmentTest extends \PHPUnit\Framework\TestCase
         $this->assertTrue($environment->isProduction());
         $this->assertFalse($environment->isDevelopment());
         $this->assertFalse($environment->isTesting());
-        $this->assertEquals(Environment::TYPE_PRODUCTION, $environment);
+        $this->assertSame(Environment::TYPE_PRODUCTION, $environment);
     }
 
     public function testDevelopmentReturnsEnvironment()
@@ -46,7 +46,7 @@ class EnvironmentTest extends \PHPUnit\Framework\TestCase
         $this->assertFalse($environment->isProduction());
         $this->assertTrue($environment->isDevelopment());
         $this->assertFalse($environment->isTesting());
-        $this->assertEquals(Environment::TYPE_DEVELOPMENT, $environment);
+        $this->assertSame(Environment::TYPE_DEVELOPMENT, $environment);
     }
 
     public function testTestingReturnsEnvironment()
@@ -57,7 +57,7 @@ class EnvironmentTest extends \PHPUnit\Framework\TestCase
         $this->assertFalse($environment->isProduction());
         $this->assertFalse($environment->isDevelopment());
         $this->assertTrue($environment->isTesting());
-        $this->assertEquals(Environment::TYPE_TESTING, $environment);
+        $this->assertSame(Environment::TYPE_TESTING, $environment);
     }
 
     /**
@@ -73,7 +73,7 @@ class EnvironmentTest extends \PHPUnit\Framework\TestCase
         $environment = Environment::fromEnvironmentVariable();
 
         $this->assertInstanceOf(Environment::class, $environment);
-        $this->assertEquals($type, $environment);
+        $this->assertSame($type, $environment);
     }
 
     /**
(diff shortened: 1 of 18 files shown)


tests/Unit/Http/Form/SignupFormTest.php
```diff
@@ -401,7 +401,7 @@ public function dataGetsSanitizedCorrectly($inputData, $expectedData)
         $form = new \OpenCFP\Http\Form\SignupForm($inputData, $this->purifier);
         $form->sanitize();
         $sanitizedData = $form->getCleanData();
-        $this->assertEquals(
```
localheinz: While I am personally 👍 for being as strict as possible, the only limitations I see are

* when we really don't care that much about being that strict
* the output of `assertSame()` is not as helpful as that of `asserrEqual()` when comparing arrays

But if that's not an issue, totally 👍!

tests/Unit/Infrastructure/Auth/SentinelAccountManagementTest.php
```diff
@@ -85,7 +85,7 @@ public function testFindByRoleReturnsArray()
         $sentinel = Mockery::mock(Sentinel::class);
         $sentinel->shouldReceive('getRoleRepository->findByName->getUsers->toArray')->andReturn([]);
         $accounts = new SentinelAccountManagement($sentinel);
-        $this->assertEquals([], $accounts->findByRole('blabla'));
+        $this->assertSame([], $accounts->findByRole('blabla'));
```
localheinz: Since [`AccountManagement::findByRole()`](https://github.com/opencfp/opencfp/blob/83c71ba09618c0fe5bab46facb5737cb57b073c7/classes/Domain/Services/AccountManagement.php#L24) has `array` as return type declaration, we can probably just as well use `assertEmpty()` here, what do you think?

tests/Unit/Http/Form/SignupFormTest.php
BackEndTea: If its a big issue we could also do something with [PHPUnitGoodPractices](https://github.com/PHPUnitGoodPractices/Traits) instead, which i think just reports that we are using Equal instead of Same. 

So we don't 'enforce' it strictly, but we are still notified if we don't.

tests/Unit/EnvironmentTest.php
```diff
@@ -102,7 +102,7 @@ public function it_should_be_resolvable_from_string(string $type)
         $environment = Environment::fromString($type);
 
         $this->assertInstanceOf(Environment::class, $environment);
-        $this->assertEquals($type, $environment);
+        $this->assertSame($type, (string) $environment);
```
localheinz: What do you think about directly invoking `__toString()` here (and below) instead?

BackEndTea: I wouldn't mind doing that, but it just feels kinda dirty directly calling a magic method.

tests/Unit/Infrastructure/Auth/SentinelAccountManagementTest.php
BackEndTea: Sound good to me, but there are probably more cases like that, so i prefer to do all those in one (later) PR

tests/Unit/EnvironmentTest.php
mdwheele: As a matter of personal taste, I don't like calling magic-methods either. I think the cast is fine, but would also be okay adding a `toString()` method to `Environment`. Unsure what the benefit of that would be though.

tests/Unit/Http/Form/SignupFormTest.php
mdwheele: I'd be fine rolling forward and if it becomes a problem (array-diffs), we can always adjust. Tests should make our lives easier :)

BackEndTea:
@mdwheele 

If we decided we don't like the strictness, we can remove the php-cs rule again. But i think that overall the strictness gives us more security in the fact that our code works.

mdwheele:
> If we decided we don't like the strictness, we can remove the php-cs rule again. But i think that overall the strictness gives us more security in the fact that our code works.

Egggggg-sacly. :+1: 

I'll let this chill until after lunch and plan on merging.

PR #799, opened 2017-12-01
Enhancement: Extract dedicated constructor

BackEndTea:
This PR

* [x] extracts a dedicated named constructor for ```AuthenticationException```
* [x] changes it from a generic exception to a runtime exception 
* [x] makes the class final

### DIFF
--- a/tests/Unit/Domain/Services/AuthenticationExceptionTest.php
+++ b/tests/Unit/Domain/Services/AuthenticationExceptionTest.php
@@ -21,10 +21,31 @@ use PHPUnit\Framework;
  */
 final class AuthenticationExceptionTest extends Framework\TestCase
 {
+    public function testIsFinal()
+    {
+        $reflection = new \ReflectionClass(AuthenticationException::class);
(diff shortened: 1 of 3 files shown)


tests/Unit/Domain/Services/AuthenticationExceptionTest.php
```diff
@@ -21,10 +21,31 @@
...
+    public function testIsRuntimeException()
+    {
+        $exception = new AuthenticationException();
+
+        $this->assertInstanceOf(\RuntimeException::class, $exception);
+    }
+
+    public function testLoginFailureHasCorrectMessage()
+    {
+        $exception = AuthenticationException::loginFailure();
```
localheinz: How about asserting that `AuthenticationException:: loginFailure()` returns an `AuthenticationException` as well?

BackEndTea: No problem

localheinz:
Thank you, @BackEndTea!

PR #800, opened 2017-12-01
Fix: Remove TalkSubmission and related classes

BackEndTea:
This PR

* [x] Removes the TalkSubmission class, and related classes/ methods
* [x] Removes the tests related to TalkSubmission
* [x] Removes (now) useless dependiencies of the Speakers class

I was trying to figure out where they were being used, but couldn't find any place the functions were used outside of the tests written for it


mdwheele:
You're good. This was effectively dead code. I want to check this out a little more in depth to double-check (since the only coverage was tests that are being removed), but honestly think it's good to go.

BackEndTea:
@mdwheele No problem, i'd rather this take a longer time to merge than having to undo it since i removed some code that actually was used.

PR #801, opened 2017-12-01
Fix: Remove unused dependency

localheinz:
This PR

* [x] removes an unused dependency

Spotted in #800.


localheinz:
Thank you, @BackEndTea!

PR #802, opened 2017-12-01
Fix: Remove empty class

localheinz:
This PR

* [x] removes an empty class


localheinz:
Thank you, @mdwheele!

PR #803, opened 2017-12-01
Enhancement: Include bin/opencfp in php-cs-fixer configuration

localheinz:
This PR

* [x] includes files named `opencfp` in `php-cs-fixer` configuration
* [x] runs `make cs`

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -11,7 +11,9 @@ the LICENSE file that was distributed with this source code.
 @see https://github.com/opencfp/opencfp
 TXT;
 
-$finder = PhpCsFixer\Finder::create()->in(__DIR__);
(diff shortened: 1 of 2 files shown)


BackEndTea:
I don't exactly know correct usage of PHP-CS, but i assume this now checks all ```.php``` files, and files named ```opencfp```?

localheinz:
@BackEndTea 

As you can see in [`PhpCsFixer\Finder::__construct()`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.8.3/src/Finder.php#L23-L35), there is already some default configuration. Now we are also including all files named `opencfp`, as you guessed. 👍

localheinz:
Thank you, @BackEndTea!

PR #806, opened 2017-12-01
Remove: FlashableTrait

BackEndTea:
This PR

* [x] Removes ```FlashableTrait``` and code related to it 

This trait was used in two places in the code, besides that all uses of flash were done directly through the session service.


localheinz:
Thank you, @BackEndTea!

PR #807, opened 2017-12-01
Remove application object from function call

BackEndTea:
This PR

* [x] Removes the application object from 2 controller function calls

There was no real reason why it had to be passed to the function directly


BackEndTea:
Thanks @mdwheele

PR #808, opened 2017-12-01
Fix: UserAccess to use Auth object

BackEndTea:
This PR

* [x] lets the UserAccess classes use the Auth object, rather than the Application object.
 
There really is no reason to pass it the entire application object, when all it needs is the Authentication object.

A future PR idea would be to rename this to something like ``RoleValidator``` and have it implement a validator interface. That way the naming would make more sense

### DIFF
--- a/tests/Unit/Infrastructure/Auth/RoleAccessTest.php
+++ b/tests/Unit/Infrastructure/Auth/RoleAccessTest.php
@@ -14,76 +14,56 @@ declare(strict_types=1);
 namespace OpenCFP\Test\Unit\Infrastructure\Auth;
 
 use Mockery;
+use Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration;
 use OpenCFP\Domain\Services\Authentication;
 use OpenCFP\Infrastructure\Auth\RoleAccess;
 use OpenCFP\Infrastructure\Auth\UserInterface;
+use OpenCFP\Test\Helper\Faker\GeneratorTrait;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 
 /**
  * @covers \OpenCFP\Infrastructure\Auth\RoleAccess
  */
-class RoleAccessTest extends \PHPUnit\Framework\TestCase
+final class RoleAccessTest extends \PHPUnit\Framework\TestCase
 {
-    use Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration;
+    use GeneratorTrait;
+    use MockeryPHPUnitIntegration;
 
-    public function testReturnsRedirectIfCheckFailed()
+    public function testReturnsRedirectResponseIfCheckFailed()
     {
+        $role = $this->getFaker()->word;
+
         $auth = Mockery::mock(Authentication::class);
         $auth->shouldReceive('check')->andReturn(false);
 
-        $this->assertInstanceOf(RedirectResponse::class, RoleAccess::userHasAccess($auth, 'admin'));
+        $this->assertInstanceOf(RedirectResponse::class, RoleAccess::userHasAccess($auth, $role));
     }
 
-    public function testReturnsFalseIfCheckSucceededButUserHasNoAdminPermission()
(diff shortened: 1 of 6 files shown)


tests/Unit/Infrastructure/Auth/RoleAccessTest.php
```diff
@@ -55,23 +54,36 @@ public function testReturnsNothingIfCheckSucceededAndUserHasAdminPermission()
...
+        $user = Mockery::mock(UserInterface::class);
+        $user->shouldReceive('hasAccess')->with('admin')->andReturn(true);
+        $user->shouldReceive('hasAccess')->with('reviewer')->andReturn(false);
+
+        $auth = Mockery::mock(Authentication::class);
+        $auth->shouldReceive('check')->andReturn(true);
+        $auth->shouldReceive('user')->andReturn($user);
+
+        $this->assertInstanceOf(RedirectResponse::class, RoleAccess::userHasAccess($auth, 'reviewer'));
+        $this->assertNull(RoleAccess::userHasAccess($auth, 'admin'));
```
localheinz: Since we act twice in this test, can we split it into two tests?

tests/Unit/Infrastructure/Auth/RoleAccessTest.php
```diff
@@ -55,23 +54,36 @@ public function testReturnsNothingIfCheckSucceededAndUserHasAdminPermission()
...
+        $user = Mockery::mock(UserInterface::class);
+        $user->shouldReceive('hasAccess')->with('reviewer')->andReturn(true);
+        $user->shouldReceive('hasAccess')->with('admin')->andReturn(false);
+
+        $auth = Mockery::mock(Authentication::class);
+        $auth->shouldReceive('check')->andReturn(true);
+        $auth->shouldReceive('user')->andReturn($user);
+
+        $this->assertInstanceOf(RedirectResponse::class, RoleAccess::userHasAccess($auth, 'admin'));
+        $this->assertNull(RoleAccess::userHasAccess($auth, 'reviewer'));
```
localheinz: Since we act twice in this test, can we split it into two tests?

tests/Unit/Infrastructure/Auth/RoleAccessTest.php
```diff
@@ -55,23 +55,60 @@ public function testReturnsNothingIfCheckSucceededAndUserHasAdminPermission()
...
+        $auth->shouldReceive('user')->andReturn($user);
+
+        $this->assertInstanceOf(RedirectResponse::class, RoleAccess::userHasAccess($auth, 'reviewer'));
+    }
+
+    public function testAdminCanGetToAdminPage()
+    {
+        $user = Mockery::mock(UserInterface::class);
+        $user->shouldReceive('hasAccess')->with('admin')->andReturn(true);
+        $user->shouldReceive('hasAccess')->with('reviewer')->andReturn(false);
```
localheinz: Can we remove this expectation? The method `User::hasAccess()` is only invoked with `'admin'` in the context of this test.

tests/Unit/Infrastructure/Auth/RoleAccessTest.php
```diff
@@ -55,23 +55,60 @@ public function testReturnsNothingIfCheckSucceededAndUserHasAdminPermission()
...
         $auth->shouldReceive('check')->andReturn(true);
         $auth->shouldReceive('user')->andReturn($user);
-        $this->swap(Authentication::class, $auth);
 
         //The middleware doesn't do anything if the user is an admin, so it returns null (void)
-        $this->assertNull(RoleAccess::userHasAccess($this->app, 'admin'));
+        $this->assertNull(RoleAccess::userHasAccess($auth, 'admin'));
     }
 
     public function testReviewerCantGetToAdminPages()
```
localheinz: The entire test just duplicates `testReturnsFalseIfCheckSucceededButUserHasNoAdminPermission()`, can we remove this test here?

PR #809, opened 2017-12-01
Add type declarations to the AccountManagement interface

BackEndTea:
This PR

* [x] Adds type declarations to the Account Management Interface

### DIFF
--- a/classes/Infrastructure/Auth/SentinelAccountManagement.php
+++ b/classes/Infrastructure/Auth/SentinelAccountManagement.php
@@ -36,7 +36,7 @@ final class SentinelAccountManagement implements AccountManagement
      *
      * @return UserInterface
      */
-    public function findById($userId): UserInterface
+    public function findById(int $userId): UserInterface
     {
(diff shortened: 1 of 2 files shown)


classes/Infrastructure/Auth/SentinelAccountManagement.php
```diff
@@ -95,15 +95,15 @@ public function create($email, $password, array $data = []): UserInterface
      *
      * @return bool
      */
-    public function activate($email): bool
+    public function activate(string $email): bool
```
localheinz: Shall we stop returning something from this method and remove the return type declaration?

Alternatively, we could pull up the return type declaration into the interface. Personally, though, I would like to avoid returning a `bool` from a method to indicate it was invoked successfully.

localheinz: The reason I am asking that I think we should rely on the interface, not the implementation.

localheinz:
Thank you, @BackEndTea!

PR #813, opened 2017-12-02
Fix: Use type declaration

localheinz:
This PR

* [x] uses a type declaration for the `$role` argument

Spotted in #808.


localheinz:
Thank you, @BackEndTea!

PR #814, opened 2017-12-02
Fix: Add missing docblocks

localheinz:
This PR

* [x] adds missing docblocks


classes/Infrastructure/Auth/SentinelAccountManagement.php
```diff
@@ -29,13 +29,6 @@ public function __construct(Sentinel $sentinel)
         $this->sentinel = $sentinel;
     }
 
-    /**
-     * @param int $userId
-     *
-     * @throws UserNotFoundException
-     *
-     * @return UserInterface
-     */
```
localheinz: Removing the docblock here and below because they are automatically inherited.

localheinz:
Thank you, @chartjes!

PR #815, opened 2017-12-02
Fix: Reduce visibility

localheinz:
This PR

* [x] reduces visibility of `setUp()` from `public` to `protected`

💁‍♂️ This is the visibility as declared on the parent; there's no need to increase it.


localheinz:
Thank you, @BackEndTea!

PR #816, opened 2017-12-02
Fix: Do not return anything from activate()

localheinz:
This PR

* [x] stops returning a `bool` from `SentinelAccountManagement::activate()`

Follows https://github.com/opencfp/opencfp/pull/809#discussion_r154369159.

### DIFF
--- a/classes/Infrastructure/Auth/SentinelAccountManagement.php
+++ b/classes/Infrastructure/Auth/SentinelAccountManagement.php
@@ -88,19 +88,12 @@ final class SentinelAccountManagement implements AccountManagement
         throw new UserExistsException();
     }
 
-    /**
-     * @param string $email
-     *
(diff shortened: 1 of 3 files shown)


classes/Infrastructure/Auth/SentinelAccountManagement.php
```diff
@@ -88,19 +88,12 @@ public function create(string $email, string $password, array $data = []): UserI
         throw new UserExistsException();
     }
 
-    /**
-     * @param string $email
-     *
-     * @throws UserNotFoundException
-     *
-     * @return bool
-     */
```
localheinz: Removing the docblock here as it is added to the `AccountManagement` interface in #814.

localheinz:
Thank you, @BackEndTea!

PR #817, opened 2017-12-02
Fix: Extract variable for role name

localheinz:
This PR

* [x] extracts a variable for the role name


localheinz:
@chartjes 

It’s the plan to merge these commands into two commands

- one for promoting
- one for demoting

👍

localheinz:
Thank you, @chartjes!

PR #818, opened 2017-12-02
Enhancement: Throw RoleNotFoundException if role was not found

localheinz:
This PR

* [x] implements a `RoleNotFoundException`
* [x] throws a `RoleNotFoundException` if a role was not found


localheinz:

Quick heads up, as the plan is to merge the commands promoting users from and to *admin* and *reviewer* roles, and as the cases where the roles haven't been found are currently unhandled in these commands, I suggest handling these exceptions once the commands have been merged.

localheinz:
Thank you, @BackEndTea!

PR #819, opened 2017-12-02
Enhancement: Merge commands for promoting and demoting users

localheinz:
This PR

* [x] merges commands for promoting and demoting users


localheinz:
@BackEndTea @mdwheele 

Feel free to take a look as well!

localheinz:
Note that the coverage is dropping because we are removing code that is covered 100%, while the rest of the code base isn't.

localheinz:
Thank you, @chartjes!

PR #822, opened 2017-12-02
Fix: Wrapping

localheinz:
This PR

* [x] fixes argument wrapping


localheinz:
Thank you, @chartjes!

PR #823, opened 2017-12-02
Fix: Remove unused class

localheinz:
This PR

* [x] removes an unused class

Follows #790.


localheinz:
Thank you, @BackEndTea!

PR #824, opened 2017-12-02
Fix: Simplify promotion and demotion

localheinz:
This PR

* [x] simplifies the promotion and demotion of users

Follows #819.

💁‍♂️ Next step, after merging #818: handling of `RoleNotFoundException`.

### DIFF
--- a/classes/Console/Command/UserDemoteCommand.php
+++ b/classes/Console/Command/UserDemoteCommand.php
@@ -72,7 +72,10 @@ EOF
         ));
 
         try {
-            $this->accountManagement->findByLogin($email);
+            $this->accountManagement->demoteFrom(
+                $email,
+                $roleName
+            );
         } catch (Auth\UserNotFoundException $exception) {
             $io->error(\sprintf(
                 'Could not find account with email "%s".',
@@ -81,11 +84,6 @@ EOF
 
             return 1;
         }
-
-        $this->accountManagement->demoteFrom(
-            $email,
-            $roleName
-        );
 
         $io->success(\sprintf(
             'Removed account with email "%s" from the "%s" group.',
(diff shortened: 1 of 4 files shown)


localheinz:
@chartjes 

> There are two possible exceptions that could be thrown: one that the email doesn't exist ...

Both [`AccountManagement::promoteTo()`](https://github.com/opencfp/opencfp/blob/85879c6f6bd63a7e15b0d8ca8ac1d0d8c402e554/classes/Domain/Services/AccountManagement.php#L65-L71) and [`AccountManagement::demoteFrom()`](https://github.com/opencfp/opencfp/blob/85879c6f6bd63a7e15b0d8ca8ac1d0d8c402e554/classes/Domain/Services/AccountManagement.php#L73-L79) will pass through a [`UserNotFoundException`](https://github.com/opencfp/opencfp/blob/85879c6f6bd63a7e15b0d8ca8ac1d0d8c402e554/classes/Infrastructure/Auth/UserNotFoundException.php) if a user was not found, so this case is covered. 

What we are currently doing is this

1. try to find the user by login, if not found stop processing with an error
2. then try to promote / demote the user, where we again try to find the user by login again

This PR removes **1.**, as there's no reason for the commands to do again what is already handled by `AccountManagement`.

> and one that for whatever reason we were unable to update the database record for this user

This case isn't currently handled at all, but we can of course handle it in a separate PR.

What do you think?

chartjes:
I realize we aren't currently testing for it but we should think about if this is a failure mode we need to worry about. Error messages that are not about the actual error can be confusing.

localheinz:
I am happy to catch a generic exception and display a corresponding error message in a separate pull request, @chartjes!

localheinz:
Thank you, @chartjes!

PR #825, opened 2017-12-02
Fix: Increase visibility of services shared by container

localheinz:
This PR

* [x] increases the visibility of services shared by the application container

💁‍♂️ Introducing temporary variables for services fetched from the application container allows to annotate them with the type to which they resolve, and thus increases visibility - making it easier to see which services are actually used, as well as navigating the code.

### DIFF
--- a/classes/Http/Controller/Admin/TalksController.php
+++ b/classes/Http/Controller/Admin/TalksController.php
@@ -20,6 +20,7 @@ use OpenCFP\Domain\Talk\TalkHandler;
 use OpenCFP\Domain\ValidationException;
 use OpenCFP\Http\Controller\BaseController;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Session;
 
 class TalksController extends BaseController
 {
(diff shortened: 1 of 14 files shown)


classes/Http/Controller/Admin/TalksController.php
```diff
@@ -128,15 +144,21 @@ public function commentCreateAction(Request $req)
...
-        $this->service(TalkHandler::class)
+        /** @var TalkHandler $talkHandler */
+        $talkHandler = $this->service(TalkHandler::class);
+
+        $talkHandler
             ->grabTalk($talkId)
             ->commentOn($req->get('comment'));
 
-        $this->service('session')->set('flash', [
-                'type'  => 'success',
```
localheinz: While at it, fixing the indentation here, too.

BackEndTea:
👍 For me if the conflicts get resolved

localheinz:
@BackEndTea 

Rebased!

localheinz:
Thank you, @BackEndTea!

PR #826, opened 2017-12-02
Fix: Use tag when requiring mockery/mockery

localheinz:
This PR

* [x] uses a tag when requiring `mockery/mockery`
* [x] fixes a failing test


localheinz:
Thank you, @chartjes!

PR #827, opened 2017-12-02
Fix: Instructions for promoting and demoting users

localheinz:
This PR

* [x] updates instructions for promoting and demoting users

Follows #819.


localheinz:
Thank you, @BackEndTea!

PR #828, opened 2017-12-03
Fix: Use type declaration

localheinz:
This PR

* [x] uses a type declaration


classes/Domain/Speaker/SpeakerRepository.php
```diff
@@ -21,13 +21,13 @@
     /**
      * Retrieves a speaker with associated talks.
      *
-     * @param string $speakerId
+     * @param int $speakerId
```
localheinz: We are using `int` everywhere as identifiers.

localheinz:
Thank you, @mdwheele!

PR #830, opened 2017-12-03
Fix: Rename $req to $request

localheinz:
This PR

* [x] renames parameters and variables from `$req` to `$request`

💁‍♂️ There's no benefit in abbreviating.


localheinz:
Thank you, @mdwheele!

PR #831, opened 2017-12-03
Fix: Set background-size to cover

localheinz:
This PR

* [x] sets the `background-size` for hero images to `cover`

💁‍♂️ This mostly affects larger displays, but I think it wasn't intended to look as it does.

### Before

screen shot 2017-12-03 at 15 23 21
screen shot 2017-12-03 at 15 23 26
screen shot 2017-12-03 at 15 23 31

### After

screen shot 2017-12-03 at 15 23 44
screen shot 2017-12-03 at 15 23 47
screen shot 2017-12-03 at 15 23 51

### DIFF
--- a/resources/views/home.twig
+++ b/resources/views/home.twig
@@ -1,7 +1,7 @@
 {% extends "layouts/default.twig" %}
 
 {% block header %}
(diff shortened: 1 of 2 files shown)


mdwheele:
Thanks! Looks great.

localheinz:
Thank you, @mdwheele!

PR #832, opened 2017-12-03
Fix: Limit airport to 3 characters

localheinz:
This PR

* [x] limits the length of the `airport` field to `3` characters

💁‍♂️ We should probably also update validation, but I couldn't find any.

### DIFF
--- a/migrations/20171203144758_limit_airport_to_three_characters.php
+++ b/migrations/20171203144758_limit_airport_to_three_characters.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Copyright (c) 2013-2017 OpenCFP
+ *
+ * For the full copyright and license information, please view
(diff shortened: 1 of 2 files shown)


BackEndTea:
Id like to either enforce the fact that the key entered has to be in the airport table, or leave it as is.

Personally i'd like to keep it as is, since it allows speakers who don't know about airport codes and all that to still enter something useful after #794

localheinz:
@chartjes @mdwheele 

Do you have an opinion?

chartjes:
If we're not going to implement some kind of search where people type the name of an airport and we provide the code, we should leave it as it is

localheinz:
@chartjes 

Closing this for now, then, we can still implement the suggested functionality at some other point in time.

mdwheele:
I'll write the autocomplete if I can use VueJS :grinning:

PR #833, opened 2017-12-03
Fix: Rename method

localheinz:
This PR

* [x] renames `Authentication::check()` to `Authentication::isAuthenticated()`

💁‍♂️ I believe something like `isAuthenticated()` or `isSignedIn()` or `isLoggedIn()` reveals the intent better than something as generic as `check()`. It also aligns nicely with the [`Authentication::authenticate()`](https://github.com/opencfp/opencfp/blob/d872d646e9b561e9cb8ed41fc3e1a4abdbaddfeb/classes/Domain/Services/Authentication.php#L20-L28).

### DIFF
--- a/classes/Domain/Services/Authentication.php
+++ b/classes/Domain/Services/Authentication.php
@@ -50,7 +50,7 @@ interface Authentication
      *
      * @return bool
      */
-    public function check(): bool;
+    public function isAuthenticated(): bool;
 
     /**
      * Determine whether the user is a non-authenticated guest.
--- a/classes/Http/Controller/SignupController.php
+++ b/classes/Http/Controller/SignupController.php
@@ -25,7 +25,7 @@ class SignupController extends BaseController
     {
         $auth = $this->service(Authentication::class);
 
-        if ($auth->check()) {
+        if ($auth->isAuthenticated()) {
             return $this->redirectTo('dashboard');
         }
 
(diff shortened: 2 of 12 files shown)


classes/Domain/Services/Authentication.php
```diff
@@ -50,7 +50,7 @@ public function userId(): int;
      *
      * @return bool
      */
-    public function check(): bool;
+    public function isAuthenticated(): bool;
```
localheinz: Also see #835.

localheinz: @BackEndTea @chartjes @mdwheele 

Since we have

* [`authenticate()`](https://github.com/opencfp/opencfp/blob/d872d646e9b561e9cb8ed41fc3e1a4abdbaddfeb/classes/Domain/Services/Authentication.php#L20-L28)
* [`check()`](https://github.com/opencfp/opencfp/blob/d872d646e9b561e9cb8ed41fc3e1a4abdbaddfeb/classes/Domain/Services/Authentication.php#L48-L53)
* [`guest()`](https://github.com/opencfp/opencfp/blob/d872d646e9b561e9cb8ed41fc3e1a4abdbaddfeb/classes/Domain/Services/Authentication.php#L55-L60), also see #835 
* [`logout()`](https://github.com/opencfp/opencfp/blob/d872d646e9b561e9cb8ed41fc3e1a4abdbaddfeb/classes/Domain/Services/Authentication.php#L62-L65)

what do you think about aligning them altogether?

### Example 1

* `authenticate()`
* `isAuthenticated()`
* `clear()`

### Example 2

* `signIn()`
* `isSignedIn()`
* `signOut()`

### Example 3

* `logIn()`
* `isLoggedIn()`
* `logOut()`

?

localheinz:
Thank you, @chartjes!

PR #835, opened 2017-12-03
Fix: Remove unused method

localheinz:
This PR

* [x] removes an unused method

Related to #833.

### DIFF
--- a/classes/Domain/Services/Authentication.php
+++ b/classes/Domain/Services/Authentication.php
@@ -53,13 +53,6 @@ interface Authentication
     public function isAuthenticated(): bool;
 
     /**
-     * Determine whether the user is a non-authenticated guest.
-     *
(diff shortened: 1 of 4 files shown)


classes/Domain/Services/Authentication.php
```diff
@@ -52,13 +52,6 @@ public function userId(): int;
      */
     public function check(): bool;
 
-    /**
-     * Determine whether the user is a non-authenticated guest.
-     *
-     * @return bool
-     */
-    public function guest(): bool;
```
localheinz: Other than that the method isn't currently used, we probably don't need two methods for a similar thing.

localheinz:
@chartjes 

Rebased!

localheinz:
Thank you, @chartjes!

PR #836, opened 2017-12-03
Fix: Mark getFaker() as final

localheinz:
This PR

* [x] marks the method `getFaker()` as `final`

💁‍♂️ This prevents needlessly importing the `GeneratorTrait` multiple times along an inheritance chain.


localheinz:
Thank you, @chartjes!

PR #837, opened 2017-12-03
Fix: Pull up import of GeneratorTrait

localheinz:
This PR

* [x] pulls up the import of the `GeneratorTrait`

Follows #836.


localheinz:
Thank you, @BackEndTea!

PR #839, opened 2017-12-04
Enhancement: Catch RoleNotFoundException

localheinz:
This PR

* [x] catches `RoleNotFoundExceptions` thrown during `AccountManagemenht::demoteFrom()` and `AccountManagement::promoteTo()`

Follows #818.

PR #838, opened 2017-12-04
Enhancement: Catch generic exception

localheinz:
This PR

* [x] catches generic exceptions thrown during `AccountManagemenht::demoteFrom()` and `AccountManagement::promoteTo()`

Follows https://github.com/opencfp/opencfp/pull/824#issuecomment-348823711.

PR #840, opened 2017-12-04
Fix: Rename test methods

localheinz:
This PR

* [x] renames test methods

💁‍♂️ This brings them more in line with those added in #838 and #839.


localheinz:
Thank you, @BackEndTea!

PR #841, opened 2017-12-04
Enhancement: Require localheinz/test-util

localheinz:
This PR

* [x] requires `localheinz/test-util`
* [x] uses `Localheinz\Test\Util\Helper` instead of `OpenCFP\Test\Helper\Faker\GeneratorTrait`
* [x] removes `OpenCFP\Test\Helper\Faker\GeneratorTrait`
* [x] uses assertions as provided by `Localheinz\Test\Util\Helper`

💁‍♂️ For reference, see [`localheinz/test-util`](https://github.com/localheinz/test-util).


tests/Unit/Domain/Services/AuthenticationExceptionTest.php
```diff
@@ -21,24 +22,16 @@
...
-        $reflection = new \ReflectionClass(AuthenticationException::class);
-        $this->assertTrue($reflection->isFinal());
+        $this->assertClassIsFinal(AuthenticationException::class);
     }
-
-    public function testIsException()
-    {
-        $exception = new AuthenticationException();
-
-        $this->assertInstanceOf(\Exception::class, $exception);
```
localheinz: Probably don't need to assert here that `RuntimeException` extends `Exception`.

localheinz:

What do you think?

BackEndTea:
👍 For me

localheinz:
Thank you, @chartjes!

PR #842, opened 2017-12-04
Fix: Mark test classes abstract or final

localheinz:
This PR

* [x] asserts that test classes are `abstract` or `final`
* [x] marks test classes as `final`

Follows #841.


localheinz:
Thank you, @chartjes!

PR #843, opened 2017-12-04
Enhancement: Assert that production classes have unit tests

localheinz:
This PR

* [x] asserts that production classes have unit tests
* [x] excludes production classes without unit tests

Follows #842.

### DIFF
--- a/tests/Unit/ProjectCodeTest.php
+++ b/tests/Unit/ProjectCodeTest.php
@@ -14,6 +14,9 @@ declare(strict_types=1);
 namespace OpenCFP\Test\Unit;
 
 use Localheinz\Test\Util\Helper;
+use OpenCFP\Domain;
+use OpenCFP\Http;
+use OpenCFP\Provider;
 use PHPUnit\Framework;
 
 /**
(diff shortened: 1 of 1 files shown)


tests/Unit/ProjectCodeTest.php
```diff
@@ -0,0 +1,87 @@
...
+use OpenCFP\Domain;
+
+use OpenCFP\Http;
+use OpenCFP\Provider;
+use PHPUnit\Framework;
+
+/**
+ * @coversNothing
+ */
+final class ProjectCodeTest extends Framework\TestCase
```
BackEndTea: I think we need a better name for this test class

localheinz: I'm open for suggestions, @BackEndTea!

mdwheele:
You know... I have never thought to do this... but it's pretty awesome. :+1: from me.

localheinz:
Glad you like it, @mdwheele!

BackEndTea:
@chartjes The idea would be that we would never have to add classes to it, since all new code is tested.

localheinz:
@chartjes 

> Does this list have to be manually updated if we add a new class? 

Yes.

> If so, this needs to be able to just scan all those classes and make sure we have test code for it.

Not sure I understand, can you explain?

chartjes:
@localheinz I guess my question is this:

If I create a new class and do not have tests for it, will it be detected by this new code? If so, then I have no objections.

What I am saying is that if we are going to add in something that says "hey, there are no tests for this class" then it has to be automated and no manual adding of classes to a list.

Does that make sense?

localheinz:
Thanks for clarifying, @chartjes!

> If I create a new class and do not have tests for it, will it be detected by this new code? If so, then I have no objections.

Yes, then this test will fail, and we either have to add a unit test, or we have to add the name of the class to the list of excluded classes.

> What I am saying is that if we are going to add in something that says "hey, there are no tests for this class" then it has to be automated and no manual adding of classes to a list.

Ha, I get it - you want to automate it so it's easier for people not to write tests?!

:trollface:

chartjes:
@localheinz LOL - of course I want people to write tests! I just want to make sure if we have automated systems looking out for us that they actually detect when we have new code without tests and properly alert us.

localheinz:
@chartjes 

Hehe, just kidding!

localheinz:
Thank you, @chartjes!

PR #845, opened 2017-12-04
Fix php internal server

BackEndTea:
This PR

* [x] Fixes an issue with the PHP internal server for php version 7.1.12

I'm not 100% sure why, but since upgrading to PHP 7.1.12 i had an issue with the internal php server, for some reason it was looking for the server.php in the web folder, rather then the root of the project. 

So this add a check to see if the user is on php 7.1.12, and if so, it points it to the correct place

### DIFF
--- a/script/server
+++ b/script/server
@@ -9,4 +9,9 @@ cd "$(dirname "$0")/.."
 # ensure everything in the app is up to date.
 script/update
 
-CFP_ENV=development php -S localhost:8000 -t web server.php
+export PHPVER=$(php -r "echo PHP_VERSION;")
+if [ "$PHPVER" = "7.1.12" ]; then
+    CFP_ENV=development php -S localhost:8000 -t web ../server.php
+else
+    CFP_ENV=development php -S localhost:8000 -t web server.php
+fi


chartjes:
This generates a WTF from me. I wonder what it is about PHP 7.1.12 that causes the issue

mdwheele:
This is caused by https://github.com/php/php-src/commit/816758eda2bcdd69ba505fb6bbb79124a7bf2254. The new behaviour is that the script is relative to document root. So you have to `../` because server is outside of the document root. Seems like a BC break but in reality, I think the argument is that because of the #75287 (https://bugs.php.net/bug.php?id=75287), the way this has *been* working was never the intended behaviour. 

I think this is an *okay* general direction for fix, but we can't just lock to 7.1.12, if this is even something we want to support at all. I'd much rather see us [dip into the Dockers](https://github.com/mdwheele/opencfp-docker) and bail on the built-in. But I'm an asshole like that.

mdwheele:
Honestly, I take all that back. This is getting patched into 7.0.26 and was also patched into 7.2.0. I think the best thing to do is close this out and just wait for 7.0.26 to be released. Thoughts?

BackEndTea:
I'm okay with going for Docker support and just letting this functionality go. But if we want to support it, it should work, as long as i'm running php 7+

chartjes:
So we have two options:

* accept the PR so it works for people running PHP 7.1.12
* bump the minimum PHP version to PHP 7.1.13

I am okay with the PR.

BackEndTea:
Thank you, @chartjes

PR #846, opened 2017-12-04
Fix: Use faker() instead of unavailable getFaker()

localheinz:
This PR

* [x] uses `Helper::faker()` instead of the unavailable `GeneratorTrait::getFaker()`

Follows #841.

🤦‍♂️ Sorry!


localheinz:
Thank you, @chartjes!

PR #593, opened 2017-11-13
Enhancement: Enable blank_line_before_statement fixer

localheinz:
❗️ Blocked by #596.

This PR

* [x] enables and configures the `blank_line_before_statement` fixer
* [x] runs `make cs`

Replaces #584.
Follows #587.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.8.0#usage:

>**blank_line_before_statement** [`@Symfony`]
>
>An empty line feed must precede any configured statement.
>
>Configuration options:
>
>* `statements` (`array`): list of statements which must be preceded by an empty line; defaults to `['break', 'continue', 'declare', 'return', 'throw', 'try']`

PR #847, opened 2017-12-05
Enhancement: Configure blank_line_before_statement fixer

localheinz:
This PR

* [x] configures the `blank_line_before_statement` fixer
* [x] runs `make cs`

Follows #593.

💁‍♂️ The idea is to give a bit more breathing room before something important is happening.


localheinz:
Thank you, @BackEndTea!

PR #848, opened 2017-12-05
Fix: Remove userId()

localheinz:
This PR

* [x] removes `Authentication::userId()`

💁‍♂️ We already have [`Authentication::user()`](https://github.com/opencfp/opencfp/blob/8972fcbb1d0f6dfd50d6161906eb174e2b2e2315/classes/Domain/Services/Authentication.php#L30-L37) and the implementation of [`Authentication::userId()`](https://github.com/opencfp/opencfp/blob/8972fcbb1d0f6dfd50d6161906eb174e2b2e2315/classes/Infrastructure/Auth/SentinelAuthentication.php#L67-L70) just hides the fact that we violate the *Law of Demeter*.


localheinz:
Thank you, @chartjes!

PR #849, opened 2017-12-05
Fix: Exclude cached php files from php cs fixer

BackEndTea:
This PR

* [x] excludes the cache dir from php-cs fixer


localheinz:
Thank you, @BackEndTea!

BackEndTea:
Thank you, @localheinz

PR #851, opened 2017-12-05
Fix: Inline data passed to render()

localheinz:
This PR

* [x] inlines data passed to `render()`

### DIFF
--- a/classes/Http/Controller/Admin/TalksController.php
+++ b/classes/Http/Controller/Admin/TalksController.php
@@ -77,7 +77,7 @@ class TalksController extends BaseController
         $pagerfanta->setCurrentPage($request->get('page'));
         $pagination = $pagerfanta->createView('/admin/talks?', $request->query->all());
 
-        $templateData = [
+        return $this->render('admin/talks/index.twig', [
             'pagination'   => $pagination,
             'talks'        => $pagerfanta->getFanta(),
             'page'         => $pagerfanta->getCurrentPage(),
@@ -87,9 +87,7 @@ class TalksController extends BaseController
             'per_page'     => $perPage,
             'sort'         => $request->get('sort'),
             'order_by'     => $request->get('order_by'),
-        ];
-
-        return $this->render('admin/talks/index.twig', $templateData);
+        ]);
     }
 
     public function viewAction(Request $request)
@@ -107,7 +105,9 @@ class TalksController extends BaseController
             return $this->redirectTo('admin_talks');
         }
 
-        return $this->render('admin/talks/view.twig', ['talk' => $handler->getProfile()]);
+        return $this->render('admin/talks/view.twig', [
+            'talk' => $handler->getProfile(),
+        ]);
     }
 
     public function rateAction(Request $request)
--- a/classes/Http/Controller/ForgotController.php
+++ b/classes/Http/Controller/ForgotController.php
@@ -56,12 +56,11 @@ class ForgotController extends BaseController
     public function indexAction()
     {
         $form = $this->formFactory->createBuilder(ForgotForm::class)->getForm();
-        $data = [
+
+        return $this->render('security/forgot_password.twig', [
             'form'         => $form->createView(),
             'current_page' => 'Forgot Password',
-        ];
-
-        return $this->render('security/forgot_password.twig', $data);
+        ]);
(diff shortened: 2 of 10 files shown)


classes/Http/Controller/Admin/TalksController.php
```diff
@@ -86,7 +84,9 @@ public function viewAction(Request $request)
             return $this->app->redirect($this->url('admin_talks'));
         }
 
-        return $this->render('admin/talks/view.twig', ['talk' => $handler->getProfile()]);
```
localheinz: While at it, I figured it would make sense to streamline how we pass data to the templates - this is an example.

classes/Http/Controller/ForgotController.php
```diff
@@ -142,10 +140,10 @@ public function resetAction(Request $request)
 
         $form = $formFactory->create(new ResetForm());
 
-        $data['form']  = $form->createView($formOptions);
```
localheinz: Heads up, `$data` is implicitly created here.

classes/Http/Controller/ForgotController.php
```diff
@@ -168,7 +166,9 @@ public function processResetAction(Request $request)
             $form->get('user_id')->setData($userId);
             $form->get('reset_code')->setData($resetCode);
 
-            return $this->render('user/reset_password.twig', ['form' => $form->createView()]);
```
localheinz: While at it, I figured it would make sense to streamline how we pass data to the templates - this is an example.

classes/Http/Controller/ForgotController.php
```diff
@@ -211,7 +211,9 @@ public function updatePasswordAction(Request $request)
         $form->handleRequest($request);
         
         if (!$form->isValid()) {
-            return $this->render('user/reset_password.twig', ['form' => $form->createView()]);
```
localheinz: While at it, I figured it would make sense to streamline how we pass data to the templates - this is an example.

classes/Http/Controller/ProfileController.php
```diff
@@ -119,12 +117,12 @@ public function processAction(Request $request)
             'ext'   => \implode('<br>', $form->getErrorMessages()),
         ]);
 
-        $formData['formAction'] = $this->url('user_update');
-        $formData['buttonInfo'] = 'Update Profile';
-        $formData['id']         = $userId;
-        $formData['flash']      = $session->get('flash');
-
-        return $this->render('user/edit.twig', $formData);
+        return $this->render('user/edit.twig', \array_merge($formData, [
```
localheinz: Merging in additional / replacement data here instead of assigning as before.

classes/Http/Controller/Reviewer/TalksController.php
```diff
@@ -85,7 +83,9 @@ public function viewAction(Request $request)
             return $this->app->redirect($this->url('admin_talks'));
         }
 
-        return $this->render('reviewer/talks/view.twig', ['talk' => $handler->getProfile()]);
```
localheinz: While at it, I figured it would make sense to streamline how we pass data to the templates - this is an example.

classes/Http/Controller/TalkController.php
```diff
@@ -61,14 +61,18 @@ public function viewAction(Request $request)
         /* @var Speakers $speakers */
         $speakers = $this->service('application.speakers');
 
+        $talkId = (int) $request->get('id');
```
localheinz: While at it, I'm moving this out of the `try` block, as I don't see anything that could go wrong here.

classes/Http/Controller/TalkController.php
```diff
@@ -61,14 +61,18 @@ public function viewAction(Request $request)
...
+
         try {
-            $talkId = (int) $request->get('id');
-            $talk   = $speakers->getTalk($talkId);
+            $talk = $speakers->getTalk($talkId);
         } catch (NotAuthorizedException $e) {
             return $this->redirectTo('dashboard');
         }
 
-        return $this->render('talk/view.twig', \compact('talkId', 'talk'));
```
localheinz: While at it, removing the usage of `compact()` here.

localheinz:
Thank you, @BackEndTea!

PR #850, opened 2017-12-05
Controllers as services

derrabus:
This PR removes the `ContainerAware` trait from all controllers, thus making them independant from the Silex application. This refactoring can be a preparation for migrating away from Silex.

Related to #618.

PR #852, opened 2017-12-05
Fix: Add return type declarations to controller actions

localheinz:
This PR

* [x] asserts that `public` methods in controllers use `Symfony\Component\HttpFoundation\Response` as return type declaration
* [x] adds return type declarations to controller actions

Somewhat related to #850.

💁‍♂️ There are a few controller actions that actually also return `false`, not sure what's kicking in there later to actually return a response.

### DIFF
--- a/tests/Unit/ProjectCodeTest.php
+++ b/tests/Unit/ProjectCodeTest.php
@@ -13,12 +13,14 @@ declare(strict_types=1);
 
 namespace OpenCFP\Test\Unit;
 
+use Localheinz\Classy;
 use Localheinz\Test\Util\Helper;
 use OpenCFP\Domain;
 use OpenCFP\Http;
 use OpenCFP\Infrastructure;
 use OpenCFP\Provider;
 use PHPUnit\Framework;
+use Symfony\Component\HttpFoundation;
 
 /**
  * @coversNothing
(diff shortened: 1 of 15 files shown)


BackEndTea:
@localheinz I'd like to change those who currently return false to return some kind of json object. Since the false gets seen as a success by the ajax call most of the time. So the front end is never informed of the failure.

localheinz:
@BackEndTea 

Let me put this PR on hold until your suggestion has been sorted out!

BackEndTea:
@localheinz If you are still working on this, it might be a good idea to add a test that ensures all public functions of a controller return a ```Response object```

I threw this together
```php
    public function testControllersReturnResponse()
    {
        $controllers = [
            Http\Controller\Admin\DashboardController::class,
            Http\Controller\Admin\ExportsController::class,
            Http\Controller\Admin\SpeakersController::class,
            Http\Controller\Admin\TalksController::class,
            Http\Controller\DashboardController::class,
            Http\Controller\ForgotController::class,
            Http\Controller\PagesController::class,
            Http\Controller\ProfileController::class,
            Http\Controller\Reviewer\DashboardController::class,
            Http\Controller\Reviewer\SpeakersController::class,
            Http\Controller\Reviewer\TalksController::class,
            Http\Controller\SecurityController::class,
            Http\Controller\SignupController::class,
            Http\Controller\TalkController::class,
        ];
        foreach ($controllers as $controller) {
            foreach (\get_class_methods($controller) as $method) {
                $methodReflection = new \ReflectionMethod($controller, $method);
                $this->assertEquals(Response::class,
                    $methodReflection->getReturnType(),
                    'Failed asserting that' . $method . ' in the class' . $controller .' returns a Response '
                );
            }
        }

    }
```
Thought it may be a better idea to auto generate the list of controllers, in case a new controller gets added

localheinz:
@BackEndTea 

Your recommendation is good, and I had something similar planned already to assert that controller action names are suffixed with `Action()` - we can probably mix in your suggestion here, too!

BackEndTea:
@localheinz 

If we make a lot of these kind of tests, we might want to split those from the normal integration and unit tests, and during CI, we can run them during the style check. Since those don't need to be checked on every php version.

tests/Unit/ProjectCodeTest.php
```diff
@@ -83,6 +85,50 @@ public function testProductionClassesHaveUnitTests()
             ]
         );
     }
+
+    public function testControllerActionsUseResponseReturnType()
```
BackEndTea: I think it may be cleaner to split up this function a bit. Prehaps move the retrieval of controllers to a dataprovider

localheinz: I am not using a data provider here because we want a *single* instead of *n* failure messages.

localheinz: If we want to add more tests, we can extract a method that returns all controllers.

localheinz:
Thank you, @BackEndTea!

PR #853, opened 2017-12-05
Fix: Use data provider

localheinz:
This PR

* [x] uses data providers

Related to https://github.com/opencfp/opencfp/pull/852#issuecomment-349336156.

💁‍♂️ Probably best reviewed with https://github.com/opencfp/opencfp/pull/853/files?diff=split.

### DIFF
--- a/tests/Integration/Http/Controller/Admin/TalksControllerTest.php
+++ b/tests/Integration/Http/Controller/Admin/TalksControllerTest.php
@@ -185,51 +185,57 @@ final class TalksControllerTest extends WebTestCase
 
     /**
      * @test
+     * @dataProvider providerValidRating
+     *
+     * @param mixed $rating
      */
-    public function rateActionWorksCorrectly()
+    public function rateActionWorksCorrectly($rating)
     {
         $talk = self::$talks->first();
 
         $this->asAdmin()
-            ->post('/admin/talks/' . $talk->id . '/rate', ['rating' => 1])
+            ->post('/admin/talks/' . $talk->id . '/rate', ['rating' => $rating])
             ->assertSee('1')
             ->assertSuccessful();
     }
 
-    /**
-     * @test
-     */
-    public function rateActionRetunsFalseOnWrongRate()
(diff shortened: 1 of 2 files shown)


tests/Integration/Http/Controller/Admin/TalksControllerTest.php
```diff
@@ -185,51 +185,61 @@ public function favoriteActionDoesNotErrorWhenTryingToDeleteFavoriteThatDoesNoEx
...
+            'int'        => 1,
+            'integerish' => '0',
+        ];
 
-        $this->asAdmin()
-            ->post('/admin/talks/' . $talk->id . '/rate', ['rating' => 12])
-            ->assertNotSee('1')
-            ->assertSuccessful();
+        foreach ($values as $key => $value) {
+            yield $key => [
```
mdwheele: First, big fan of named datasets. Really helps with failure messages!

That said, I'm not "against" using `yield` here, but is it really much better than just:

```php
return [
    'int' => [1],
    'integerish' => ['0']
];
```

I mostly mention because:

1. It's repeated several times / boilerplate
2. Most folks familiar with `@dataProvider` are used to the array of arrays that is each case's arguments
3. Only benefit I see is memory footprint of tests is smaller as only one dataset is loaded at a time? 

Thoughts? I've never seen this done (which isn't a critique, but more to say I might be ignorant or missing the benefit).

localheinz: @mdwheele 

I've been using `Generator`s for data providers mostly because I'm too lazy to wrap the set of arguments that should be passed to the test in an array, hehe!

mdwheele: Hahahah! If that's the only reason, I think I'd prefer simply returning an array. It's 4 braces vs variable assignment + yields

You're a rockstar. :tada:

localheinz:
Thank you, @BackEndTea!

PR #854, opened 2017-12-05
Fix: Simplify ControllerResolver

localheinz:
This PR

* [x] simplifies the `ControllerResolver`

Related to https://github.com/opencfp/opencfp/pull/850#issuecomment-349365995.


derrabus:
You could simplify it even more, if you override `instantiateController` instead of `createController`.

localheinz:
@derrabus 

Nice, makes it even more readable!

localheinz:
Thank you, @chartjes and @derrabus!

PR #855, opened 2017-12-05
Load all controllers as services

derrabus:
This PR is basically a light version of #850.

All controllers a being load from the container. While the `BaseController` class receives its dependencies via DI, most child controllers still use the `ContainerAware` trait.

Merging this PR to master would allow me to split #850 into smaller PRs that are easier to review and maintain. Also, the diff of #850 itself would become smaller.


mdwheele:
:tada:

PR #860, opened 2017-12-05
Removed ContainerAware from TalkController

derrabus:
The ProfileController is being constructed via DI.

This is a rather large controller with a lot of dependencies that are not used by all actions. It should definitely be split up in a later PR.

Follows #855.
Related to #618.


derrabus:
Done.

mdwheele:
Thanks @derrabus !

PR #866, opened 2017-12-05
Converted error middleware into an event subscriber

derrabus:
Preparation for #618: Middlewares are a Silex-only feature. Under the hood they're implemented as event listeners. So if we want them to survive the framework switch, we need to turn them into event subscribers.

This PR does this for the error middleware.

After the switch, this subscriber can probably be replaced with the exception listener that is provided by TwigBundle.


mdwheele:
Excellent! Thanks @derrabus

PR #867, opened 2017-12-05
Converted authentication middleware into an event subscriber

derrabus:
Preparation for #618: All middlewares that check for an authenticated user have been replaced by a single event subscriber that can be reused in Symfony.

### DIFF
--- a/tests/Unit/Infrastructure/Event/AuthenticationListenerTest.php
+++ b/tests/Unit/Infrastructure/Event/AuthenticationListenerTest.php
@@ -1,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Copyright (c) 2013-2017 OpenCFP
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ *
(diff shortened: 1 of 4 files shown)


tests/Unit/Infrastructure/Event/AuthenticationListenerTest.php
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @see https://github.com/opencfp/opencfp
+ */
+
+namespace OpenCFP\Test\Unit\Infrastructure\Event;
+
+use OpenCFP\Test\WebTestCase;
+
+/**
+ * @coversNothing
```
BackEndTea: Could you add  ```@covers \OpenCFP\Infrastructure\Event\AuthenticationListener``` instead of ```@coversNothing``` here

derrabus: Sure, sorry. I did not notice that CS fixer adds those.

BackEndTea:
Looks good to me, what do you think @mdwheele

mdwheele:
:+1:

PR #868, opened 2017-12-06
Fix: Add @covers annotation to EventListenerTest

BackEndTea:
This PR

* [x] Adds an ```@covers``` annotation to the ```ExceptionListenerTest```

Follows: #866

PR #869, opened 2017-12-06
Fixed covers annotation.

derrabus:
In my last PR I forgot to adjust the `@covers` annotation. Sorry for that.

Follows #866.


derrabus:
Oh, somebody way faster, see #868.

PR #870, opened 2017-12-06
Fix: Use more specific assertions for responses

localheinz:
This PR

* [x] uses more specific assertions when testing the rating system

Follows #853.
Follows https://github.com/opencfp/opencfp/pull/852#issuecomment-349339120.

### DIFF
--- a/tests/Integration/Http/Controller/Admin/TalksControllerTest.php
+++ b/tests/Integration/Http/Controller/Admin/TalksControllerTest.php
@@ -193,10 +193,13 @@ final class TalksControllerTest extends WebTestCase
     {
         $talk = self::$talks->first();
 
-        $this->asAdmin()
-            ->post('/admin/talks/' . $talk->id . '/rate', ['rating' => $rating])
-            ->assertSee('1')
-            ->assertSuccessful();
+        $response = $this->asAdmin()->post('/admin/talks/' . $talk->id . '/rate', [
+            'rating' => $rating,
+        ]);
+
+        $response->assertSuccessful();
+
+        $this->assertSame('1', $response->getContent());
     }
 
     public function providerValidRating(): array
@@ -221,10 +224,13 @@ final class TalksControllerTest extends WebTestCase
     {
         $talk = self::$talks->first();
 
-        $this->asAdmin()
-            ->post('/admin/talks/' . $talk->id . '/rate', ['rating' => $rating])
-            ->assertNotSee('1')
-            ->assertSuccessful();
+        $response = $this->asAdmin()->post('/admin/talks/' . $talk->id . '/rate', [
+            'rating' => $rating,
+        ]);
+
+        $response->assertSuccessful();
+
+        $this->assertSame('', $response->getContent());
     }
 
     public function providerInvalidRating(): array
--- a/tests/Integration/Http/Controller/Reviewer/TalksControllerTest.php
+++ b/tests/Integration/Http/Controller/Reviewer/TalksControllerTest.php
@@ -79,10 +79,13 @@ final class TalksControllerTest extends WebTestCase
     {
         $talk = self::$talks->first();
 
-        $this->asReviewer()
-            ->post('/reviewer/talks/' . $talk->id . '/rate', ['rating' => $rating])
-            ->assertSee('1')
-            ->assertSuccessful();
+        $response = $this->asReviewer()->post('/reviewer/talks/' . $talk->id . '/rate', [
+            'rating' => $rating,
+        ]);
+
+        $response->assertSuccessful();
+        
+        $this->assertSame('1', $response->getContent());
     }
 
     public function providerValidRating(): array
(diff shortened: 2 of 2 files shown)


tests/Integration/Http/Controller/Admin/TalksControllerTest.php
```diff
@@ -193,10 +194,12 @@ public function rateActionWorksCorrectly($rating)
     {
         $talk = self::$talks->first();
 
-        $this->asAdmin()
-            ->post('/admin/talks/' . $talk->id . '/rate', ['rating' => $rating])
-            ->assertSee('1')
```
localheinz: This only asserts that the response *contains* `'1'`.

tests/Integration/Http/Controller/Admin/TalksControllerTest.php
```diff
@@ -221,10 +224,12 @@ public function rateActionReturnsFalseOnWrongRate($rating)
     {
         $talk = self::$talks->first();
 
-        $this->asAdmin()
-            ->post('/admin/talks/' . $talk->id . '/rate', ['rating' => $rating])
-            ->assertNotSee('1')
```
localheinz: This only asserts that the response does *not* contains `'1'`.

tests/Integration/Http/Controller/Admin/TalksControllerTest.php
```diff
@@ -193,10 +194,12 @@ public function rateActionWorksCorrectly($rating)
...
 
-        $this->asAdmin()
-            ->post('/admin/talks/' . $talk->id . '/rate', ['rating' => $rating])
-            ->assertSee('1')
-            ->assertSuccessful();
+        $response = $this->asAdmin()->post('/admin/talks/' . $talk->id . '/rate', [
+            'rating' => $rating,
+        ]);
+
+        $this->assertSame(Response::HTTP_OK, $response->getStatusCode());
```
BackEndTea: Can't we use ```assertSuccessful```?

tests/Integration/Http/Controller/Reviewer/TalksControllerTest.php
```diff
@@ -79,10 +79,13 @@ public function rateActionWorksCorrectly($rating)
...
-            ->post('/reviewer/talks/' . $talk->id . '/rate', ['rating' => $rating])
-            ->assertSee('1')
-            ->assertSuccessful();
+        $response = $this->asReviewer()->post('/reviewer/talks/' . $talk->id . '/rate', [
+            'rating' => $rating,
+        ]);
+
+        $response->assertSuccessful();
+        
+        $this->assertSame('1', $response->getContent());
```
BackEndTea: What is the difference between this and ```assertSee('1')```?

localheinz: [`TestResponse::assertSee()`(https://github.com/opencfp/opencfp/blob/0bede4bdf0c6e5eaa3ac1bede442651961bfd2ae/tests/TestResponse.php#L85-L90) asserts that the response body *contains* `'1'`, but we want to assert that it is the same.

On a separate note, while I understand that the idea behind the `TestResponse` is to be able to use the fluent interface provided by it to perform assertions on itself, I am inclined not to use `TestResponse` (and I would prefer to remove it) for the following reasons

* it's not the responsibility of the `TestResponse` or any `Response` object to make assertions about itself
* we will end up adding more and more assertion methods to it

/cc @mdwheele

mdwheele: `TestResponse` is a decorator to provide assertions against a `Response`, not itself. I consider it tightly-coupled to `WebTestCase` so it's removal would mean addition of traits or helper methods to `WebTestCase` to provide the same functionality.

localheinz: @mdwheele 

Sure, it *composes* the actual response object, but by definition, it also *is* a response object - and from the test perspective, we send messages to it. That the response decorator then performs assertions on the composed response is something different.

I would prefer to provide additional - domain-specific - assertions via a trait.

mdwheele: Could we open a separate issue to discuss before a PR is sent to remove? It's certainly outside of the scope of work here and not worth blocking this PR.

To be honest, I'm unsure how much `WebTestCase` really offers, if anything, other than a crutch to make it easier to test an interface that feels more "end to end". If we were using Behat, would we even be having this discussion? Maybe not! I'd love to discuss / organize a plan in another issue.

Thanks @localheinz!

localheinz: @mdwheele 

Not intending to remove anything within this PR, I prefer small PRs, addressing one issue at a time!

localheinz: @mdwheele 

For the record, opened an issue at #877!

BackEndTea:
@mdwheele Are we good on merging this?

mdwheele:
@BackEndTea Yeah I'm :+1:

localheinz:
Thank you, @mdwheele!

PR #871, opened 2017-12-06
Refactored twig middleware into an event subscriber and a twig extension

derrabus:
Preparation for #618:
* The middleware that sets global Twig variables based on the current request has been refactored to an event subscriber.
* All custom Twig functions have been bundled in a single Twig extension.
* Some tests have been added to cover that functionality.

It should be possible to reuse both classes when switching to Symfony.

### DIFF
--- a/tests/Unit/Infrastructure/Event/TwigGlobalsListenerTest.php
+++ b/tests/Unit/Infrastructure/Event/TwigGlobalsListenerTest.php
@@ -1,0 +1,121 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Copyright (c) 2013-2017 OpenCFP
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ *
+ * @see https://github.com/opencfp/opencfp
+ */
+
+namespace OpenCFP\Test\Unit\Infrastructure\Event;
+
+use Mockery;
+use OpenCFP\Domain\CallForPapers;
+use OpenCFP\Domain\Services\Authentication;
+use OpenCFP\Infrastructure\Auth\UserInterface;
+use OpenCFP\Infrastructure\Event\TwigGlobalsListener;
+use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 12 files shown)


tests/Unit/Infrastructure/Event/TwigGlobalsListenerTest.php
```diff
@@ -0,0 +1,121 @@
...
+            Request::create($uri),
+            HttpKernelInterface::MASTER_REQUEST
+        ));
+
+        $output = $twig->render('globals.txt.twig');
+
+        $this->assertStringEqualsFile(__DIR__ . '/Fixtures/' . $fixture, $output);
+    }
+
+    public function provideTestSetup(): iterable
```
mdwheele:
```
Fatal error: Generators may only declare a return type of Generator, Iterator or Traversable, OpenCFP\Test\Unit\Infrastructure\Event\iterable is not permitted in /home/travis/build/opencfp/opencfp/tests/Unit/Infrastructure/Event/TwigGlobalsListenerTest.php on line 69
```

This fails on PHP 7 because `iterable` wasn't a thing at that point. 

Could this `@dataProvider` be modified (and return type adjusted):

```
return [
    [$this->mockAdmin(), true, '/foo', null, 'global_with_admin.txt'],
    [$this->mockReviewer(), true, '/foo', 'You\'ve got mail.', 'global_with_reviewer.txt'],
    [$this->mockAnonymous(), false, '/bar', null, 'global_with_anonymous_user.txt']
]
```

I'm unsure what benefit this provides beyond personal preference. I'm approaching this mostly from a consistency perspective. Obviously, if there is clear reason to return an `iterable` for a specific `TestCase`, we can do it. The examples I've seen recently coming in with contributions seem to be preference-based.

derrabus: Fixed.

mdwheele:
I've created #873 to discuss what, if any preference, we have (for consistency-sake). It's just one of those things that hasn't really come up, but I think would be a reasonable thing to get on the same page about.

Thanks! This looks great.

PR #874, opened 2017-12-06
Fix: Explicitly return Response objects

localheinz:
This PR

* [x] explicitly returns `Response` objects from controller actions

Follows https://github.com/opencfp/opencfp/pull/852#issuecomment-349339120.

### DIFF
--- a/classes/Http/Controller/Reviewer/TalksController.php
+++ b/classes/Http/Controller/Reviewer/TalksController.php
@@ -20,6 +20,7 @@ use OpenCFP\Domain\Talk\TalkHandler;
 use OpenCFP\Domain\ValidationException;
 use OpenCFP\Http\Controller\BaseController;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Twig_Environment;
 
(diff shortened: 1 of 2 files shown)


BackEndTea:
I'm assuming these don't have return type declarations, as those will be added in #852 ?

localheinz:
@BackEndTea 

> I'm assuming these don't have return type declarations, as those will be added in #852 ?

Exactly!

localheinz:
Thank you, @BackEndTea!

derrabus:
Good catch. :-)

classes/Http/Controller/Reviewer/TalksController.php
```diff
@@ -113,11 +114,15 @@ public function rateAction(Request $request)
                 'rating' => 'required|integer',
             ]);
 
-            return $this->talkHandler
+            $content = (string) $this->talkHandler
                 ->grabTalk((int) $request->get('id'))
                 ->rate((int) $request->get('rating'));
+
+            return $content;
```
localheinz: I made a mistake here, see #878.

classes/Http/Controller/Reviewer/TalksController.php
```diff
@@ -113,11 +114,15 @@ public function rateAction(Request $request)
...
         } catch (ValidationException $e) {
-            return false;
+            $content = '';
         }
+
+        return new Response($content);
```
localheinz: In a separate PR, we should probably not just return a plain text response with HTTP status code 200, but

* a JSON response
* use appropriate response codes (`200`, `400`)

PR #875, opened 2017-12-06
Fix: Move integration tests

localheinz:
This PR

* [x] moves integration tests into the `OpenCFP\Test\Integration` namespace


tests/Integration/Http/Controller/Admin/DashboardControllerTest.php
```diff
@@ -15,7 +15,7 @@
 
 use OpenCFP\Domain\Model\Talk;
 use OpenCFP\Test\Helper\RefreshDatabase;
-use OpenCFP\Test\WebTestCase;
```
localheinz: As far as I can tell, extending this `abstract` test class indicates an integration test.

tests/Unit/ProjectCodeTest.php
```diff
@@ -60,6 +61,10 @@ public function testProductionClassesHaveUnitTests()
                 Http\Form\ForgotForm::class,
                 Http\Form\ResetForm::class,
                 Http\View\TalkHelper::class,
+                Infrastructure\Event\AuthenticationListener::class,
+                Infrastructure\Event\ExceptionListener::class,
```
BackEndTea: If i'm correct this list contains classes that don't have tests. Don't these have tests?

localheinz:
Thank you, @chartjes!

PR #876, opened 2017-12-06
Decoupled direct container access in tests via PSR-11

derrabus:
This PR decouples most tests from the Pimple container by making them access services through the PSR-11 interface. This way, it should not matter to them whether the container is Pimple or Symfony DI.

Related to #618.

TODO in a follow-up: The service swapping is going to be interesting with a compiled container. 😱


mdwheele:
Awesome! I wanted to leave a response to "service swapping". I feel like if we were in a better spot regarding DI, in general (which we are, with your contributions to controllers), we might not need to rely as much on swapping services during test run-time. There are follow-up patches we could make to support it, but knowing that there are plans to implement a better e2e suite, I wonder how much benefit testing at the kernel-level provides (not just technically, but from a DX perspective).

PR #878, opened 2017-12-06
Fix: Do not return string

localheinz:
This PR

* [x] removes a `return` statement 

Follows #874.
Blocks #852.


classes/Http/Controller/Reviewer/TalksController.php
```diff
@@ -117,8 +117,6 @@ public function rateAction(Request $request)
             $content = (string) $this->talkHandler
                 ->grabTalk((int) $request->get('id'))
                 ->rate((int) $request->get('rating'));
-
-            return $content;
```
localheinz: We don't actually want to return a `string` here, but a `Response` object.

localheinz:
@mdwheele 

Can we get this merged so I can open #852 for review?

localheinz:
Thank you, @BackEndTea and mdwheele!

Issue #873, opened 2017-12-06
Do we prefer `@dataProviders` stick to array-of-arg-arrays, yield or /shrug?


mdwheele:
Related: https://github.com/opencfp/opencfp/pull/853#discussion_r155086734
Related: https://github.com/opencfp/opencfp/pull/871#discussion_r155227907

We've had a few PRs come in that use some arrangement of `yield` to return a `Generator` or `iterable` object. It looks like in both cases, this was done for readability, personal-preference, etc. My reviews have suggested sticking with array-of-arg-array madness as a matter of consistency, mostly. However, I am a proponent of doing what's most readable at the time with some reasonable default suggestions.

Things that go through my mind with data providers / readability:

* Named data-sets are a wonderful thing to help distinguish meaning / purpose for that dataset. They also improve test output on failure. Current codebase doesn't use them nearly enough.
  * #853 uses named data-sets to distinguish different scenarios
  * #871 returns `iterable` and uses multiple `yield` statements to separate.

  Thoughts: I would prefer use of named data-sets for new code, no matter the return type. We could definitely send some TLC to existing data providers.
* The biggest argument against array-of-arg-array madness is... it's mad. To me, it is a smell to have a test method that receives > 3 arguments from a data-provider. There are exceptions, but most times, it is a signal that the test method might be doing too much and is an opportunity to rethink models. This obviously doesn't apply so much to initial PRs against legacy, but that's my personal thought process.
  * Under that heuristic, the readability of array-of-arg-array madness is improved and is arguably equivalent to any `yield`-based solution.
  * Considering cases where a single argument is being sent, it's actually more code / arguably less-readable.

I'm making this issue to just open the discussion, but I don't think we need to bikeshed too much or anything like that. For now, I'm going to review in terms of consistency and we can organize a more-widespread effort to improve tests with regard to data providers in separate PRs.

chartjes:
I want to see:

* named data sets
* continue to use array-of-arg-array for data providers

It also never occurred to me that too many arguments from a data provider was a combination test-and-code smell but upon reflection that makes sense.

I will submit a PR adding this requirement to the CONTRIBUING.md document

localheinz:
What about naming of data providers?

We could enforce that data providers follow a specific naming scheme, too.

chartjes:
@localheinz I am not sure how much I care about the name of a data provider, but open to suggestions

mdwheele:
For reference, this code...

```php
<?php
[...]

PR #880, opened 2017-12-06
Added some details about test contributions

chartjes:
This PR

* [x] Addresses issues outline in #873


mdwheele:
Could we decide on `snake_case_test_methods` vs `camelCaseTestMethods`? 

I think `snake_case` is more readable, but things like this get shoved to "personal preference" a lot. What's our project's preference?

PR #881, opened 2017-12-06
Enhancement: Extract test helper trait

### LINKED ISSUE #877: Do we want to remove TestResponse in favour of providing assertions elsewhere?
Currently, we use the [`TestResponse`](https://github.com/opencfp/opencfp/blob/ee3305b58f7054f2b242b748a29bc7fb68ad3806/tests/TestResponse.php) class as a means to simplify assertions on a response object returned from invoking a controller action.

As pointed out in https://github.com/opencfp/opencfp/pull/870/files#r155239203, I would suggest to remove the `TestResponse` class and instead provide domain-specific assertions either via 

- a trait
- an abstract test case

localheinz:
This PR

* [x] extracts a test helper trait

Fixes #877.


tests/Helper/ResponseHelper.php
```diff
@@ -0,0 +1,157 @@
...
+ */
+
+namespace OpenCFP\Test\Helper;
+
+use Symfony\Component\HttpFoundation;
+
+/**
+ * @mixin \PHPUnit\Framework\TestCase
+ */
+trait ResponseHelper
```
localheinz: This trait is currently only used by the `WebTestCase`, but I am hesitant to pull it up into the `WebTestCase` itself, as I feel the assertions should not be able to access any internal state of `WebTestCase` and children of it.

tests/Helper/ResponseHelper.php
```diff
@@ -0,0 +1,157 @@
...
+{
+    final protected function assertResponseIsSuccessful(HttpFoundation\Response $response)
+    {
+        $this->assertTrue($response->isSuccessful(), 'Failed asserting that the response is successful.');
+    }
+
+    final protected function assertResponseIsRedirect(HttpFoundation\Response $response)
```
localheinz: About `assertResponseIsSuccessful()` and `assertResponseIsRedirect()`:

Personally, I would remove them. They are not specific enough. For example, we probably want to assert that a redirect is permanent, or temporary. Same for successful responses, we probably want to be more specific as well and assert the *actual* status code.

What do you think?

mdwheele: I like them. Unless I'm building a JSON API, I'm not sure I find myself writing specs that specific. For typical web interactions, I mostly just want to know "they were redirected to the correct endpoint after submission". I'm not sure I care `301` vs `302` and might see that as over-specification.

What if we had assertions per response class (`2xx`, `3xx`, `4xx`, `5xx`) with the option of passing in a second argument for specific response code? It's just syntactic sugar, but I think that's important for readability. 

Case in point, `assertResponseStatus(Response::HTTP_OK, $response)` hurts my eyes more than `assertResponseIsSuccessful($response)`.

tests/Helper/ResponseHelper.php
mdwheele: This turned out similar to what I expected. There is a lot of repetition in readability of tests because we've extracted, more or less a stateless procedural interface for assertions on `Response`. I think that any `WebTestCase` (or Behat `FeatureContext`, more aptly named) carries with it a usefully stateful context, by definition. 

It's the difference between:

```php
# Current behaviour and arguably ambiguous. Is this an assertion? How do I know?
$this->get('/')->isSuccessful();
```
and...
```php
// Okay, makes sense.
$response = $this->get('/');

// I feel like I'm typing response over and over again. 
// It's mechanical-feeling but is 100% an assertion.
$this->assertResponseIsSuccessful($response); // response is response is response is.
```

Personally (and I think this is really just preferences honestly) I'm fine with decorating a response in tests and making assertions through it. It's 6 vs a half-dozen compared to a procedural interface through trait, in my opinion.

It's certainly readable. What if we removed the ambiguity of "is this an assertion"? Something like:

```php
// Normal get request.
$this->get('/')
    ->assertSuccessful()
    ->assertSee('OpenCon 2017');

// I'm creating a resource via API (D:)
$this->post('/users', [ /* data */ ])
    ->assertCreated();   // vs assertSuccessful(Response::HTTP_CREATED)
```

If we're definitely moving away from a decorator with assertions, I am currently leaning more towards stateful assertions on `WebTestCase`, itself to maximize readability. Could reset `$this->response` on `call(...)` and everything would be good to go. 

To completely flip the table, I really do feel like the original purpose of `WebTestCase` was to provide a more-human interface for making high-level behavioural assertions from and end-to-end perspective. Like I've said previously, it may just be time to actually make that move.

localheinz:
Thank you, @chartjes and @mdwheele!

@mdwheele 

About hurting the eyes, I *think* I understand what you mean, and I agree. I think one of the reasons it doesn't hurt my eyes o much is that I think I have gotten used to my eyes hurting.

But, maybe the best part of what I did here was just separating 

* building and issuing a request
* making assertions about *components* of the response

I have an idea of what we could improve, and maybe that would move more into the direction I think you intended to move things, but you tell me, hehe!

Maybe it was just that the naming `TestResponse` and some other things made me want to change things. Maybe we can just move further and dissect it. After all, we are trying to make assertions about components.

Some things I disliked about `isSuccessful()` and `isRedirect()`:

* it's not specific enough
* it ignores context (for example, depending on context, I would understand things are a success if we sent a `400` response status code, as opposed to sending a `200` response status code with an error message)

For example, we could have some were generic "assertions", that don't actually assert much, but return specific objects with fluent interfaces, as entry-points to provide more readable assertions (much like the initial `TestResponse`)

Imagine this:

```php
$response = $this->get('/foo/bar');

$this->assertResponse($response)->isSuccessful();

$this->assertResponseStatus($response)
    ->isValid()
    ->is(HttpFoundation\Response::HTTP_OK);

$this->assertResponseHeader($response)
    ->has('Content-Type)
    ->hasContentTypeEqualTo('application/json')
    ->has('Location')
    ->hasLocationEqualTo('/foo/bar/baz')
    ->hasCookie()
    ->hasCookieEqualTo('hahaha')
    ->hasCookieExpiringIn(9001)

$this->assertResponseBody()
    ->isEmpty()
    ->isJson()
    ->isJsonEqualToString($jsonString)
    ->isJsonEqualToArray($jsonArray)
    ->contains('Hey, right now is not a good time!')
    ->isXml();
```

And so on!

Would that make any sense, and hurt the eyes less?

```php
final protected function assertResponse($response): ResponseAssertion;
final protected function assertResponseBody($response): ResponseBodyAssertion;
final protected function assertResponseHeader($response): ResponseHeaderAssertion;
final protected function assertResponseStatus($response): ResponseStatusAssertion;
```

Naming is just an idea, of course.

This way we could provide meaningful assertions per *component* (header, body, and header probably split up a bit) of the response. Or the request. And the session.

mdwheele:
@localheinz 

You've got it! Haha. Sorry, these types of things are hard to articulate for me, sometimes. Especially via text! I promise there is a lot of body language going on behind the keyboard that would probably help :grin: 

I like the direction you propose. I have some thoughts below based on the samples you provided above. Maybe we could take this up in an issue somewhere or a follow-up PR.

```php
// I like the separation. We make a specific request; get a specific response.
// It's not cluttered with assertions, etc.
$response = $this->get('/foo/bar');

// I can get down with this.
// "assertResponse($response)" becomes "Assert this *particular* response..." in my brain.
$this->assertResponse($response)->isSuccessful();

// This starts to trigger the "eye hurt" (which I don't mean to be derogatory, just 
// hard to articulate... like "cringe" or "wincing") I also recognize we're "riffing" and 
// playing around with interfaces here.
$this->assertResponseStatus($response)
    ->isValid()  // Unsure what's here, but it doesn't matter.
    ->is(HttpFoundation\Response::HTTP_OK);

// What about...
$this->assertResponse($response)
    ->wasSuccessful();  // Sugar on top of hasStatus(HttpFoundation\Response::HTTP_OK)

// "Assert this response was successful."
// "Assert this response has status 200."
// I would expect how the interface reads to influence the error messages when 
// assertions fail.

// Riffing on...
$this->assertResponseHeader($response)
    ->has('Content-Type')
    ->hasContentTypeEqualTo('application/json')
    ->has('Location')
    ->hasLocationEqualTo('/foo/bar/baz')
    ->hasCookie()
    ->hasCookieEqualTo('hahaha')
    ->hasCookieExpiringIn(9001)

// What about...
$this->assertResponse($response)->header()
    ->has('Content-Type', 'application/json')  // assertSame
    ->has('Location', '/foo/bar/baz')
    ->has('Cookie', 'hahaha', 9001)  // or ->has(...)->expiringIn(9001);
```

Having `assertResponse` return an assertion that also had factory methods for different component-level assertions of the response (header, status, body, etc.) might be a good way to go, indeed.

localheinz:
@mdwheele 

Since I figured this could be useful beyond using in `opencfp/opencfp`, I started working on something at https://github.com/localheinz/symfony-test-util.

Got to get going (🎄 party), but will be back later. I think this has potential!

PR #883, opened 2017-12-06
Removed orphaned routes.

derrabus:
While analyzing the routing, I've found some routes that point to controller actions that do not exist (anymore?). I think, we can remove them.

### DIFF
--- a/classes/Provider/Gateways/WebGatewayProvider.php
+++ b/classes/Provider/Gateways/WebGatewayProvider.php
@@ -235,8 +235,6 @@ class WebGatewayProvider implements BootableProviderInterface, ServiceProviderIn
             ->bind('user_new');
         $web->post('/signup', 'OpenCFP\Http\Controller\SignupController::processAction')
             ->bind('user_create');
-        $web->get('/signup/success', 'OpenCFP\Http\Controller\SignupController::successAction')
-            ->bind('user_success');
 
         // Edit Profile/Account
         $web->get('/profile/edit/{id}', 'OpenCFP\Http\Controller\ProfileController::editAction')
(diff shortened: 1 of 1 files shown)


classes/Provider/Gateways/WebGatewayProvider.php
```diff
@@ -318,10 +314,6 @@ public function boot(Application $app)
             ->bind('reviewer_talks');
         $reviewer->get('/talks/{id}', 'OpenCFP\Http\Controller\Reviewer\TalksController::viewAction')
             ->bind('reviewer_talk_view');
-        $reviewer->post('/talks/{id}/favorite', 'OpenCFP\Http\Controller\Reviewer\TalksController::favoriteAction')
-            ->bind('reviewer_talk_favorite');
-        $reviewer->post('/talks/{id}/comment', 'OpenCFP\Http\Controller\Reviewer\TalksController::commentCreateAction')
-            ->bind('reviewer_talk_comment_create');
```
localheinz: As far as I can tell, both of these routes map to controller actions which exist. Are we sure we want to remove them?

classes/Provider/Gateways/WebGatewayProvider.php
```diff
@@ -253,8 +251,6 @@ public function boot(Application $app)
             ->bind('forgot_password');
         $web->post('/forgot', 'OpenCFP\Http\Controller\ForgotController::sendResetAction')
             ->bind('forgot_password_create');
-        $web->get('/forgot_success', 'OpenCFP\Http\Controller\ForgotController::successAction')
```
localheinz: 👍 Controller action doesn't exist.

classes/Provider/Gateways/WebGatewayProvider.php
```diff
@@ -235,8 +235,6 @@ public function boot(Application $app)
             ->bind('user_new');
         $web->post('/signup', 'OpenCFP\Http\Controller\SignupController::processAction')
             ->bind('user_create');
-        $web->get('/signup/success', 'OpenCFP\Http\Controller\SignupController::successAction')
```
localheinz: 👍 Controller action doesn't exist.

classes/Provider/Gateways/WebGatewayProvider.php
derrabus: They exist on the `Admin\TalksController`, but not on the `Reviewer\TalksController`.

derrabus: Also, I couldn't find any references to the names of both routes apart from the routing configuration itself.

BackEndTea: This one is my fault 😅 I added routes for the favorite and comment actions for the reviewer when creating all the routes, but later the decision was made to not allow them those privileges.

localheinz:

PR #886, opened 2017-12-07
Moved Eloquent setup into a very early event listener

derrabus:
The setup logic for Eloquent was moved out of the application class.

Related to #618.


classes/Provider/ApplicationServiceProvider.php
```diff
@@ -91,4 +94,11 @@ public function register(Container $app)
             return new PseudoRandomStringGenerator();
         };
     }
+
+    public function subscribe(Container $app, EventDispatcherInterface $dispatcher)
+    {
+        $dispatcher->addSubscriber(new DatabaseSetupListener(
+            $app[Capsule::class]
```
localheinz: Since we only pass in  a single argument to the constructor of  `DatabaseSetupListener`, can we just instantiate the latter in a single line?

```php
$dispatcher->addSubscriber(new DatabaseSetupListener($app[Capsule::class]));
```

derrabus: Sure.

derrabus:
Done.

localheinz:
Thank you, @derrabus!

PR #888, opened 2017-12-07
Turned CSRF checker into an event subscriber

derrabus:
CSRF tokens are being checked in an event subscriber now.

This one is a bit tricky. Since Symfony does not offer the possibility to attach listeners to routes, I'm marking the affected routes with a request attribute.

As a auggestion: You should consider using Symfony forms because would gain CSRF protection out of the box.

Related to #618.


chartjes:
@derrabus We used to use Symfony forms but they got removed when we did a big UI overhaul. Probably wise to go back to using them again.

PR #889, opened 2017-12-07
Remove unused variable being passed to twig template

BackEndTea:
This PR

* [x] removes unneeded information passed to twig templates 

Only the home page uses the number_of_talks variable, so there was no need to pass it to the other templates.


localheinz:
Thank you, @BackEndTea!

BackEndTea:
Thank you, @localheinz

PR #890, opened 2017-12-07
Fix: Remove app reference in security controller

BackEndTea:
This PR

* [x] Removes a ```$this->app``` reference in the security controller

This caused a fatal error when trying to log in with wrong credentials


derrabus:
Ah, I must have missed that one.

BackEndTea:
No problem, this looks like the only ```$this->app[``` i can find in the project.

localheinz:

BackEndTea:
Thank you, @localheinz and @derrabus

PR #891, opened 2017-12-07
Fix: normalize twig provider

BackEndTea:
This PR

* [x] Makes the twig provider behave like other providers


localheinz:

BackEndTea:
Thank you @localheinz and @derrabus

PR #893, opened 2017-12-07
Construct the console application via DI

derrabus:
This PR moves the construction of the console application to the DI container. It also decouples the cache clear command from the `Path` object which simplifies some test cases.

### DIFF
--- a/bin/opencfp
+++ b/bin/opencfp
@@ -1,38 +1,24 @@
 #!/usr/bin/env php
 <?php
 
+namespace OpenCFP;
+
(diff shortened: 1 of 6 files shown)


chartjes:
@localheinz Don't want to step on your toes here, please check out the other changes and then approve/reject/comment at your leisure since I told CodeClimate we understand the register method is very long and full of returns

localheinz:
@chartjes 

No worries, feel free to take over, working on something different at the moment!

PR #896, opened 2017-12-07
Use a mockable authenticator instead of service swapping

derrabus:
Swapping services is really not fun with the Symfony container. With this PR, I tried to get around swapping services for authentication. WDYT?

Related to #618.

### DIFF
--- a/classes/Application.php
+++ b/classes/Application.php
@@ -27,6 +27,7 @@ use OpenCFP\Provider\TalkFilterProvider;
 use OpenCFP\Provider\TalkHandlerProvider;
 use OpenCFP\Provider\TalkHelperProvider;
 use OpenCFP\Provider\TalkRatingProvider;
+use OpenCFP\Provider\TestingServiceProvider;
 use OpenCFP\Provider\TwigServiceProvider;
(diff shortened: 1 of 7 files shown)


chartjes:
I *really* like this

mdwheele:
:heart: :heart_eyes: :heart:

classes/Infrastructure/Auth/MockableAuthenticator.php
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @see https://github.com/opencfp/opencfp
+ */
+
+namespace OpenCFP\Infrastructure\Auth;
+
+use OpenCFP\Domain\Services\Authentication;
+use OpenCFP\Domain\Services\NotAuthenticatedException;
+
+class MockableAuthenticator implements Authentication
```
BackEndTea: I really like this, but maybe it should be moved to the ```OpenCfp\Test\Helper``` namespace?

chartjes: 👍 to that suggestion

derrabus: All right, I will move it.

PR #899, opened 2017-12-08
Reduce twig priority

BackEndTea:
This PR

* [x] Reduces the priority of the twig global listener

Due to this having a priority of 512, the same as the ```DatabaseSetupListener``` ,it would fire before the db connection was set up and would throw a fatal error.

### DIFF
--- a/classes/Infrastructure/Event/TwigGlobalsListener.php
+++ b/classes/Infrastructure/Event/TwigGlobalsListener.php
@@ -59,7 +59,7 @@ class TwigGlobalsListener implements EventSubscriberInterface
     public static function getSubscribedEvents()
     {
         return [
-            KernelEvents::REQUEST => ['onKernelRequest', 512],
+            KernelEvents::REQUEST => ['onKernelRequest', 256],
         ];
     }
 


classes/Infrastructure/Event/TwigGlobalsListener.php
```diff
@@ -59,7 +59,7 @@ public function __construct(
     public static function getSubscribedEvents()
     {
         return [
-            KernelEvents::REQUEST => ['onKernelRequest', 512],
+            KernelEvents::REQUEST => ['onKernelRequest', -1024],
```
derrabus: This might be too low. If for instance the router stops the event propagation (because of a 404), the exception listener would not get the globals when rendering the error page.

BackEndTea: What would be a good priority for this listener?

derrabus: The router should be on 32, so anything between 32 and 512 should do. 256 maybe?

localheinz:

Issue #901, opened 2017-12-08
Swiftmailer 6


derrabus:
OpenCFP depends on Swiftmailer 5, but Symfony 4's Swiftmailer integration will require Swiftmailer 6. We should upgrade.

Prerequisite for #618.

chartjes:
I will work on a PR to update this

derrabus:
Fixed.

PR #902, opened 2017-12-08
Removing use of Silex provider for SwiftMailer

chartjes:
This PR

* [x] Updated OpenCFP to use SwiftMailer 6.0

Address #901


derrabus:
Oh, Silex's provider is incompatible with Swiftmailer 6? Good to know. 😕

chartjes:
@localheinz This is now ready for your review

classes/Domain/Services/TalkEmailer.php
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @see https://github.com/opencfp/opencfp
+ */
+
+namespace OpenCFP\Domain\Services;
+
+use Swift_Mailer;
+use Swift_Message;
+
+class TalkEmailer
```
localheinz: Can we mark this class as `final`? That is, unless we intend it not to be `final` so other people can extend it.

classes/Domain/Services/TalkEmailer.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    /**
+     * @var \Swift_Mailer
+     */
+    private $swiftMailer;
+
+    /**
+     * @param Swift_Mailer $swiftMailer
+     */
```
localheinz: Since the docblock doesn't add any value, can we remove it?

classes/Domain/Services/TalkEmailer.php
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * @return int
+     */
+    public function send(Swift_Message $message)
+    {
+        try {
+            return $this->swiftMailer->send($message);
+        } catch (\Exception $e) {
+            echo $e->getMessage();
+            die();
```
localheinz: Are we sure we just want to die here, rather than 

* letting the exception bubble up
* catching the exception and throwing a different exception (one that hides the underlying implementation)

?

classes/Domain/Services/TalkEmailer.php
localheinz: About the naming, so far I can't see anything here that indicates that the class has anything to do with talks, maybe we want to remove the `Talk` prefix for now?

classes/Domain/Services/TalkEmailer.php
```diff
@@ -0,0 +1,48 @@
...
+    public function __construct(Swift_Mailer $swiftMailer)
+    {
+        $this->swiftMailer = $swiftMailer;
+    }
+
+    /**
+     * @param Swift_Message $message
+     *
+     * @return int
+     */
```
localheinz: Can we remove the docblock and instead use a return type declaration?

classes/Provider/TalkEmailerServiceProvider.php
```diff
@@ -0,0 +1,43 @@
...
+use Pimple\Container;
+use Pimple\ServiceProviderInterface;
+use Swift_Mailer;
+use Swift_SmtpTransport;
+
+final class TalkEmailerServiceProvider implements ServiceProviderInterface
+{
+    /**
+     * {@inheritdoc}
+     */
```
localheinz: As docblocks are automatically inherited, can we remove this docblock?

classes/Provider/TalkEmailerServiceProvider.php
```diff
@@ -0,0 +1,43 @@
...
+        $options = [
+            'host'     => $app->config('mail.host'),
+            'port'     => $app->config('mail.port'),
+            'username' => $app->config('mail.username'),
+            'password' => $app->config('mail.password'),
+        ];
+        $transport = (new Swift_SmtpTransport($options['host'], $options['port']))
+            ->setUsername($options['username'])
+            ->setPassword($options['password']);
+        $swiftMailer         = new Swift_Mailer($transport);
```
localheinz: As of the moment, we create all of these even if we never use the service.

How about moving all of this into the closure, and thus simplifying this method to

```php
public function register(Container $app)
{
    $app['talk_emailer'] = function ($app) {
        /** @var Application $app */
        $transport = new Swift_SmtpTransport(
            $app->config('mail.host'),
            $app->config('mail.port')
        );

        $transport
            ->setUsername($app->config('mail.username'))
            ->setPassword($app->config('mail.password'));

        return new TalkEmailer(new Swift_Mailer($transport));
    };
}
```

?

classes/Provider/ResetEmailerServiceProvider.php
```diff
@@ -24,12 +26,23 @@
...
+        $options = [
+            'host'     => $app->config('mail.host'),
+            'port'     => $app->config('mail.port'),
+            'username' => $app->config('mail.username'),
+            'password' => $app->config('mail.password'),
+        ];
+        $transport = (new Swift_SmtpTransport($options['host'], $options['port']))
+            ->setUsername($options['username'])
+            ->setPassword($options['password']);
+        $swiftMailer          = new Swift_Mailer($transport);
```
localheinz: As of the moment, we create all of these even if we never use the service.

How about moving all of this into the closure, and thus simplifying this method to

```php
public function register(Container $app)
{
    $app['reset_emailer'] = function ($app) {
        /** @var Application $app */
        $transport = new Swift_SmtpTransport(
            $app->config('mail.host'),
            $app->config('mail.port')
        );

        $transport
            ->setUsername($app->config('mail.username'))
            ->setPassword($app->config('mail.password'));

        return new ResetEmailer(
            new Swift_Mailer($transport),
            $app['twig'],
            $app->config('application.email'),
            $app->config('application.title')
        );
    };
}
```

classes/Provider/ResetEmailerServiceProvider.php
```diff
@@ -24,12 +26,23 @@
...
+        $app['reset_emailer'] = function ($app) use ($options, $swiftMailer) {
             return new ResetEmailer(
-                $app['mailer'],
+                $swiftMailer,
                 $app['twig'],
                 $app->config('application.email'),
-                $app->config('application.title')
+                $app->config('application.title'),
+                $options
```
localheinz: As far as I can tell, the constructor of `ResetEmailer` doesn't need this argument, can we remove it?

tests/Unit/Domain/Services/ResetEmailerTest.php
```diff
@@ -50,12 +50,18 @@ public function it_sends_the_expected_email()
...
+            'username' => 'test',
+            'password' => 'test',
+        ];
         $resetEmailer = new ResetEmailer(
             $swiftMailer,
             $twig,
             'admin@example.com',
-            'Reset'
+            'Reset',
+            $options
```
localheinz: This argument is not needed, can we remove it?

composer.json
```diff
@@ -26,7 +26,7 @@
         "pimple/pimple": "3.*",
         "robmorgan/phinx": "*",
         "silex/silex": "2.*",
-        "swiftmailer/swiftmailer": "5.0",
+        "swiftmailer/swiftmailer": "^6.0",
```
localheinz: Looking at the diff suggests that this was edited manually, as unrelated dependencies (for example, `phpunit/phpunit`) have been updated.

Can we revert the changes to `composer.lock` and run

```
$ composer require swiftmailer/swiftmailer:^6.0.2 --update-with-dependencies
```

instead?

tests/Unit/Domain/Services/TalkEmailerTest.php
```diff
@@ -0,0 +1,42 @@
...
+final class TalkEmailerTest extends \PHPUnit\Framework\TestCase
+{
+    use MockeryPHPUnitIntegration;
+
+    /** @test */
+    public function it_sends_the_expected_email()
+    {
+        /* @var Swift_Mailer $swiftMailer */
+        $swiftMailer = Mockery::mock(Swift_Mailer::class);
+        $swiftMailer->shouldReceive('send')->andReturn(1);
```
localheinz: Can we also assert here that the method `send()` is invoked with the actual message?

tests/Unit/Domain/Services/TalkEmailerTest.php
```diff
@@ -0,0 +1,42 @@
...
+
+        // Create a message
+        $talkEmailer = new TalkEmailer($swiftMailer);
+        $response    = $talkEmailer->send(new Swift_Message());
+
+        $this->assertNotFalse($response);
```
localheinz: Can we instead assert that the return value is the same as what we set up our test double to return?

classes/Domain/Services/TalkEmailer.php
derrabus: This seems to be "best practise" when sending emails…

https://github.com/opencfp/opencfp/blob/fab87b36c7c5a950eb53d33d6dc042309edc30e0/classes/Domain/Services/ResetEmailer.php#L63-L70

I was already wondering about this when I came about that other piece of code. We should probably log the exception and display an error message to the user instead. Since there are more `die`s in the codebase, should we file a cleanup task for that?

classes/Provider/ResetEmailerServiceProvider.php
derrabus: Is there a reason why `ResetEmailer` and `TalkEmailer` use separate `Swift_Mailer` instances?

composer.json
derrabus: I wouldn't worry about that, tbh.

classes/Domain/Services/TalkEmailer.php
localheinz: @derrabus 👍

classes/Provider/ResetEmailerServiceProvider.php
localheinz: @derrabus 

Good question, I think they should share the same instance (unless there's a reason for different configuration or something).

classes/Domain/Services/TalkEmailer.php
derrabus: I don't really see the purpose of that class. It's a thin wrapper around the `Swift_Mailer` without any real logic. Why does the upgrade to Swiftmailer 6 make it necessary to introduce that class?

chartjes:
@localheinz Please take a look again at my changes

classes/Provider/Gateways/WebGatewayProvider.php
```diff
@@ -49,6 +51,20 @@
 {
     public function register(Container $app)
     {
+        $app['mailer'] = function ($app) {
+            $options = [
```
localheinz: What do we need `$options` for if we initialize it, and then access its values?

Can we skip the step and just access the values directly?

Can we just get rid of

classes/Provider/ResetEmailerServiceProvider.php
```diff
@@ -24,7 +24,13 @@
...
     public function register(Container $app)
     {
-        $app['reset_emailer'] = function ($app) {
+        $options = [
+            'host'     => $app->config('mail.host'),
+            'port'     => $app->config('mail.port'),
+            'username' => $app->config('mail.username'),
+            'password' => $app->config('mail.password'),
+        ];
+        $app['reset_emailer'] = function ($app) use ($options) {
```
localheinz: Can we remove `$options`? It's not used anywhere.

tests/Unit/Domain/Services/ResetEmailerTest.php
```diff
@@ -50,7 +50,12 @@ public function it_sends_the_expected_email()
         $twig->shouldReceive('loadTemplate')
             ->withArgs(['emails/reset_password.twig'])
             ->andReturn($template);
-
+        $options = [
```
localheinz: `$options` is not used anywhere, can we remove it?

classes/Provider/Gateways/WebGatewayProvider.php
```diff
@@ -49,6 +51,20 @@
 {
     public function register(Container $app)
     {
+        $app['mailer'] = function ($app) {
```
derrabus: I'm not sure if the `WebGatewayProvider` is the right spot for this service. It's not really web functionality.

chartjes: @derrabus I am open to a suggestion on a better place to do this -- I have no strong opinion and I do not believe we established a place for non-web-specific services to be registered.

Should this be a service provider that gets registered in our Application object? (say, call it SwiftmailServiceProvider)?

derrabus: Since we're replacing the original `SwiftmailerServiceProvider`, I'd go for naming our provider `SwiftmailerServiceProvider` as well.

Alternatively, we could rename `ResetEmailerServiceProvider` to `MailerServiceProvider` and move the Swiftmailer definition there, so everything related to sending emails is in one provider.

tests/Unit/ProjectCodeTest.php
```diff
@@ -77,6 +77,8 @@ public function testProductionClassesHaveUnitTests()
                 Provider\ImageProcessorProvider::class,
                 Provider\ResetEmailerServiceProvider::class,
                 Provider\SentinelServiceProvider::class,
+                Provider\SwiftMailerServiceProvider::class,
+                Provider\TalkEmailerServiceProvider::class,
```
derrabus: `TalkEmailerServiceProvider` is gone, isn't it?

chartjes: Good catch, fixed

localheinz:
Thank you, @chartjes and @derrabus!

PR #904, opened 2017-12-08
Fix: Assertion failure messages

localheinz:
This PR

* [x] fixes assertion failure messages

Follows #881.


localheinz:
Thank you, @mdwheele!

PR #903, opened 2017-12-08
Enhancement: Update friendsofphp/php-cs-fixer

localheinz:
This PR

* [x] updates `friendsofphp/php-cs-fixer`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/compare/v2.8.3...v2.9.0.

PR #884, opened 2017-12-07
Use test annotation

BackEndTea:
This PR

* [x] makes all tests use ```@test``` annotation
* [x] changes one line ```@test``` doc blocks into multi line doc blocks for consistency 

Related to #880

Issue #882, opened 2017-12-06
Are there any PHP-CS-Fixer checks we could be using to enforce CONTRIBUTING.md


mdwheele:
See: https://github.com/opencfp/opencfp/pull/880

- **All test methods must have the `@test` annotation** -- We value descriptiveness in test method names
- **All data providers must return array-of-args-arrays** -- The use of generators may let us not remember the number of arguments but too many arguments is a sign of potential code- and test-smells
- **All data provider methods must be prefixed `provider`**
- **All test method names should be `snake_case_test_methods`** 

I know @BackEndTea is working on something for the `@test` annotation in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3275. There were hints of something to enforce `snake_case` vs `camelCase` as a separate PR.

Do we have anything for data providers? Anybody want to volunteer to start a PR :grin:

BackEndTea:
Maybe i'm misunderstanding something, but wouldn't it make more sense to have dataproviders be suffixed with ```provider``` instead of prefixed?

```validRatingProvider()``` makes more sense then ```proivderForValidRatings()```

And issue number one will be resolved once my PR gets accepted. (Which i found a bug in when running it on this project code 😁)

I was planning on writing another fix after this one is all done, but i could take a look at camel to snake case, though there are about a million edge cases on that one, so i expect the fixer for that to take a while.

BackEndTea:
I created an issue surrounding camel and snake case in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3302 So we'll see if there is a desire for that one.

keradus:
on the other side, we use convension for providers naming:
test method: `testFooBar`,
provider: `provideFooBarCases`
(as objects shall be named after nouns and methods after verbs)

keradus:
I do believe that 1st, 3rd and 4th concepts could be done easily as fixers, 2nd could be challenging.
Yet, they have to be configured (like enforce one or other notation).

@BackEndTea is already done great work for 1st idea.
Would you consider picking up one of the rule and create a PR for it, @mdwheele  ?

keradus:
(oh boy, I thought it's PHP CS Fixer repo, how did I get here? sorry ;) )

mdwheele:
> And issue number one will be resolved once my PR gets accepted. (Which i found a bug in when running it on this project code )

[...]

PR #905, opened 2017-12-08
Enhancement: Enable and configure php_unit_test_annotation fixer

localheinz:
This PR

* [x] enables and configures the `php_unit_test_annotation` fixer
* [x] runs `make cs`

Follows #903.
Replaces #884.
Follows https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3312.
Follows #905.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/2.9#usage:

>**php_unit_test_annotation**
>
>Adds or removes `@test` annotations from tests, following configuration.
>
>*Risky rule: this fixer may change the name of your tests, and could cause incompatibility with abstract classes or interfaces.*
>
>Configuration options:
>
>* `case` (`'camel'`, `'snake'`): whether to camel or snake case when adding the test prefix; defaults to `'camel'`
>* `style` (`'annotation'`, `'prefix'`): whether to use the `@test` annotation or not; defaults to `'prefix'`

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -114,6 +114,9 @@ return PhpCsFixer\Config::create()
         'php_unit_expectation' => true,
         'php_unit_no_expectation_annotation' => true,
         'php_unit_strict' => true,
+        'php_unit_test_annotation' => [
+            'style' => 'annotation'
+        ],
         'php_unit_test_class_requires_covers' => false,
         'phpdoc_align' => true,
         'phpdoc_annotation_without_dot' => true,
--- a/tests/Integration/Domain/Model/AirportTest.php
+++ b/tests/Integration/Domain/Model/AirportTest.php
@@ -52,7 +52,10 @@ final class AirportTest extends WebTestCase
         $this->airports->withCode('foobarbaz');
     }
 
-    public function testItIsNotCaseSensitive()
+    /**
+     * @test
+     */
+    public function itIsNotCaseSensitive()
     {
         $airport = $this->airports->withCode('aac');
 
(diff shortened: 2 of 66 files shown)


keradus:
This post is to challenge annotation style, not PR itself.

With annotation style, how would one handle former `testGetName() {}` test ?
With annotation style, it would be
```php
/** @test */
public function getName() {}
```

which would crash on compilation level, as `getName` is in parent class already, yet it is not a test and has different prototype

mdwheele:
I wouldn't write a test for a getter unless there was some behavior behind that interface. Would you mind opening an issue to discuss? Further comments I would have are long-winded and would distract from this PR.

keradus:
`getName` is an example, you have a lot more methods in that base class you know? :)

For me it's exactly this PR concern, as this PR is aiming to enforce some rule, which (in my opinion) is problematic, as it could lead to conflicts with base class.
That's why I'm raising concern here, to avoid "happy merging" of style change, and then, in other issue, realize that given style could be problematic.

mdwheele:
> getName is an example, you have a lot more methods in that base class you know? :)

Sure! Haha. 

First, I expect this fixer (and this PR) to break a few things (for this project, in *this* particular case). Our preference isn't *just* use of annotation, but also a specific naming convention for test methods. In the case of `getName` (if I stick with this), we would actually have:

```php
/** @test */
public function get_name() {}

# Or...

/** @test */
public function it_gets_the_name_of_something() {}
```

Which should not conflict with any system under test or `TestCase`-inherited method name.

keradus:
OK, combining multiple "exotic" (for me) rules will indeed end up with safe setup :+1:

I must admit every time I see this snippet behat come to my mind, not phpunit ;)

localheinz:
@keradus @mdwheele 

For the record, as I pointed out in https://github.com/opencfp/opencfp/issues/882#issuecomment-349809362, I don't think that using the `@test` annotation adds any value (at all), but I'm fine using it when the projects dictates it, and there's a way of enforcing it.

keradus:
same here, so I would prefer to use name prefix instead,
but still using only one syntax and have it automatically checked provides a great value :+1:

localheinz:
Thank you, @BackEndTea!

PR #906, opened 2017-12-09
Fix: Suffix controller actions with Action

localheinz:
This PR

* [x] asserts that controller actions are suffixed with `Action`
* [x] renames controller actions


localheinz:
Thank you, @chartjes!

PR #907, opened 2017-12-11
Fix: Reduce visibility

localheinz:
This PR

* [x] reduces the visibility of methods


localheinz:
Thank you, @BackEndTea!

PR #908, opened 2017-12-12
Fix: Remove unused parameter

localheinz:
This PR

* [x] removes unused parameters

PR #909, opened 2017-12-12
Fix: Remove unused method

localheinz:
This PR

* [x] removes an unused method

Related to #908.

💁‍♂️ Running

```
$ git grep -i createdFormattedOutput
```

yields no results other than in tests, the interface, and the single implementation.


localheinz:
@mdwheele 

Rebased!

localheinz:
Thank you, @mdwheele!

PR #912, opened 2017-12-14
Enhancement: Turn DashboardController into action

localheinz:
This PR

* [x] turns the `DashboardController` into an action

💁‍♂️ If you like this, we can continue extracting actions from existing controllers.


BackEndTea:
Would the end goal be to have every controller action be its own class?

localheinz:
@BackEndTea 

> Would the end goal be to have every controller action be its own class?

Exactly!

This would have the following advantages

* since we inject dependencies through the constructor, we only need to inject dependencies that will be used by that specific action
* as it currently stands with the routing (I assume this will change when switching to Symfony, as I further assume that we will use YAML configuration), this will increase the visibility of actions in the routing definitions

BackEndTea:
In that case i'd like to see the actions that are currently together in a controller be in a folder together like so: 

📂  Actions
\ 📂 Admin
\  \ 📂 Speakers
\ \ 📄 IndexAction
\ \ 📄 ViewAction
etc.

What do you think?

localheinz:
@BackEndTea 

That's the plan!

chartjes:
I'm going to merge this to get the ball rolling and also approve of @BackEndTea's suggestion of how to organize the action (hopefully we will namespace them the same way)

localheinz:

Issue #894, opened 2017-12-07
RFC:  Changing where images are uploaded


chartjes:
In order to deploy to Heroku, we need to be able to change where our images are stored. https://devcenter.heroku.com/articles/s3#file-uploads talks about some potential solutions.

I tried abstracting out some stuff using Flysystem but I got stuck on not being able to create a backwards-compatible way of implementing it. Right now we assume all images uploaded will be served out of /uploads/<name of image> and we store the record in the database.

Open to suggestions on how to make the location configurable.

BackEndTea:
👍  for Flysystem. One problem is that currently there is not a single place where the uploads location is defined.

There is the ```Path::uploadPath()``` function, but it is never actually used outside of its tests. Currently some views use the twig ```uploads``` function, some just add ```/uploads/``` to it themselves, and some expect it to be already prefixed with ```/uploads/``` by the controller.

So a first step would be to normalise the uploading and downloading of images, and have 2 ways of dealing with images.

1. A reference to the place it is uploaded to
2. A reference to the place it is downloaded from ( for display)

When saving to the local filesystem, these would be the same. But if we are saving to google cloud buckets for example, we use their stream wrapper to upload the images to a bucket, but we use the static url to display them.

BackEndTea:
I made some first steps on solving this, but i think it might be smart to wait with implementing FlySystem until we fully made the change to Symfony, so we can use the [the FlySystem bundle](https://github.com/1up-lab/OneupFlysystemBundle).

PR #913, opened 2017-12-14
Enhancement: remove hardcoded references to uploads folder

BackEndTea:
This PR

* [x] Changes the Path and PathInterface to split up the uploads function 
* [x] Removes all references to 'uploads' in php classes and use the path function instead
* [x] Removes all references to 'uploads' in twig templates and uses the twig function instead

Related to #894.

This is a first step to making the uploading of profile pictures a bit more manageable on production environments where we can't, or don't want to, write to our own file system. If this gets merged, the only thing that needs to be changed in order to use another storage is the reference in the Path functions.

### DIFF
--- a/classes/Http/Controller/ProfileController.php
+++ b/classes/Http/Controller/ProfileController.php
@@ -46,16 +46,6 @@ class ProfileController extends BaseController
      */
     private $path;
 
-    /**
-     * ProfileController constructor.
-     *
-     * @param Authentication        $authentication
-     * @param HTMLPurifier          $purifier
-     * @param ProfileImageProcessor $profileImageProcessor
-     * @param Twig_Environment      $twig
-     * @param UrlGeneratorInterface $urlGenerator
-     * @param PathInterface         $path
-     */
     public function __construct(
         Authentication $authentication,
         HTMLPurifier $purifier,
--- a/classes/Http/Controller/ProfileController.php
+++ b/classes/Http/Controller/ProfileController.php
@@ -80,6 +80,7 @@ class ProfileController extends BaseController
             'speaker_info'   => $speakerData['info'],
             'speaker_bio'    => $speakerData['bio'],
             'speaker_photo'  => $speakerData['photo_path'],
+            //TODO: use Path downloadFrom function function
             'preview_photo'  => '/uploads/' . $speakerData['photo_path'],
             'airport'        => $speakerData['airport'],
             'transportation' => $speakerData['transportation'],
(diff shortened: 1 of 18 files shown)


classes/Http/Controller/ProfileController.php
```diff
@@ -40,16 +41,33 @@ class ProfileController extends BaseController
...
+    /**
+     * ProfileController constructor.
+     *
+     * @param Authentication        $authentication
+     * @param HTMLPurifier          $purifier
+     * @param ProfileImageProcessor $profileImageProcessor
+     * @param Twig_Environment      $twig
+     * @param UrlGeneratorInterface $urlGenerator
+     * @param PathInterface         $path
+     */
```
localheinz: As this docblock only repeats the arguments and their types, which - as far as I can tell - are already sufficiently declared, can we remove the docblock?

classes/Path.php
```diff
@@ -41,11 +41,16 @@ public function configPath(): string
         return $this->path . "/config/{$this->env}.yml";
     }
 
-    public function uploadPath(): string
+    public function uploadToPath(): string
     {
         return $this->path . '/web/uploads';
     }
 
+    public function downloadFromPath(): string
```
localheinz: I am not entirely sure about the naming.

As far as I understand, both `uploadToPath()` and `downloadFromPath()` refer to the same location, with the difference that the former represents the path in the file system, and the latter represents the path relative to the web root, is that right?

What do you think?

BackEndTea: The upload to path is the path in the filesystem, used to save the profile image.
The download from function is the path relative from the web root, correct.

I agree that the naming might be a bit confusing, so if you have any ideas i'd be glad to change them.

But if we take for example, the uploading of images using a google cloud bucket this makes a bit more sense.
We'd use the ```gs://``` streamwrapper to upload the images, since that is done on the server itself.
But when we are showing the images to the users, that is done from their browser, so link to the images from ```https://storage.googleapis.com/```.

localheinz: @BackEndTea 

About your example, I guess it would be best to abstract the *actual* location of an image away, that is, an image could be an object that has a URL, and then none of us have to care where that image is located, or whether it needs prefixing with anything.

Maybe it would be better to move the part that deals with the download path to the `SpeakerProfile` for now? What do you think?

BackEndTea: I wouldn't mind doing that, but a lot of places referring to the image location do not use the ```SpeakerProfile```.

localheinz:
Thank you, @BackEndTea!

BackEndTea:
Thanks @localheinz

PR #914, opened 2017-12-14
Enhancement: Enable and configure class_attributes_separation fixer

localheinz:
This PR

* [x] enables and configures the `class_attributes_separation` fixer
* [x] runs `make cs`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.9.0#usage:

>**class_attributes_separation** [`@Symfony`]
>
>Class, trait and interface elements must be separated with one blank line.
>
>Configuration options:
>
>* `elements` (`array`): list of classy elements; `'const'`, `'method'`, `'property'`; defaults to `['const', 'method', 'property']`

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -50,6 +50,12 @@ return PhpCsFixer\Config::create()
             ],
         ],
         'cast_spaces' => true,
+        'class_attributes_separation' => [
+            'elements' => [
+                'method',
+                'property',
+            ],
+        ],
         'combine_consecutive_unsets' => true,
         'concat_space' => [
             'spacing' => 'one',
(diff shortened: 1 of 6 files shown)


.php_cs.dist
```diff
@@ -50,6 +50,12 @@ return PhpCsFixer\Config::create()
             ],
         ],
         'cast_spaces' => true,
+        'class_attributes_separation' => [
+            'elements' => [
+                'method',
```
localheinz: Skipping separation of constants, which, however, we can enable, too.

localheinz:
Thank you, @BackEndTea!

PR #915, opened 2017-12-14
Fix: Add missing docblocks

localheinz:
This PR

* [x] adds missing docblocks

Spotted in #914.


localheinz:
Thank you, @BackEndTea!

PR #916, opened 2017-12-14
Fix: Clean up forms

localheinz:
This PR

* [x] cleans up forms

### DIFF
--- a/classes/Http/Form/Form.php
+++ b/classes/Http/Form/Form.php
@@ -18,22 +18,22 @@ abstract class Form
     /**
      * @var array
      */
-    protected $options;
+    private $options;
 
     /**
      * @var string[]
      */
-    protected $messages;
+    private $messages = [];
 
     /**
      * @var array
      */
-    protected $cleanData;
+    protected $cleanData = [];
 
     /**
      * @var array
      */
-    protected $taintedData;
+    protected $taintedData = [];
 
     /**
      * @var \HTMLPurifier
@@ -45,18 +45,10 @@ abstract class Form
      */
     protected $fieldList = [];
 
-    /**
-     * @param $data array of form data
-     * @param \HTMLPurifier $purifier
-     * @param $options
-     */
-    public function __construct($data, \HTMLPurifier $purifier, array $options = [])
+    public function __construct(array $data, \HTMLPurifier $purifier, array $options = [])
     {
-        $this->purifier    = $purifier;
-        $this->options     = $options;
-        $this->messages    = [];
-        $this->cleanData   = [];
-        $this->taintedData = [];
+        $this->purifier = $purifier;
+        $this->options  = $options;
 
         $this->populate($data);
     }
@@ -81,7 +73,7 @@ abstract class Form
     {
         // The $taintedData property might have been already set by
         // the populate() method.
-        $this->taintedData = \array_merge($this->taintedData, (array) $data);
+        $this->taintedData = \array_merge($this->taintedData, $data);
     }
 
     /**
@@ -132,7 +124,7 @@ abstract class Form
      *
      * @return null|mixed
      */
-    public function getTaintedField($name, $default = null)
+    public function getTaintedField(string $name, $default = null)
     {
         return $this->taintedData[$name] ?? $default;
     }
@@ -155,20 +147,24 @@ abstract class Form
      *
      * @return mixed The option value
      */
-    public function getOption($name, $default = null)
+    public function getOption(string $name, $default = null)
     {
         return $this->options[$name] ?? $default;
     }
 
     /**
      * Validates the form's submitted data.
+     *
+     * @param string $action
+     *
+     * @return bool
      */
-    abstract public function validateAll($action = 'create');
+    abstract public function validateAll(string $action = 'create'): bool;
 
     /**
      * Returns the list of error messages.
      *
-     * @return array
+     * @return string[]
      */
     public function getErrorMessages(): array
     {
@@ -215,14 +211,8 @@ abstract class Form
      */
     protected function internalSanitize(array $taintedData): array
     {
-        $purifier = $this->purifier;
-        $filtered = \array_map(
-            function ($field) use ($purifier) {
-                return $purifier->purify($field);
-            },
-            $taintedData
-        );
-
-        return $filtered;
+        return \array_map(function ($field) {
+            return $this->purifier->purify($field);
+        }, $taintedData);
     }
 }
--- a/classes/Http/Form/ResetForm.php
+++ b/classes/Http/Form/ResetForm.php
@@ -34,7 +34,7 @@ class ResetForm extends AbstractType
             ->getForm();
     }
 
-    public function getName()
+    public function getName(): string
     {
         return 'reset';
     }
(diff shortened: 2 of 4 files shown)


classes/Http/Form/Form.php
```diff
@@ -27,13 +27,10 @@
...
      * @param $options
      */
-    public function __construct($data, \HTMLPurifier $purifier, array $options = [])
+    public function __construct(array $data, \HTMLPurifier $purifier, array $options = [])
     {
-        $this->purifier    = $purifier;
-        $this->options     = $options;
-        $this->messages    = [];
-        $this->cleanData   = [];
-        $this->taintedData = [];
```
localheinz: No need to assign values in the constructor when we can assign default values to these fields.

classes/Http/Form/Form.php
```diff
@@ -16,24 +16,16 @@
...
+    protected $cleanData   = [];
+    protected $taintedData = [];
     protected $purifier;
     protected $fieldList = [];
 
-    /**
-     * @param $data array of form data
-     * @param \HTMLPurifier $purifier
-     * @param $options
-     */
```
localheinz: Docblock doesn't add any value.

classes/Http/Form/ResetForm.php
```diff
@@ -34,7 +34,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             ->getForm();
     }
 
-    public function getName()
+    public function getName(): string
```
localheinz: Not even sure whether this method is used by anything.

classes/Http/Form/SignupForm.php
```diff
@@ -35,14 +37,7 @@ class SignupForm extends Form
...
     ];
 
-    /**
-     * Validate all methods by calling all our validation methods
-     *
-     * @param string $action
-     *
-     * @return bool
-     */
-    public function validateAll($action = 'create')
```
localheinz: No need to repeat the docblock from the parent class.

classes/Http/Form/SignupForm.php
```diff
@@ -56,8 +51,6 @@ public function validateAll($action = 'create')
         $validEmail        = $this->validateEmail();
         $validFirstName    = $this->validateFirstName();
         $validLastName     = $this->validateLastName();
-        $validCompany      = $this->validateCompany();
-        $validTwitter      = $this->validateTwitter();
```
localheinz: Both of these methods return `true`, I think it's better to remove them rather than suggesting that some kind of validation actually occurs.

classes/Http/Form/SignupForm.php
```diff
@@ -88,22 +81,25 @@ public function validateSpeakerPhoto(): bool
             return true;
         }
 
+        /** @var HttpFoundation\File\UploadedFile $speakerPhoto */
+        $speakerPhoto = $this->taintedData['speaker_photo'];
```
localheinz: Introducing an explaining variable here so we can add a docblock (auto-completion).

classes/Http/Form/SignupForm.php
```diff
@@ -112,13 +108,6 @@ public function validateSpeakerPhoto(): bool
         return true;
     }
 
-    /**
-     * Method that applies validation rules to email
-     *
-     * @return bool
-     *
-     * @internal param string $email
-     */
```
localheinz: I believe all of these docblocks aren't actually necessary.

classes/Http/Form/SignupForm.php
```diff
@@ -112,13 +108,6 @@ public function validateSpeakerPhoto(): bool
...
     public function validateEmail(): bool
```
localheinz: All of these validation methods currently have `public` visibility, however, they only currently need to be `public` because they are directly invoked by tests. I suggest to reduce the visibility to `private` and modify the tests (in a separate PR).

classes/Http/Form/TalkForm.php
```diff
@@ -171,24 +145,4 @@ public function validateCategory(): bool
 
         return true;
     }
-
-    public function validateDesired(): bool
```
localheinz: See above comment about removing methods that suggest validation, but actually don't validate anything at all.

classes/Http/Form/Form.php
```diff
@@ -215,14 +211,8 @@ public function sanitize()
      */
     protected function internalSanitize(array $taintedData): array
     {
-        $purifier = $this->purifier;
-        $filtered = \array_map(
-            function ($field) use ($purifier) {
```
localheinz: No need to import anything here.

classes/Http/Form/Form.php
```diff
@@ -215,14 +211,8 @@ public function sanitize()
      */
     protected function internalSanitize(array $taintedData): array
     {
-        $purifier = $this->purifier;
-        $filtered = \array_map(
```
localheinz: No need for a temporary variable here.

classes/Http/Form/SignupForm.php
```diff
@@ -71,8 +64,7 @@ public function validateAll($action = 'create')
             $validSpeakerBio = $this->validateSpeakerBio();
         }
 
-        return
-            $validEmail && $validPasswords && $validFirstName && $validLastName && $validCompany && $validTwitter && $validUrl && $validSpeakerInfo && $validSpeakerBio && $validSpeakerPhoto && $agreeCoc;
+        return $validEmail && $validPasswords && $validFirstName && $validLastName && $validUrl && $validSpeakerInfo && $validSpeakerBio && $validSpeakerPhoto && $agreeCoc;
```
BackEndTea: Here we use variables and return whether all those are true, but in the ```TalkFrom``` we don't use temporary variables. 
It might be nicer to make both work in the same way, what do you think?

localheinz: In which direction

* [ ]  inline variables
* [ ] introduce variables

?

BackEndTea: I don't have a preference for one or the other really, which do you think looks better?

classes/Http/Form/ResetForm.php
BackEndTea: I can't find any place that uses it. Looks safe to delete.

classes/Http/Form/SignupForm.php
```diff
@@ -35,44 +37,31 @@ class SignupForm extends Form
...
-        $validCompany      = $this->validateCompany();
-        $validTwitter      = $this->validateTwitter();
-        $validUrl          = $this->validateUrl();
-        $validSpeakerPhoto = $this->validateSpeakerPhoto();
-        $validSpeakerInfo  = true;
-        $validSpeakerBio   = true;
+        $validSpeakerInfo = true;
 
         if (!empty($this->taintedData['speaker_info'])) {
             $validSpeakerInfo = $this->validateSpeakerInfo();
```
BackEndTea: What if we move the check to see if it is empty to the ```validateSpeakerInfo()``` function?
And the same for the speaker bio?

localheinz:
Thank you, @BackEndTea!

classes/Http/Form/SignupForm.php
localheinz: 👍 for cleaning up further!

PR #917, opened 2017-12-14
Enhancement: Test coverage and infection testing

BackEndTea:
This PR

* [x] Removes the php_unit_test_class_requires_covers rule
* [x] Removes the ```@coversNothing``` annotations
* [x] Moves back to using one phpunit.xml, and split up tests by testsuite. 
* [x] Changes references to the old config files to use the testsuites instead

This makes a few things possible.

* The mutation tests now take our integration tests into account, which means we can make PR's require a certain mutation 'score': https://infection.github.io/guide/using-with-ci.html
* To run a specific integration test, you can once again run ```vendor/bin/phpunit <path to tests>```, without having to specify a configuration to use.

I left running the infection tests on travis out of this PR. One of the downsides is that it takes quite a while to run the full suite. I suggest that before running this on travis we 
* fix some of the **really** slow tests
* decide what mutators we are interested in
* decide on minimum msi

For more information on infection, see: https://infection.github.io/guide/

Our current metrics are:
```
714 mutations were generated:
     476 mutants were killed
     106 mutants were not covered by tests
      61 covered mutants were not detected
      63 errors were encountered
       8 time outs were encountered

Metrics:
     Mutation Score Indicator (MSI): 77%
     Mutation Code Coverage: 85%
     Covered Code MSI: 90%
```


chartjes:
I don't think that we should be getting Travis to run the mutation tests as part of the normal build. I feel like the mutation tests belong in the same category of using static analyzers -- good tools for the developers to find things behind the scenes to fix.

I'll fix the conflict and merge. Thanks folks

BackEndTea:
Thank you @chartjes

Issue #774, opened 2017-11-29
RFC: Better e2e testing


BackEndTea:
Right now we use our controller integration tests as e2e tests, but they widely differ in what they actually test. Some mock out just about everything and try to be it a unit test, while others mock next to nothing.

We also run into the issue that while PHPUnit is great for unit tests, its maybe not the best choice for e2e testing. I think we may greatly benefit from something like [behat](http://behat.org/en/latest/) to do these kind of big tests.

BackEndTea:
Some options that i found

* [Codeception](http://codeception.com/docs/03-AcceptanceTests)
* [Behat](http://behat.org/en/latest/)
* [StoryPlayer](http://datasift.github.io/storyplayer/)

mdwheele:
We used to have a Behat suite (that I used originally to give me some stability while making large changes). I would prefer it to Codeception.

TBH, I think we go test crazy on controllers and a good, solid and readable e2e test would better express the behaviours we're after. The fact that we have like 90km of mock for 1m of test is a signal I think.

I would personally be very-much behind a Behat-based e2e suite as long as it described *behaviour*, not over-specification / over-use of "browser-based clicky clicky". IMO, that "clicky clicky" belongs in the Context, while Gherkin should clearly describe behaviour that happens to implemented via web interface (or an `ApiContext`). Same behaviour (Gherkin) in multiple contexts (`WebContext`, `ApiContext`)

chartjes:
Closing because we've started work on the Behat test suite

PR #919, opened 2017-12-14
Updated travis configuration to now just call phpunit directly without

chartjes:
any test suite information

This PR

* [x] updates the Travis configuration file to now call phpunit directly

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,9 +70,8 @@ jobs:
         - if [[ "$WITH_COVERAGE" == "true" ]]; then code-climate-before-script; fi
 
       script:
-        - vendor/bin/phpunit --testsuite integration
         - if [[ "$WITH_COVERAGE" == "true" ]]; then xdebug-enable; fi
-        - if [[ "$WITH_COVERAGE" == "true" ]]; then vendor/bin/phpunit --testsuite unit --coverage-clover build/logs/clover.xml; else vendor/bin/phpunit --testsuite unit; fi
+        - if [[ "$WITH_COVERAGE" == "true" ]]; then vendor/bin/phpunit --coverage-clover build/logs/clover.xml; else vendor/bin/phpunit; fi
         - if [[ "$WITH_COVERAGE" == "true" ]]; then xdebug-disable; fi
 
       after_script:


.travis.yml
```diff
@@ -70,9 +70,8 @@ jobs:
         - if [[ "$WITH_COVERAGE" == "true" ]]; then code-climate-before-script; fi
 
       script:
-        - vendor/bin/phpunit --testsuite integration
```
localheinz: No, we don't want to collect coverage for integration tests, do we?

That was the point of #629.

BackEndTea:
Wasnt the idea of spliting up the tests so we could collect only the coverage of the unit tets?

chartjes:
I want accurate code coverage reports because I feel like we are ignoring what Code Climate tells us now. It does not bother me that the integration tests are included in coverage. It bothers me that our stats say 34% coverage when that is not accurate at all.

BackEndTea:
Im okay with the change. Id suggest to add covers  annotation to the integration tests. To at least show what they are testing

localheinz:
@chartjes 

As far as I'm concerned, the problems with collecting code coverage for integration tests are these:

* it encourages to write untestable code (no need to inject and mock dependencies)
* integration tests are slow, and when we collect coverage for integration tests, it is even slower
* there will be code that is *accidentally* covered, which gives a false impression on what is actually tested

Because of these reasons I would not collect coverage for integration tests.

chartjes:
So how do we reconcile this with the code coverage metric Code Climate produces? It will show an artificially low value.

—
Chris

localheinz:
@chartjes 

I guess it boils down to whether we want to take shortcuts in achieving higher code coverage, or if we don't. 

I'm convinced that most of the integration tests can be turned into unit tests, except for those where we don't inject dependencies, but access methods of Eloquent models statically (and these should probably be modified, so that we can properly test them.

mdwheele:
I think the purpose of integration tests should be... to test the integration of our domain to some specific infrastructure outside of our domain, be that a database, an external API, etc. 

Integration tests should not be "mock theater" either. They exist to test *actual* integration with *actual* infrastructure. How will I know if I am covered if I do not collect coverage? That said, @BackEndTea's suggestion to specifically target coverage is ideal, from my perspective.

I make these statement outside the context of what our Integration suite *currently* is, but what I think it should be directed towards. I don't think @chartjes is suggesting shortcuts to achieve a better grade from CodeClimate.

localheinz:
@mdwheele 

I'm mostly concerned with cycle time. 

```
$ vendor/bin/phpunit --testsuite integration

PHPUnit 6.5.3 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.0.25
Configuration: /Users/am/Sites/opencfp/opencfp/phpunit.xml.dist

.........................................................................................................................................................................                                                                                                                                     169 / 169 (100%)

Time: 18.8 seconds, Memory: 316.00MB

OK (169 tests, 1036 assertions)
```

vs

```
$ vendor/bin/phpunit --testsuite unit
PHPUnit 6.5.3 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.0.25
Configuration: /Users/am/Sites/opencfp/opencfp/phpunit.xml.dist

............................................................................................................................................................................................................................................................................................................. 301 / 327 ( 92%)
..........................                                                                                                                                                                                                                                                                                    327 / 327 (100%)

Time: 715 ms, Memory: 66.00MB

OK (327 tests, 728 assertions)
```

Running integration tests is clearly **a lot** slower than running unit tests, so personally, I prefer to have unit tests rather than integration tests. Especially I'd like to reduce tests that interact with the database. We currently have twice as many unit tests as we have integration tests, yet they run in less than 1 second, while the integration tests take 20 seconds. That is too much time to wait for, to be honest.

localheinz:
@mdwheele 

>Somewhat related (but not to coverage), I think it's silly to attempt to unit test Active Record models with regard to persistence.

I would suggest not to use *Active Record* models in the first place, but that is a different story.

BackEndTea:
@localheinz 
> * it encourages to write untestable code (no need to inject and mock dependencies)

I agree that we should move to make sure our code stays testable
> * integration tests are slow, and when we collect coverage for integration tests, it is even slower

It is true they are quite a bit slower, but thats mostly due to the integration tests that actually talk to the database. Personally, i'm willing to have our test suite be a bit slower, if it gives me that security that all the components are working nicely together.

We can have a thousand passing unit tests and 100% coverage, but an application that is completely broken. So i prefer to test that me telling sentinel to activate an account actually activates it, and check that it has properly done so.

> * there will be code that is accidentally covered, which gives a false impression on what is actually tested

I'm making a PR that adds ```@covers``` annotations to the integration tests, so we know what they are supposed to test.

Now i'm still in favor of #774 and use behat for those e2e tests that our controllers currently are.

chartjes:
To be 100% clear, I am not suggesting any types of short cuts. Yes, we should have more unit tests and fewer integration tests AND we should only count coverage when it comes to unit tests BUT the tool we are using to report on our progress in that department currently does not allow us to make that fine-grained distinction. Instead it reports a number that is much lower than our overall test coverage when you include the integration tests.

I don't like seeing that number. It projects an image of the stabilty of the application that is just wrong.

I'm closing this PR because I feel this is an issue that needs to be hammered out before any PR about is created. I could just force it through, but I won't.

localheinz:

There's another reason why I for unit tests over integration tests (not saying that there's no reason to write integration tests): A failing integration test easily tells us that something failed, but not where. With a unit test it's a lot easier to find the cause of a problem.

For example, see the proposed change in #921, and here's an unrelated test failing:

```
vendor/bin/phpunit --testsuite integration
PHPUnit 6.5.3 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.0.25
Configuration: /Users/am/Sites/opencfp/opencfp/phpunit.xml.dist

..............................................E.............................................................................................................................                                                                                                                                  172 / 172 (100%)

Time: 18.82 seconds, Memory: 316.00MB

There was 1 error:

1) OpenCFP\Test\Integration\Http\Action\DashboardActionTest::indexDisplaysUserAndTalks
TypeError: Argument 1 passed to OpenCFP\Http\View\TalkHelper::getTypeDisplayName() must be of the type string, null given, called in /Users/am/Sites/opencfp/opencfp/vendor/twig/twig/lib/Twig/Template.php on line 677

/Users/am/Sites/opencfp/opencfp/classes/Http/View/TalkHelper.php:70
/Users/am/Sites/opencfp/opencfp/vendor/twig/twig/lib/Twig/Template.php:677
/Users/am/Sites/opencfp/opencfp/vendor/twig/twig/lib/Twig/Template.php:215
/Users/am/Sites/opencfp/opencfp/vendor/twig/twig/lib/Twig/Template.php:432
/Users/am/Sites/opencfp/opencfp/vendor/twig/twig/lib/Twig/Template.php:403
/Users/am/Sites/opencfp/opencfp/vendor/twig/twig/lib/Twig/Template.php:432
/Users/am/Sites/opencfp/opencfp/vendor/twig/twig/lib/Twig/Template.php:403
/Users/am/Sites/opencfp/opencfp/vendor/twig/twig/lib/Twig/Template.php:411
/Users/am/Sites/opencfp/opencfp/vendor/twig/twig/lib/Twig/Environment.php:363
/Users/am/Sites/opencfp/opencfp/classes/Http/Action/DashboardAction.php:53
/Users/am/Sites/opencfp/opencfp/vendor/symfony/http-kernel/HttpKernel.php:153
/Users/am/Sites/opencfp/opencfp/vendor/symfony/http-kernel/HttpKernel.php:68
/Users/am/Sites/opencfp/opencfp/vendor/silex/silex/src/Silex/Application.php:496
/Users/am/Sites/opencfp/opencfp/tests/Integration/WebTestCase.php:116
/Users/am/Sites/opencfp/opencfp/tests/Integration/WebTestCase.php:124
/Users/am/Sites/opencfp/opencfp/tests/Integration/Http/Action/DashboardActionTest.php:52

ERRORS!
Tests: 172, Assertions: 1013, Errors: 1.
make: *** [integration] Error 2
```

This indicates that there's a starting point, but who knows how many files one needs to look through to find the cause of the error? With a unit test it's a lot easier to find and fix issues.

BackEndTea:
Except if it was a unit test the call to that function could have been mocked out, and there would be no error.
There is also the fact that we use templates, which need to be fed specific information, which in turn, could cause an error.

localheinz:
One more thing: code coverage is a useless metric anyway. If one follows TDD, it’s always 100%. If it drops, it just means code was written before tests. 

That said, I don’t care much how we collect it. 

It’s only that I prefer unit over integration tests in regard to speed, and in regard to the resulting code.

mdwheele:
> I would suggest not to use Active Record models in the first place, but that is a different story.

Everything has a time and a place. There is no end-all be-all solution. I prefer to abstain from surface-level "I hate ____ pattern or practice."

> One more thing: code coverage is a useless metric anyway. If one follows TDD, it’s always 100%. If it drops, it just means code was written before tests.

That's a bit too absolute / blunt-force, I think. This project provides certain value to a conference organizer running a CFP. However, one of the biggest reasons I keep contributing here is that I have pointed no less than 30 entry-level developers to this project to demonstrate that chasing perfection is less important than providing real value. The fact that our tests *aren't* perfect, but we're even having this conversation, is an implicit value to the surrounding community. I have more than a 100 examples of positive interactions I can use from this project as exemplars of "how to work in open-source".

I don't think code coverage is useless and I think shooting for 100% sets a bar that newcomers cannot achieve. There is nothing absolute about development. It's grey areas and trade-offs all the way down with sparks of human interaction. Obviously, that isn't quite aligned with "provide a stable solution for running a CFP", but I feel there is value there anyway.

> A failing integration test easily tells us that something failed, but not where. With a unit test it's a lot easier to find the cause of a problem.

If `SentinelAuthentication` fails, I know it's because a dependency broke. If I mock everything, I'm basically implementing the `Sentinel@x.x.x` contract as mocks. If there is ever any question of where the break occurred in an integration, that is an opportunity to improve the test, I think. Make the test more intelligent to know "Hey! You didn't migrate your DB so *of course* this integration will fail." However, I don't think that's a signal that a unit test would have inherently provided a better solution.

---

I think I would have preferred merging this and following up with specifying `@covers` for each integration as @BackEndTea has suggested. If we truly were worried about "integration tests masquerading as unit tests", then identifying those, creating an issue, and delegating work to clean up would probably work just fine. Then we could approach each one case-by-case and make the right calls per-PR.

I think we're all on the same page as far as the "testing triangle" Number of unit > integration > e2e. This PR was about coverage, not suite run-time. I think we want *true* integrations covered. I think we should do that and follow-up by identifying poor integration tests and deciding "Is this actually better as a Unit test? Or is this an *actual* integration and we need to hit database / sessions to test it."

PR #923, opened 2017-12-14
Fix: Adjust covers annotation

localheinz:
This PR

* [x] adjusts a `@covers` annotation

Follows #912.


localheinz:
Thank you, @chartjes!

PR #924, opened 2017-12-15
Enhancement: Extract ViewAction from TalkController

localheinz:
This PR

* [x] extracts a `ViewAction` from the `TalkController`

Follows #912.


localheinz:
Thank you, @BackEndTea!

PR #925, opened 2017-12-15
Enhancement: Extract EditAction from TalkController

localheinz:
This PR

* [x] extracts an `EditAction` from the `TalkController`

Follows #912.


classes/Http/Action/Talk/EditAction.php
```diff
@@ -0,0 +1,119 @@
...
+
+        if (empty($talkId)) {
+            $url = $this->urlGenerator->generate('dashboard');
+
+            return new HttpFoundation\RedirectResponse($url);
+        }
+
+        $userId = $this->authentication->user()->getId();
+
+        $talk = Model\Talk::find($talkId);
```
localheinz: Because of this static access, we can only write unit tests up until here.

localheinz:
Thank you, @chartjes!

PR #926, opened 2017-12-15
Enhancement: Extract Pages Actions

BackEndTea:
This PR

* [x] Extracts all the actions from the pages controller into their own actions

Follows #912 

Since all three actions are rather small i put them into one PR.

PR #930, opened 2017-12-15
Fix: Assert identity

localheinz:
This PR

* [x] uses `assertSame()` instead of `assertContains()`

Follows #926.


localheinz:
Thank you, @BackEndTea!

PR #932, opened 2017-12-15
Don't swap the CallForPapers service

derrabus:
This PR uses reflection calls in integration tests to modify the `CallForPapers` service instead of swapping the whole service with a mocked one.

Related to #618.


chartjes:
Thanks @derrabus

PR #933, opened 2017-12-15
Render flash message on reset form and validate the form.

derrabus:
This PR adds validation to the reset form and makes sure that flash messages are displayed on that form.

Unfortunately, `ResetControllerTest` does not catch either of those issues because the form validation is mocked and the flash bag is accessed directly. I'm currently checking how to improve tests like that one.


localheinz:
Thank you, @derrabus!

PR #934, opened 2017-12-15
Use Silex' test client for integration tests

derrabus:
This PR integrates Silex' test client into the integration test suite. This allows us to emulate user interaction better and with less mocking.

The PR also contains a better `ResetControllerTest` that covers the issue fixed by PR #933.

Follows #933.
Related to #618.

### DIFF
--- a/classes/Infrastructure/Event/TwigGlobalsListener.php
+++ b/classes/Infrastructure/Event/TwigGlobalsListener.php
@@ -59,7 +59,7 @@ class TwigGlobalsListener implements EventSubscriberInterface
     public static function getSubscribedEvents()
     {
         return [
-            KernelEvents::REQUEST => ['onKernelRequest', 256],
+            KernelEvents::REQUEST => ['onKernelRequest', 128],
         ];
     }
 
--- a/tests/BaseTestCase.php
+++ b/tests/BaseTestCase.php
@@ -21,8 +21,9 @@ use OpenCFP\Test\Helper\DataBaseInteraction;
 use OpenCFP\Test\Helper\RefreshDatabase;
 use Pimple\Psr11\Container;
 use Psr\Container\ContainerInterface;
+use Silex\WebTestCase;
 
-abstract class BaseTestCase extends \PHPUnit\Framework\TestCase
+abstract class BaseTestCase extends WebTestCase
 {
     use Helper;
     use MockeryPHPUnitIntegration;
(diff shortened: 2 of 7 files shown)


tests/BaseTestCase.php
```diff
@@ -21,8 +21,9 @@
 use OpenCFP\Test\Helper\RefreshDatabase;
 use Pimple\Psr11\Container;
 use Psr\Container\ContainerInterface;
+use Silex\WebTestCase;
 
-abstract class BaseTestCase extends \PHPUnit\Framework\TestCase
+abstract class BaseTestCase extends WebTestCase
```
localheinz: @derrabus 

What do you think about getting rid of the `BaseTestCase` entirely?

Currently, we have `OpenCFP\Test\Integration\WebTestCase` which extends from `OpenCFP\Test\BaseTestCase`. If we now have the `BaseTestCase` extend the `Silex\WebTestCase`, I don't really see any use for the `BaseTestCase` at all.

What do you think?

tests/Integration/Http/Controller/ForgotControllerTest.php
```diff
@@ -51,26 +51,30 @@ public function indexDisplaysCorrectForm()
     public function sendResetDisplaysCorrectMessage()
     {
         $accounts = m::mock(AccountManagement::class);
-        $accounts->shouldReceive('findByLogin')->andReturn($this->createUser());
+        $accounts->shouldReceive('findByLogin')
+            ->withArgs(['someone@example.com'])
```
localheinz: 👍 

I believe quite a few tests where we currently set up expectations could benefit from being more specific.

derrabus:
@localheinz Have a look at #933. That PR only contains the fix.

localheinz:
Thank you, @derrabus, merged #993!

Happy to merge this one, once rebased, and then I would follow up with #936.

tests/BaseTestCase.php
derrabus: If that means that all unit tests will extend PHPUnit's test case class directly, go for it. If that means, some unit tests are switched to `WebTestCase`, I'm not really sure. That class contains a toolset that is not meant for unit tests.

derrabus:
Rebased.

classes/Infrastructure/Event/TwigGlobalsListener.php
```diff
@@ -59,7 +59,7 @@ public function __construct(
     public static function getSubscribedEvents()
     {
         return [
-            KernelEvents::REQUEST => ['onKernelRequest', 256],
+            KernelEvents::REQUEST => ['onKernelRequest', 128],
```
localheinz: Not sure about changing the priorities here, but I trust you on this one, @derrabus.

derrabus: The priority had to be lowered because this listener would otherwise start a session in integration tests before Silex' test session listener is called.

tests/BaseTestCase.php
localheinz: > If that means that all unit tests will extend PHPUnit's test case class directly, go for it. 

👍

localheinz:
Thank you, @derrabus!

classes/Infrastructure/Event/TwigGlobalsListener.php
localheinz: Thank you for clarifying, @derrabus!

PR #935, opened 2017-12-15
Enhancement: Extract DeleteAction from TalkController

localheinz:
This PR

* [x] extracts a `DeleteAction` from the `TalkController`

Follows #912.

### DIFF
--- a/classes/Http/Action/Talk/DeleteAction.php
+++ b/classes/Http/Action/Talk/DeleteAction.php
@@ -1,0 +1,66 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Copyright (c) 2013-2017 OpenCFP
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ *
+ * @see https://github.com/opencfp/opencfp
+ */
+
+namespace OpenCFP\Http\Action\Talk;
+
+use OpenCFP\Domain\CallForPapers;
+use OpenCFP\Domain\Model;
+use OpenCFP\Domain\Services;
+use Symfony\Component\HttpFoundation;
+
+final class DeleteAction
+{
+    /**
+     * @var Services\Authentication
+     */
+    private $authentication;
+
+    /**
+     * @var CallForPapers
+     */
+    private $callForPapers;
+
+    public function __construct(Services\Authentication $authentication, CallForPapers $callForPapers)
+    {
+        $this->authentication = $authentication;
+        $this->callForPapers  = $callForPapers;
+    }
+
+    public function __invoke(HttpFoundation\Request $request): HttpFoundation\Response
+    {
+        if (!$this->callForPapers->isOpen()) {
+            return new HttpFoundation\JsonResponse([
+                'delete' => 'no',
+            ]);
+        }
+
+        $talkId = $request->get('tid');
+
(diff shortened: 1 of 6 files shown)


classes/Http/Action/Talk/DeleteAction.php
```diff
@@ -0,0 +1,66 @@
...
+                'delete' => 'no',
+            ]);
+        }
+
+        $talkId = $request->get('tid');
+
+        $userId = $this->authentication->user()->getId();
+
+        /** @var Model\Talk $talk */
+        $talk = Model\Talk::find($talkId, ['id', 'user_id']);
```
localheinz: Because of static access here we can't test any further than the branch above.

classes/Http/Action/Talk/DeleteAction.php
```diff
@@ -0,0 +1,66 @@
...
+
+        if ($talk->user_id != $userId) {
```
localheinz: This could potentially go wrong if we haven't found a talk for the given identifier.

classes/Http/Action/Talk/DeleteAction.php
```diff
@@ -0,0 +1,66 @@
...
+    public function __construct(Services\Authentication $authentication, CallForPapers $callForPapers)
+    {
+        $this->authentication = $authentication;
+        $this->callForPapers  = $callForPapers;
+    }
+
+    public function __invoke(HttpFoundation\Request $request): HttpFoundation\Response
+    {
+        if (!$this->callForPapers->isOpen()) {
+            return new HttpFoundation\JsonResponse([
```
localheinz: We should probably also set an appropriate status code here.

classes/Http/Action/Talk/DeleteAction.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $talkId = $request->get('tid');
+
+        $userId = $this->authentication->user()->getId();
+
+        /** @var Model\Talk $talk */
+        $talk = Model\Talk::find($talkId, ['id', 'user_id']);
+
+        if ($talk->user_id != $userId) {
+            return new HttpFoundation\JsonResponse([
```
localheinz: We should probably also set an appropriate status code here, or even better, only set an appropriate status code.

classes/Http/Action/Talk/DeleteAction.php
```diff
@@ -0,0 +1,66 @@
...
+                'delete' => 'no',
+            ]);
+        }
+
+        $talk->delete();
+
+        return new HttpFoundation\JsonResponse([
```
localheinz: We should probably also set an appropriate status code here, or even better, only set an appropriate status code.

classes/Http/Action/Talk/DeleteAction.php
mdwheele: Can either test for not-null (`if ($talk && $talk->user_id != $userId)`) or use `Model\Talk::findOrFail`, which will toss an `Illuminate\Database\Eloquent\ModelNotFoundException` if a talk for given identifier was not found. Do you want to fix that as part of this PR?

localheinz: I’d prefer to do this in a separate PR.

localheinz:
Thank you, @mdwheele!

PR #936, opened 2017-12-15
Fix: Remove BaseTestCase

localheinz:
This PR

* [x] removes the `BaseTestCase`

Follows  #934.
Follows https://github.com/opencfp/opencfp/pull/934/files#r157262574.

PR #937, opened 2017-12-15
Fix: Move integration test into appropriate namespace

localheinz:
This PR

* [x] moves an integration test into the appropriate namespace

Blocks #936.

/cc @derrabus

### DIFF
--- a/tests/Integration/Provider/SentinelServiceProviderTest.php
+++ b/tests/Integration/Provider/SentinelServiceProviderTest.php
@@ -11,14 +11,11 @@ declare(strict_types=1);
  * @see https://github.com/opencfp/opencfp
  */
 
-namespace OpenCFP\Test\Unit\Provider;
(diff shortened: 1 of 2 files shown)


tests/Integration/Provider/SentinelServiceProviderTest.php
```diff
@@ -11,14 +11,11 @@
...
 
 use Cartalyst\Sentinel\Sentinel;
 use OpenCFP\Test\BaseTestCase;
 
-/**
- * @covers \OpenCFP\Provider\SentinelServiceProvider
- */
 final class SentinelServiceProviderTest extends BaseTestCase
 {
     public function testAllRepositoriesAreSet()
```
localheinz: Since we access the application container in this test, I believe this should be an integration test.

localheinz:
Thank you, @BackEndTea!

PR #938, opened 2017-12-15
Fix: Extend from PHPUnit\Framework\TestCase

localheinz:
This PR

* [x] adjusts a unit tests that unnecessarily extends from `BaseTestCase`

Blocks #936.

/cc @derrabus


localheinz:

PR #939, opened 2017-12-15
Fix: Mark methods as final to prevent multiple imports

localheinz:
This PR

* [x] marks methods provided by test helper traits as `final`

Related to #936.

💁‍♂️ This is useful to prevent multiple imports, as well as tests attempting to override these methods.


localheinz:
Thank you, @BackEndTea!

PR #940, opened 2017-12-15
Fix: Assert redirect url

localheinz:
This PR

* [x] asserts that URLs have been set as redirect URLs

Follows #912.
Follows #924.


localheinz:
Thank you, @mdwheele!

PR #941, opened 2017-12-15
Fix: Assert that flash is not null

localheinz:
This PR

* [x] asserts that flash pulled from session is not `null`

### DIFF
--- a/tests/Helper/ResponseHelper.php
+++ b/tests/Helper/ResponseHelper.php
@@ -149,7 +149,10 @@ trait ResponseHelper
 
     final protected function assertSessionHasFlashMessage(string $message, HttpFoundation\Session\SessionInterface $session)
     {
-        $this->assertContains($message, $session->get('flash'), \sprintf(
+        $flash = $session->get('flash');
+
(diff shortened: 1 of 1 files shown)


tests/Helper/ResponseHelper.php
```diff
@@ -149,7 +149,10 @@ protected function assertRedirectResponse(HttpFoundation\Response $response)
 
     final protected function assertSessionHasFlashMessage(string $message, HttpFoundation\Session\SessionInterface $session)
     {
-        $this->assertContains($message, $session->get('flash'), \sprintf(
+        $flash = $session->get('flash');
+
+        $this->assertNotNull($flash, 'Failed asserting that the session has flash messages.');
+        $this->assertContains($message, $flash, \sprintf(
```
localheinz: If we don't assert that the flash is *not* `null`, then this assertion will fail with

```
PHPUnit\Framework\Exception: Argument #2 (No Value) of PHPUnit\Framework\Assert::assertContains() must be a array, traversable or string
```

localheinz:
Thank you, @BackEndTea!

PR #942, opened 2017-12-15
Use the database instead of mocking AccountManagement

derrabus:
This PR removes swapping of the `AccountManagement` service. Instead, the real service is used with data that has been written to the database.

Follows #934.
Related to #618.


localheinz:
Thank you, @derrabus!

PR #943, opened 2017-12-15
Don't swap the speakers service

derrabus:
This PR removes all swapping of the `application.speakers` service.

Related to #618.

### DIFF
--- a/classes/Provider/TestingServiceProvider.php
+++ b/classes/Provider/TestingServiceProvider.php
@@ -14,7 +14,9 @@ declare(strict_types=1);
 namespace OpenCFP\Provider;
 
 use OpenCFP\Domain\Services\Authentication;
+use OpenCFP\Domain\Services\IdentityProvider;
 use OpenCFP\Test\Helper\MockableAuthenticator;
+use OpenCFP\Test\Helper\MockableIdentityProvider;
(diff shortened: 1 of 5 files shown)


chartjes:
@derrabus Could you explain to me how the tests know to use the MockableIdentityProvider? It's not immediately obvious to me.

derrabus:
@chartjes This works the same way as with the `MockableAuthenticator`.

In the testing environment, the `IdentityProvider` service points to `MockableIdentityProvider` which decorates the Sentinel implementation. If we feed a user object into that class, it will always serve that object. If not, it will fall back to Sentinel.

chartjes:
@derrabus Again, sorry for my not understanding this. To me it looks like magic because I cannot see the mechanism where it actually happens. If it happens because of the annotation, that's fine. I am just trying to understand it! 👍

chartjes:
Thanks @derrabus

derrabus:
@chartjes No, it's not the annotation that does this. It's this piece of code that we only load in testing environment:

https://github.com/opencfp/opencfp/blob/d58e15ac757bad16bdbf48ddcfaaf73dfd26ed4f/classes/Provider/TestingServiceProvider.php#L30-L32

PR #944, opened 2017-12-15
Moved SymfonySentinelSession out of the Provider namespace

derrabus:
This PR moves the class `SymfonySentinelSession` out of the provider namespace as it is not really a service provider. Plus, when switching to Symfony, that class would be the only one left in that namespace.

Related to #618.


chartjes:
Thanks @derrabus

PR #945, opened 2017-12-15
Fix: Rename and move repository

localheinz:
This PR

* [x] renames and moves a repository

### DIFF
--- a/classes/Domain/Repository/UserRepository.php
+++ b/classes/Domain/Repository/UserRepository.php
@@ -11,28 +11,28 @@ declare(strict_types=1);
  * @see https://github.com/opencfp/opencfp
  */
 
-namespace OpenCFP\Domain\Speaker;
+namespace OpenCFP\Domain\Repository;
 
 use OpenCFP\Domain\EntityNotFoundException;
 use OpenCFP\Domain\Model\User;
 
-interface SpeakerRepository
+interface UserRepository
 {
     /**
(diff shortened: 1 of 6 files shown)


classes/Infrastructure/Auth/SentinelIdentityProvider.php
```diff
@@ -28,14 +28,14 @@
     private $sentinel;
 
     /**
-     * @var SpeakerRepository
+     * @var UserRepository
      */
-    private $repo;
```
localheinz: While at it, renaming this field, too.

classes/Domain/Repository/UserRepository.php
```diff
@@ -11,28 +11,28 @@
  * @see https://github.com/opencfp/opencfp
  */
 
-namespace OpenCFP\Domain\Speaker;
+namespace OpenCFP\Domain\Repository;
```
localheinz: I think it could be useful to add more repositories here (and actual implementations), so that we can use those in controllers, instead of using static methods to retrieve models. This would have the advantage that more tests can be turned into unit tests.

chartjes: Yes, correctly grouping them as repositories would be good.

chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #946, opened 2017-12-15
Enhancement: Inline trait and use interface to announce requirement

localheinz:
This PR

* [x] inlines the `DatabaseInteraction` trait into `WebTestCase` and uses an interface to announce the requirement to refresh the database

Follows #936.


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #947, opened 2017-12-15
When in testing environment, don't deliver any mails

derrabus:
This PR disables email delivery in testing environment. This way, we don't need to mock SwiftMailer just to make it stop sending mails.

Follows #902.

### DIFF
--- a/classes/Provider/SwiftMailerServiceProvider.php
+++ b/classes/Provider/SwiftMailerServiceProvider.php
@@ -13,9 +13,11 @@ declare(strict_types=1);
 
 namespace OpenCFP\Provider;
 
+use OpenCFP\Environment;
 use Pimple\Container;
 use Pimple\ServiceProviderInterface;
 use Swift_Mailer;
+use Swift_NullTransport;
 use Swift_SmtpTransport;
 
 final class SwiftMailerServiceProvider implements ServiceProviderInterface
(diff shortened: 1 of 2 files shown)


classes/Provider/SwiftMailerServiceProvider.php
```diff
@@ -25,12 +27,16 @@
      */
     public function register(Container $app)
     {
-        $app['mailer'] = function ($app) {
-            $transport = (new Swift_SmtpTransport($app->config('mail.host'), $app->config('mail.port')))
-                ->setUsername($app->config('mail.username'))
-                ->setPassword($app->config('mail.password'));
+        $app['mailer.transport'] = function ($app) {
+            return $app['env'] === Environment::TYPE_TESTING
```
localheinz: To enhance legibility, can we turn this ternary statement into an `if` with an early `return` instead?

classes/Provider/SwiftMailerServiceProvider.php
```diff
@@ -25,12 +27,28 @@
      */
     public function register(Container $app)
     {
-        $app['mailer'] = function ($app) {
-            $transport = (new Swift_SmtpTransport($app->config('mail.host'), $app->config('mail.port')))
+        $app['mailer.transport'] = function ($app) {
+            /** @var Environment $environment */
+            $environment = $app['env'];
```
localheinz: See [`OpenCFP\Application`](https://github.com/opencfp/opencfp/blob/fdd32f4dcbeb01061afbfaa0592aa308eeebdf39/classes/Application.php#L46-L51.)

localheinz:
Thank you, @derrabus!

PR #948, opened 2017-12-15
Fix: Create database dump directly after migration

localheinz:
This PR

* [x] ignores `tests/dump.sql` and has it exported directly after running migrations

💁‍♂️ This ensures that we don't have do update it manually, and that it is up-to-date at all times.


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #895, opened 2017-12-07
Switch to Symfony

derrabus:
This PR switches the application to Symfony. 

* [x] Upgrade to Symfony 3.4.
* [x] Eliminate container-awareness.
* [x] Convert middlewares to event subscribers.
* [x] Configure a Symfony kernel for OpenCFP.
* [x] Switch front controller to the Symfony kernel.
* [x] Switch console to the Symfony kernel.
* [x] Switch integration tests to the Symfony kernel.
* [x] Remove the `Application` and all providers.
* [x] Remove `silex/silex`.
* ~~Raise php requirement to 7.1.~~
* ~~Upgrade to Symfony 4.0.~~

This PR switches OpenCFP to a new Symfony kernel and removes all dependencies to the Silex package.

Fixes #618.

Some notes:
* I've placed all Symfony configuration in `/resources/config` because Symfony's default `/config` is already used for config files that the user is supposed to edit. If you know a better place, please tell me.
* The routing configuration looks ugly because I use a script to regenerate it every time I rebase against master. We can split that file after the PR has been merged – or switch to annotations if you like.
* We still depend on a couple of libraries that are not compatible with Symfony 4. I suggest to remain on Symfony 3.4 for the time being.
* I know the diff is large, but I don't see much that I could pull out of the PR. The biggest part is removing the Silex bootstrap and replacing it with a Symfony one.
* I did not include Symfony Flex because it enforces a very opinionated repository layout that would clash with the current structure of this application.

PR #950, opened 2017-12-16
Removed switch to disable the cache

derrabus:
This PR removes the switch to disable the cache. Once we switch to Symfony, we don't have a choice but to maintain a cache directory, because Symfony operates on a compiled container and router and therefore requires a cache.

Related to #618.
Preparation for #895.


localheinz:
Thank you, @derrabus!

PR #951, opened 2017-12-16
Removed remaining calls to WebTestCase::swap()

derrabus:
This PR removes the remaining calls to the `WebTestCase::swap()` method. That means that our integration tests do not modify the container anymore, which is very helpful for the Symfony migration.

Unfortunately, I had to kill one test case for this. I don't see a good way to simulate inside an integration test that the reset email could not be sent. It would probably be better to cover this with unit tests.

Follows #618.
Prepares #895.


chartjes:
Thanks @derrabus

PR #952, opened 2017-12-16
Fix: Properly set up test doubles

localheinz:
This PR

* [x] properly sets up test doubles (and adds missing tests)
* [x] reduces scope of try-and-catch blocks

Blocks #895.

/cc @derrabus

PR #953, opened 2017-12-16
Fix: Do not import everything

localheinz:
This PR

* [x] reduces imports

Related to #952.

💁‍♂️ I think there's no need to always import every class, but rather appropriate parent or ancestor namespaces.


BackEndTea:
I kinda like having all imports at the top. This keeps it clear exactly what the class uses

localheinz:
@BackEndTea 

The problem with importing every class used is that one eventually ends up adding unnecessary aliases, which makes things even more complicated - also see #954.

mdwheele:
I don't think this class is at-risk for the problem referenced in #954. I feel it's more a matter of taste rather than a "gateway drug" to aliasing all the things :grin:

PR #954, opened 2017-12-16
Fix: Avoid unnecessary aliases

localheinz:
This PR

* [x] avoids unnecessary aliases

Related to #953.

### DIFF
--- a/classes/Http/Form/ForgotForm.php
+++ b/classes/Http/Form/ForgotForm.php
@@ -17,13 +17,13 @@ use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\EmailType;
 use Symfony\Component\Form\Extension\Core\Type\SubmitType;
 use Symfony\Component\Form\FormBuilderInterface;
-use Symfony\Component\Validator\Constraints as Assert;
+use Symfony\Component\Validator;
 
 class ForgotForm extends AbstractType
 {
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $builder->add('email', EmailType::class, ['constraints' => [new Assert\Email()]])
+        $builder->add('email', EmailType::class, ['constraints' => [new Validator\Constraints\Email()]])
             ->add('send', SubmitType::class, ['label' => 'Reset my password']);
     }
 }
--- a/classes/Infrastructure/Auth/SentinelAccountManagement.php
+++ b/classes/Infrastructure/Auth/SentinelAccountManagement.php
@@ -15,7 +15,7 @@ namespace OpenCFP\Infrastructure\Auth;
 
 use Cartalyst\Sentinel\Roles;
 use Cartalyst\Sentinel\Sentinel;
-use Cartalyst\Sentinel\Users\UserInterface as SentinelUserInterface;
+use Cartalyst\Sentinel\Users;
 use OpenCFP\Domain\Services\AccountManagement;
 
 final class SentinelAccountManagement implements AccountManagement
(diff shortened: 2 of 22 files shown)


classes/Http/Form/ForgotForm.php
```diff
@@ -17,13 +17,13 @@
 use Symfony\Component\Form\Extension\Core\Type\EmailType;
 use Symfony\Component\Form\Extension\Core\Type\SubmitType;
 use Symfony\Component\Form\FormBuilderInterface;
-use Symfony\Component\Validator\Constraints as Assert;
+use Symfony\Component\Validator;
```
derrabus: It's actually a common practice to import Validator constraints with alias `Assert`. 😕

classes/Infrastructure/Auth/SentinelAccountManagement.php
```diff
@@ -34,7 +34,7 @@ public function findById(int $userId): UserInterface
     {
         $user = $this->sentinel->getUserRepository()->findById($userId);
 
-        if ($user instanceof SentinelUserInterface) {
+        if ($user instanceof Users\UserInterface) {
```
derrabus: The alias was actually more readable to me, tbh. 😕

classes/Infrastructure/Auth/SymfonySentinelSession.php
```diff
@@ -13,13 +13,13 @@
 
 namespace OpenCFP\Infrastructure\Auth;
 
-use Cartalyst\Sentinel\Sessions\SessionInterface as SentinelSessionInterface;
-use Symfony\Component\HttpFoundation\Session\SessionInterface as SymfonySessionInterface;
+use Cartalyst\Sentinel\Sessions;
+use Symfony\Component\HttpFoundation;
```
derrabus: In this file, we're adapting two different `SessionInterfaces`s. The aliases did improve readability here. 😕

mdwheele:
I agree with @derrabus here. There's nothing "wrong" with using aliases. Those that are there to resolve naming conflicts are arguably candidates for this type of change, but even still, a lot of them could be argued as aliases for readability anyway.

Not of the same class of alias, but consider the Illuminate database manager. If I did not alias that, I would be type-hinting `Manager` in my infrastructure classes rather than `Database`, which is certainly more indicative of what's going on.

I'm :-1: for sweeping change that removes all aliases.

localheinz:
@mdwheele 

What's the advantage of using aliases when they aren't needed?

mdwheele:
> What's the advantage of using aliases when they aren't needed?

When they aren't needed, they aren't needed. If they arguably improve readability (`SentinelSessionInterface` vs `SymfonySessionInterface`) or are a idiomatic / common use of a library (`Validator\Constraints as Assert`), I think they are nice to have.

localheinz:
@mdwheele 

> When they aren't needed, they aren't needed. If they arguably improve readability (SentinelSessionInterface vs SymfonySessionInterface) or are am idiomatic / common use of a library (Validator\Constraints as Assert), I think they are nice to have.

I don't understand the point you are making here, can you explain?

classes/Infrastructure/Auth/SentinelAccountManagement.php
localheinz: How is this more readable?

mdwheele:
> I don't understand the point you are making here, can you explain?

* If an alias truly is not needed, then it is justifiably a candidate for removal.
* However...
  * If it subjectively improves readability, then that is an argument to keep the alias (as exemplified by [this comment](https://github.com/opencfp/opencfp/pull/954#discussion_r157341731)) 
  * If it is common / community-standard to alias particular classes for readability (as exemplified by [this comment](https://github.com/opencfp/opencfp/pull/954#discussion_r157341716) then that is an argument to keep the alias.

I do not think that aliases make things more complicated as a blanket judgement. I think they are a tool that makes sense to use sometimes and not others. There are several changes in this PR that would be positive use-cases for alias and it seems we're blanket removing all aliases on the premise that they are unecessary. PHP, itself, is technically unnecessary. I could be writing C or any other lower-level implementation. However, PHP improves readability.

PR #956, opened 2017-12-16
Type-hint for the interface instead of the implementation.

derrabus:
This PR replaces `CsrfTokenManager` with `CsrfTokenManagerInterface`in type-hints.


localheinz:
Thank you, @derrabus!

PR #957, opened 2017-12-16
Fix: Extend from AbstractActionTestCase

localheinz:
This PR

* [x] extends from `AbstractActionTestCase`

Follows #935.


localheinz:
Thank you, @BackEndTea!

PR #958, opened 2017-12-16
Fix: Remove covers annotations

localheinz:
This PR

* [x] removes `@covers` annotations

Follows https://github.com/opencfp/opencfp/pull/917.

### DIFF
--- a/tests/Unit/ApplicationTest.php
+++ b/tests/Unit/ApplicationTest.php
@@ -16,10 +16,6 @@ namespace OpenCFP\Test\Unit;
 use OpenCFP\Application;
 use OpenCFP\Environment;
 
-/**
- * @covers \OpenCFP\Application
- * @group db
- */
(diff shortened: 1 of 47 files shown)


tests/Unit/ApplicationTest.php
```diff
@@ -16,10 +16,6 @@
 use OpenCFP\Application;
 use OpenCFP\Environment;
 
-/**
- * @covers \OpenCFP\Application
- * @group db
```
localheinz: I believe this annotation is not relevant anymore.

tests/Unit/Console/ApplicationTest.php
```diff
@@ -19,10 +19,6 @@
 use OpenCFP\Console\Application;
 use Symfony\Component\Console;
 
-/**
- * @group db
```
localheinz: I believe this annotation is not relevant anymore.

localheinz:
Thank you, @chartjes!

PR #959, opened 2017-12-16
Enhancement: Extract UpdateAction from TalkController

localheinz:
This PR

* [x] extracts an `UpdateAction` from the `TalkController`

Follows #912.


classes/Http/Action/Talk/UpdateAction.php
```diff
@@ -0,0 +1,236 @@
...
+            'level'       => $request->get('level'),
+            'category'    => $request->get('category'),
+            'desired'     => $request->get('desired'),
+            'slides'      => $request->get('slides'),
+            'other'       => $request->get('other'),
+            'sponsor'     => $request->get('sponsor'),
+            'user_id'     => $request->get('user_id'),
+        ];
+
+        $form = $this->createTalkForm($requestData);
```
localheinz: Since the form isn't injected, but created here, we can only write unit tests up until the condition above.

localheinz:
Thank you, @chartjes!

PR #960, opened 2017-12-16
Fix: No need to wrap single argument

localheinz:
This PR

* [x] removes unnecessary wrapping


localheinz:
Thank you, @mdwheele!

PR #961, opened 2017-12-16
Fix: Remove empty class attributes

localheinz:
This PR

* [x] removes empty `class` attributes


localheinz:
Thank you, @mdwheele!

PR #964, opened 2017-12-17
Enhancement: Extract IndexAction from ForgotController

localheinz:
This PR

* [x] extracts an `IndexAction` from the `ForgotController`

Follows #912.

### DIFF
--- a/resources/config/services/services.yml
+++ b/resources/config/services/services.yml
@@ -21,6 +21,17 @@ services:
     bind:
       $showSubmissionCount: '%application.show_submission_count%'
 
+  OpenCFP\Http\Action\Forgot\IndexAction:
+    public: true
+    bind:
+      $resetForm: '@OpenCFP\Http\Form\ForgotForm'
+
+  OpenCFP\Http\Form\ForgotForm:
+    class: Symfony\Component\Form\FormInterface
+    factory: [OpenCFP\Http\Form\FormFactory, create]
+    bind:
+      $formType: OpenCFP\Http\Form\ForgotFormType
+
   OpenCFP\Http\View\TalkHelper:
     arguments:
       - '%talk.categories%'
--- a/tests/Unit/Http/Form/FormFactoryTest.php
+++ b/tests/Unit/Http/Form/FormFactoryTest.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Copyright (c) 2013-2017 OpenCFP
+ *
+ * For the full copyright and license information, please view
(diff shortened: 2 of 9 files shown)


classes/Provider/Gateways/WebGatewayProvider.php
```diff
@@ -71,6 +73,18 @@ public function register(Container $app)
             return new Action\Page\TalkIdeasAction($app['twig']);
         };
 
+        $app[Action\Forgot\IndexAction::class] = function ($app) {
+            /** @var FormFactoryInterface $formFactory */
+            $formFactory = $app['form.factory'];
+
+            $resetForm = $formFactory->createBuilder(ForgotForm::class)->getForm();
```
localheinz: I believe we don't need to inject the entire `FormFactory` if all we need is an actual form.

resources/config/services/services.yml
```diff
@@ -21,6 +21,17 @@ services:
     bind:
       $showSubmissionCount: '%application.show_submission_count%'
 
+  OpenCFP\Http\Action\Forgot\IndexAction:
+    public: true
+    bind:
+      $resetForm: '@OpenCFP\Http\Form\ForgotForm'
+
+  OpenCFP\Http\Form\ForgotForm:
```
localheinz: @derrabus 

Since I believe you are the one who has the most experience with Symfony, can you take a look and suggest a better way to wire up an actual form from a form type, if you can think of one?

classes/Http/Form/ForgotFormType.php
```diff
@@ -19,7 +19,7 @@
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Validator\Constraints as Assert;
 
-class ForgotForm extends AbstractType
+final class ForgotFormType extends AbstractType
```
localheinz: Renamed this class as it's not actually a form, but a definition of a form, right?

BackEndTea: Since this change is unrelated to the rest of the PR, maybe move it to another PR?

tests/Unit/Http/Form/FormFactoryTest.php
```diff
@@ -0,0 +1,48 @@
...
+final class FormFactoryTest extends Framework\TestCase
+{
+    use Helper;
+
+    public function testCreateCreatesForm()
+    {
+        $formType = $this->faker()->word;
+
+        $form = $this->createMock(Form\FormInterface::class);
+
```
chartjes: Shouldn't this be done with Prophecy?

localheinz:
Thank you, @chartjes!

PR #969, opened 2017-12-17
Use template annotation on actions

derrabus:
This PR enables SensioFrameworkExtraBundle which allows us to configure controllers with annotations. It also adds `@Template` annotations to actions that render templates. This allows us to return an array with template variables instead of a `Response` object, so we do not need to interface with Twig directly anymore.

~~The PR includes #895 in its full glory, please only look at the last commit of the branch.~~


derrabus:
I've addressed the failing integration test and added two minimal tests for the empty action classes. I don't expect those two action classes to survive much longer, though. 😉

derrabus:
WIP label can be removed.

localheinz:
Thank you, @derrabus!

PR #970, opened 2017-12-17
Enhancement: Extract EditAction from ProfileController

localheinz:
This PR

* [x] extracts an `EditAction` from the `ProfileController`

Follows #912.


classes/Http/Action/Profile/EditAction.php
```diff
@@ -0,0 +1,96 @@
...
+                'short' => 'Error',
+                'ext'   => "You cannot edit someone else's profile",
+            ]);
+
+            $url = $this->urlGenerator->generate('dashboard');
+
+            return new HttpFoundation\RedirectResponse($url);
+        }
+
+        $speakerData = Model\User::find($user->getId())->toArray();
```
localheinz: Because of static access here we cannot write unit tests that go further than the condition above.

classes/Http/Action/Profile/EditAction.php
```diff
@@ -0,0 +1,96 @@
...
+    private $urlGenerator;
+
+    public function __construct(
+        Services\Authentication $authentication,
+        PathInterface $path,
+        Twig_Environment $twig,
+        Routing\Generator\UrlGeneratorInterface $urlGenerator
+    ) {
+        $this->authentication = $authentication;
+        $this->path           = $path;
```
derrabus: Since the information is static anyway, why don't we pass `$path->downloadFromPath()` to this class instead of the whole path object?

mdwheele:
Could you rebase onto `master`?

resources/config/services/services.yml
```diff
@@ -21,6 +21,11 @@ services:
     bind:
       $showSubmissionCount: '%application.show_submission_count%'
 
+  OpenCFP\Http\Action\Profile\EditAction:
+    public: true
+    bind:
+      $downloadFromPath: 'foo'
```
localheinz: @derrabus 

Can you give me a hint how I properly bind the `$downloadFromPath` from `PathInterface::downloadFromPath()`?

classes/Http/Action/Profile/EditAction.php
chartjes: I've been thinking about this sort of thing for a while now. It occurs to me that monkey patching is the best solution here, but I don't know if there is a good solution that works with PHPUnit.

localheinz:
Thank you, @chartjes, will do!

PR #971, opened 2017-12-17
Enhancement: Use TemplateController to render login

localheinz:
This PR

* [x] renders the login template using the `TemplateController`

Follows #976.


localheinz:
Thank you, @chartjes!

PR #972, opened 2017-12-17
Quote strings starting with percent character in YAML files.

derrabus:
This PR adds quote to a YAML string starting with a `%` character. This establishes forward compatibility with the stricter parsing rules in Symfony YAML 4.0.

Related to #618.

Reference: https://github.com/symfony/symfony/blob/4.0/src/Symfony/Component/Yaml/CHANGELOG.md#400


localheinz:
Thank you, @derrabus!

PR #973, opened 2017-12-17
Upgrade aptoma/twig-markdown to 2.0

derrabus:
This PR upgrades the `aptoma/twig-markdown` library to 2.0. The currently used version of that library accesses deprecated Twig functionality.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
     },
     "require": {
         "php": ">=7",
-        "aptoma/twig-markdown": "0.2.*",
+        "aptoma/twig-markdown": "^2",
         "cartalyst/sentinel": "^2.0.17",
         "ezyang/htmlpurifier": "dev-master",
         "guzzlehttp/guzzle": "~6.2",
--- a/composer.lock
+++ b/composer.lock
@@ -4,31 +4,36 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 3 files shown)


derrabus:
I'm scratching my head over the test failures. I cannot reproduce them and they don't seem related to my change. Also, only one of the three test runs is red. 🤔

mdwheele:
Gonna request it run again on same commit. Could be something specific to PHP 7.0. Unrelated: seeing "Unit", "Mockery" and "Infrastructure" in the same failure message really makes me want to rewrite as an Integration test. This is a good example of something breaking and the test doesn't really point any fingers in the right direction other than the test method that failed. 

At any rate, it does look like a failure outside the scope of this PR if I'm looking at this right.

localheinz:
Build failure seems unrelated, but the failing tests could be improved anyway.

Question is whether we want to merge this and risk a failing `master`, or sort out the failing tests in a separate PR first, and then see if these do not bring this build to fail here.

Have not been able to reproduce the build failures locally.

```
$ php -v

PHP 7.0.25 (cli) (built: Nov 11 2017 12:08:41) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Xdebug v2.5.5, Copyright (c) 2002-2017, by Derick Rethans
```

mdwheele:
Looks like that worked. Going to merge this and close #974.

mdwheele:
Heh... fun times... blocked from merging for some reason and can't approve.

PR #975, opened 2017-12-18
Removed the already replaced editAction from TalkController

derrabus:
The merge of #925 went a bit wrong: The old action method is still there and the routing still points to it. This PR fixes this.

Follows #925.


chartjes:
Ugh, master is still broken for me. What the heck did I do wrong?!?!?

localheinz:
Ha, nice catch, @derrabus!

PR #967, opened 2017-12-17
Enhancement: Extract PasswordAction from ForgotController

localheinz:
This PR

* [x] extracts a `PasswordAction` from the `ForgotController`

Follows #912.

PR #976, opened 2017-12-18
Deliver static templates via TemplateController

derrabus:
As suggested in https://github.com/opencfp/opencfp/pull/967#discussion_r157367804, we don't need to write controllers that only deliver a template without variables. /cc @localheinz 

This PR offloads those three actions to Symfony's `TemplateController`

Replaces #967.


mdwheele:
:+1: with the idea. We just need to fix tests.

localheinz:
Thank you, @derrabus!

PR #981, opened 2017-12-18
Enhancement: Configure memory limit in phpunit.xml.dist

### LINKED ISSUE #978: Current master integration suite hates folks with 128M limit

1. Have `memory_limit = 128M`
2. Be me, on Docker
3. Apply a bunch of fix-ups to testing configuration, Makefile and more to get a test database set up on another container
4. Once you stop getting "Connection Refused" errors after tracking all the places that assume `localhost`, run the integration suite!

# Expected Result

* Passes like champion runner! :muscle: :racehorse: :speedboat: 

# Actual Result

```bash
mdwheele@28d9116def76:/usr/src/app$ vendor/bin/phpunit --testsuite integration --stop-on-error
PHPUnit 6.5.4 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.1.12
Configuration: /usr/src/app/phpunit.xml.dist

......
Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /usr/src/app/vendor/fzaninotto/faker/src/Faker/Provider/Text.php on line 93

Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /usr/src/app/vendor/symfony/phpunit-bridge/DeprecationErrorHandler.php on line 206
```

*One* of the "memory leaks" is from static methods in test suite that instantiate `Capsule` through our database reset mechanism. There are more. If I set `memory_limit = -1`, the suite does pass, but takes a long time. I'm going to spend some time tomorrow night on figuring out why the integration suite went from ~30 seconds for me a few days ago, to ~7m30s. It's probably something on my end, but I haven't changed this Docker setup in a while.

localheinz:
This PR

* [x] configures the `memory_limit` in `phpunit.xml.dist` to `512M`

Fixes #978.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -16,6 +16,7 @@
     <php>
         <!-- https://github.com/symfony/console/blob/v3.3.12/Terminal.php#L24-L36 -->
         <env name="COLUMNS" value="120"/>
+        <ini name="memory_limit" value="512M"/>
     </php>
     <filter>
         <whitelist>


BackEndTea:
@mdwheele Does this fix your issue?

localheinz:
@BackEndTea 

It fixes the issue for me if I lower the limit in my `php.ini` to something like `50M`, and then run the tests.

mdwheele:
I'm down with this. I'm also fine pushing this back on the developer to adjust their own configs. That said, 300M to execute seems high. 

How about we merge this as a workaround and as things stabilize, we can work on cutting down memory usage.

mdwheele:
Thanks @localheinz and @BackEndTea!

localheinz:
Thank you, @BackEndTea and @mdwheele!

PR #965, opened 2017-12-17
Enhancement: Enable no_whitespace_in_blank_line fixer

localheinz:
This PR

* [x] enables the `no_whitespace_in_blank_line` fixer
* [x] runs `make cs`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.9.0#usage:

>**no_whitespace_in_blank_line** [`@Symfony`]
>
>Remove trailing whitespace at the end of blank lines.

PR #982, opened 2017-12-18
Fix: Rename class

localheinz:
This PR

* [x] renames a class

Follows https://github.com/opencfp/opencfp/pull/964#discussion_r157432415.


classes/Http/Controller/ForgotController.php
```diff
@@ -137,7 +137,7 @@ public function resetAction(Request $request): Response
                 'ext'   => $errorMessage,
             ]);
         }
-
```
localheinz: See #965.

localheinz:
Thank you, @BackEndTea!

PR #955, opened 2017-12-16
Enhancement: Extract CreateAction from TalkController

localheinz:
This PR

* [x] extracts a `CreateAction` from the `TalkController`

Follows #912.

PR #984, opened 2017-12-18
Fix broken build from forgotten mock factory

mdwheele:
See: https://travis-ci.org/opencfp/opencfp/builds/318070759

`createTwigMock` doesn't exist and tests fail. This PR adds `createTwigMock` to `AbstractActionTestCase`.

This is my bad as I didn't realize the rebase from https://github.com/opencfp/opencfp/pull/955 was broken.


mdwheele:
Going ahead and executive-merging this to get `master` green again. I'm still re-learning OpenCFP after the *storm* of recent changes over the past month so if this was the "wrong" solution, we can patch up afterwards.

mdwheele:
Aaaaaaaaand... nevermind, because apparently I can't do that. Requesting review...

localheinz:
Thank you, @mdwheele!

Issue #983, opened 2017-12-18
Is there any reason to use Mockery and the PHPUnit mock facilities?


mdwheele:
I've noticed recently we're using the PHPUnit mock facilities in new tests. We've previously preferred Mockery and had a few PHPUnit mocks sprinkled around like magic fairy dust :rainbow:.

Are there legitimate reasons for using both or is this another preference we should officially decide on? 

I personally like Mockery's interface better for readability. However, my opinion is not too strong on the matter. I mostly worry about newcomers wondering "Do I PHPUnit? Do I Mockery?" and bailing.

localheinz:
We could also consider using `phpspec/prophecy`. 

Also see https://thephp.cc/news/2015/02/phpunit-4-5-and-prophecy.

mdwheele:
> We could also consider using phpspec/prophecy.

I'm down with that. I've only ever looked at Prophecy through the eyes of PHPSpec, but if we picked this up, it looks like we'd be able to drop a dependency and keep something similar to the "more-readable" interface of Mockery.

chartjes:
Prophecy is baked into PHPUnit these days so it's not the worst idea to use it. If the goal is to reduce dependencies, then this is a worthy task.

Personally I have moved to using Prophecy to teach folks in my workshops.

mdwheele:
Let's do it!

For me, reducing dependencies happened to be a side-effect of making our approach to testing more consistent / accessible. For our current usage, it looks like Prophecy will be a great replacement for Mockery / current PHPUnit mocks.

Can we agree that new PRs that do any mock/stub should use Prophecy rather than PHPUnit mocks / Mockery?

localheinz:
👍

PR #986, opened 2017-12-18
Fix: Mark methods creating mocks as deprecated

localheinz:
This PR

* [x] marks methods creating mocks as `@deprecated`

Follows #985.
Related to #983.

💁‍♂️ Marking these methods as `@deprecated` emphasises` that - going forward - we don't want to use the internal mocking facilities of `phpunit/phpunit` anymore.

### DIFF
--- a/tests/Unit/Http/Action/AbstractActionTestCase.php
+++ b/tests/Unit/Http/Action/AbstractActionTestCase.php
@@ -20,11 +20,16 @@ use Symfony\Component\HttpFoundation;
 use Symfony\Component\Routing;
 use Twig_Environment;
 
+/**
+ * @deprecated
+ */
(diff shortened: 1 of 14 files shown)


tests/Unit/Http/Action/AbstractActionTestCase.php
```diff
@@ -20,11 +20,16 @@
 use Symfony\Component\Routing;
 use Twig_Environment;
 
+/**
+ * @deprecated
+ */
 abstract class AbstractActionTestCase extends Framework\TestCase
```
localheinz: Since all this abstract test case does is

* import the `Helper` trait
* provide methods which create mocks

we should probably remove it eventually.

localheinz:
Thank you, @chartjes!

PR #987, opened 2017-12-18
Fix: Explicitly disable php_unit_test_class_requires_covers fixer

localheinz:
This PR

* [x] explicitly disables the `php_unit_test_class_requires_covers` fixer

Follows #917.

💁‍♂️ I believe explicitly disabling the fixer, rather than just removing it, sends the signal that a) we are aware of the fixer, and b) have decided not to use it (anymore).


localheinz:
Thank you, @chartjes!

PR #988, opened 2017-12-18
Enhancement: Enable and configure general_phpdoc_annotation_remove fixer

localheinz:
This PR

* [x] enables and configures the `general_phpdoc_annotation_remove` fixer
* [x] runs `make cs`

Follows #958.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.9.0#usage:

>**general_phpdoc_annotation_remove**
>
>Configured annotations should be omitted from phpdocs.
>
>Configuration options:
>
>* `annotations` (`array`): list of annotations to remove, e.g. `["author"]`; defaults to `[]`

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -63,6 +63,12 @@ return PhpCsFixer\Config::create()
         'declare_strict_types' => true,
         'function_to_constant' => true,
         'function_typehint_space' => true,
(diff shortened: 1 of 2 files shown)


localheinz:
Thank you, @chartjes!

Issue #721, opened 2017-11-26
Tests are failing for configurations that don't use root as database user.


mdwheele:
# Steps required to reproduce the problem

1. Use Homestead (that has user `homestead:secret`, not `root:`) or any LAMP stack that doesn't use `root` database user.
2. Clone project and `script/test`

# Expected Result

* I should get similar results to TravisCI

# Actual Result

* I get a ton of DB failures related to database configuration:
```
PDOException: SQLSTATE[HY000] [1045] Access denied for user 'root'@'localhost' (using password: NO) in /home/vagrant/Code/opencfp/ven
dor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:43
```

See this diff where functionality was "broken": https://github.com/opencfp/opencfp/commit/40106c02cef65c094ee470de952284a839bc5fb3#diff-1bd3d4d018fb9e3b5295167d8a53e230

I can resolve this, but I also propose that we make a `testing.yml.dist` similar to other configs so that I can set my own settings locally. I don't use database with root users like our TravisCI config does.

mdwheele:
Thoughts:

* Is there any reason for both `DataBaseInteraction` and `RefreshDatabase` to exist? They're almost doing exactly the same thing, except `RefreshDatabase` is expecting an Illuminate Manager to exist in the container.
* Is there any benefit to maintaining a `dump.sql` file rather than simply dropping all tables and running migrations as usual?

mdwheele:
I'll make a pass at this soon. My current direction will be:

- Get a passing suite by allowing a `root` database user.
- Eliminate `DataBaseInteraction` and substitute `RefreshDatabase` at every usage. Fix broken that results from that.
- Modify `RefreshDatabase` to behave like the following:
  1. Make sure database is migrated exactly once.
  2. Wrap test methods in transactions.
  3. End-state should be a fully-migrated database with only seed data.
- Make `testing` environment database configuration variable.
- Verify tests pass. 
- Fight TravisCI for honor.

BackEndTea:
He difference between the 2 traits is that one will set up the db before every test. The other one sets up the db before the class. This was done since some tests just read from the db to test things.

The dump.sql file is there to speed up the tests. Its faster then migrating up for every test. Since that inckudes a lot of changes and all the aurports. Though i agreee its a bit of a crude option to use an sql file.

On transactions, would we still be able to test for example the delete action in the speaker controller, whih is wrapped in a transaction

[...]

PR #991, opened 2017-12-18
Fix: Mark RefreshDatabase trait as deprecated

localheinz:
This PR

* [x] marks the `RefreshDatabase` trait as deprecated

Related to https://github.com/opencfp/opencfp/issues/721.

💁‍♂️ I believe we are better of if we have a single means of resetting the database to a known state, and that should involve fetching the database connection from the container, rather than manually creating it.

### DIFF
--- a/tests/Helper/RefreshDatabase.php
+++ b/tests/Helper/RefreshDatabase.php
@@ -15,6 +15,9 @@ namespace OpenCFP\Test\Helper;
 
 use Illuminate\Database\Capsule\Manager;
 
+/**
+ * @deprecated
+ */
 trait RefreshDatabase
 {
     final protected static function setUpDatabase()


mdwheele:
The direction I've taken in my work on #721 is to either use `RefreshDatabase` or the newly created `RequiresDatabaseReset` interface to mark tests that need database to be migrated and test methods to run in a transaction. 

What would you say to making `WebTestCase` inherently wrap all test methods in a transaction (fetching the connection from container) and migrate the test database exactly once per suite run? Then we could get rid of both markers...

mdwheele:
As far as `@deprecating` this, I think that's fine, but would like to pick a direction and stop churning the same code over and over.

localheinz:
> What would you say to making WebTestCase inherently wrap all test methods in a transaction (fetching the connection from container) and migrate the test database exactly once per suite run? Then we could get rid of both markers...

I've done something like that before. Eventually, we might consider using transactions in services, so we'll need to be able to distinguish between tests that can, and those that cannot be wrapped in transactions.

First of all, I would like to get rid of setting up fixtures for an entire test class, and instead only set up fixtures per test method. This would allow to get rid of the `RefreshDatabase` trait and use the `RequiresDatabaseRefresh` interface instead. In a next step, we could rename the `RequiresDatabaseRefresh` to `TransactionalTestCase`, and then, `setUp()` and `tearDown()`, if we detect the interface, start and roll back a transaction, alternatively, restore the database using the database connection and the SQL dump. This is how I have been doing it in the past, and I believe it worked quite well.

In regard to shared fixtures, I believe it's better to set up fixtures on a per test basis as it keeps the scope small. If I have a bunch of different tests using the same fixtures, it's not quite clear whether all of these are needed for each test, or just one of them.

localheinz:
Thank you, @mdwheele!

PR #992, opened 2017-12-18
Enhancement: Support for Symfony server console command

chartjes:
This PR

* [x] adds support for the Symfony web server (http://symfony.com/doc/current/setup/built_in_web_server.html)
* [x] updates documentation for how to use it 
* [x] removes the old `script/server` script since it is not needed


composer.json
```diff
@@ -39,6 +39,7 @@
         "symfony/twig-bridge": "^3.4",
         "symfony/twig-bundle": "^3.4",
         "symfony/validator": "^3.4",
+        "symfony/web-server-bundle": "^3.4",
```
BackEndTea: Is there a way to only require this as a dev-dependency?

chartjes: Done, moved to require-dev

README.md
```diff
@@ -117,11 +117,19 @@ To run OpenCFP using [PHP's built-in web server](http://php.net/manual/en/featur
 following command can be run:
 
 ```
-$ script/server
+$ bin/console server:start
 ```
 
 The server uses port `8000`. This is a quick way to get started doing development on OpenCFP itself.
 
+Details on how to use this console command can be found at Symfony's documentation for [using the built-in web server](http://symfony.com/doc/current/setup/built_in_web_server.html)
```
localheinz: How about adding a period at the end of this sentence?

localheinz:

PR #993, opened 2017-12-18
Fix: Only use RefreshDatabase when needed

localheinz:
This PR

* [x] removes `RefreshDatabase` from a test that doesn't need it


localheinz:
Thank you, @mdwheele!

PR #994, opened 2017-12-18
Enhancement: Turn DashboardController into DashboardAction

localheinz:
This PR

* [x] turns the `Admin\DashboardController` into `Admin\DashboardAction`

Follows #912.


classes/Http/Action/Admin/DashboardAction.php
```diff
@@ -0,0 +1,55 @@
...
+        $this->talkFormatter  = $talkFormatter;
+    }
+
+    /**
+     * @Template("admin/index.twig")
+     */
+    public function __invoke(): array
+    {
+        return [
+            'speakerTotal'  => Model\User::count(),
```
localheinz: Because of `static` access here we can't really write any unit tests.

classes/Http/Action/Admin/DashboardAction.php
```diff
@@ -0,0 +1,55 @@
...
+            'talkTotal'     => Model\Talk::count(),
+            'favoriteTotal' => Model\Favorite::count(),
+            'selectTotal'   => Model\Talk::where('selected', 1)->count(),
+            'talks'         => $this->talkFormatter->formatList(
+                Model\Talk::recent()->get(),
+                $this->authentication->user()->getId()
```
localheinz: If I hadn't inlined `$userId` here, then I could write a test that asserts that we *don't* catch any exceptions thrown here, but I don't think it has a lot of value.

tests/Integration/Http/Action/Admin/DashboardActionTest.php
```diff
@@ -11,35 +11,33 @@
...
-        self::$talks = factory(Talk::class, 2)->create();
-    }
-
     /**
      * @test
      */
     public function indexDisplaysListOfTalks()
     {
+        /** @var Eloquent\Collection|Model\Talk[] $talks */
+        $talks = factory(Model\Talk::class, 2)->create();
```
localheinz: No need to set up fixtures for the entire class when there's only one test.

Also see conversation in #991.

localheinz:
Thank you, @chartjes!

PR #997, opened 2017-12-19
Enhancement: Extract IndexAction from TalksController

localheinz:
This PR

* [x] extracts an `Admin\Talk\IndexAction` from the `Admin\TalksController`

Follows #912.

### DIFF
--- a/classes/Http/Action/Admin/Talk/IndexAction.php
+++ b/classes/Http/Action/Admin/Talk/IndexAction.php
@@ -1,0 +1,87 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Copyright (c) 2013-2017 OpenCFP
+ *
+ * For the full copyright and license information, please view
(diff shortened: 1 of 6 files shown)


classes/Http/Action/Admin/Talk/IndexAction.php
```diff
@@ -0,0 +1,81 @@
...
+
+        $formattedTalks = $this->talkFilter->getTalks(
+            $adminUserId,
+            $request->get('filter'),
+            $options
+        );
+
+        $perPage = (int) $request->get('per_page') ?: 20;
+
+        $pagerfanta = new Services\Pagination($formattedTalks, $perPage);
```
localheinz: Because we create dependencies here rather than injecting them, we cannot write any unit tests.

localheinz:
Thank you, @chartjes!

PR #1000, opened 2017-12-19
Fix: Dont copy phinx.yml if it is not needed

BackEndTea:
This PR

* [x] Changes the ```make database``` command to only copy ```phinx.yml``` if it doesnt exist already.


localheinz:
Thank you, @BackEndTea!

BackEndTea:
Thank you @localheinz

PR #1002, opened 2017-12-19
Enhancement: Extract ViewAction from TalksController

localheinz:
This PR

* [x] extracts a `Reviewer\Talk\ViewAction` from the `Reviewer\TalksController`

Follows #912.

### DIFF
--- a/classes/Http/Action/Reviewer/Talk/ViewAction.php
+++ b/classes/Http/Action/Reviewer/Talk/ViewAction.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Copyright (c) 2013-2017 OpenCFP
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ *
+ * @see https://github.com/opencfp/opencfp
+ */
+
+namespace OpenCFP\Http\Action\Reviewer\Talk;
+
+use OpenCFP\Domain\Talk;
+use Symfony\Component\HttpFoundation;
+use Symfony\Component\Routing;
+use Twig_Environment;
+
+final class ViewAction
+{
+    /**
+     * @var Talk\TalkHandler
+     */
+    private $talkHandler;
+
+    /**
+     * @var Twig_Environment
+     */
+    private $twig;
+
+    /**
+     * @var Routing\Generator\UrlGeneratorInterface
+     */
+    private $urlGenerator;
+
+    public function __construct(
+        Talk\TalkHandler $talkHandler,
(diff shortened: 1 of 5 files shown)


classes/Http/Action/Reviewer/Talk/ViewAction.php
```diff
@@ -0,0 +1,70 @@
...
+        $this->talkHandler->grabTalk((int) $request->get('id'));
+
+        if (!$this->talkHandler->view()) {
+            $request->getSession()->set('flash', [
+                'type'  => 'error',
+                'short' => 'Error',
+                'ext'   => 'Could not find requested talk',
+            ]);
+
+            $url = $this->urlGenerator->generate('admin_talks');
```
localheinz: Note how there's a mistake here, as the route `admin_talks` should be `reviewer_talks`.

classes/Http/Controller/Reviewer/TalksController.php
```diff
@@ -35,25 +35,6 @@ public function __construct(TalkHandler $talkHandler, Twig_Environment $twig, Ur
...
-        $this->talkHandler->grabTalk((int) $request->get('id'));
-
-        if (!$this->talkHandler->view()) {
-            $request->getSession()->set('flash', [
-                'type'  => 'error',
-                'short' => 'Error',
-                'ext'   => 'Could not find requested talk',
-            ]);
-
-            return $this->redirectTo('admin_talks');
```
localheinz: Note how there's a mistake here, as the route `admin_talks` should be `reviewer_talks`.

classes/Http/Action/Reviewer/Talk/ViewAction.php
```diff
@@ -0,0 +1,70 @@
...
+        Talk\TalkHandler $talkHandler,
+        Twig_Environment $twig,
+        Routing\Generator\UrlGeneratorInterface $urlGenerator
+    ) {
+        $this->talkHandler  = $talkHandler;
+        $this->twig         = $twig;
+        $this->urlGenerator = $urlGenerator;
+    }
+
+    public function __invoke(HttpFoundation\Request $request): HttpFoundation\Response
```
BackEndTea: Is there a reason for not using ```@Template``` here?

localheinz: Yes, the reason is that we would then have to remove the type declaration, as 

* for a redirect, we would need to return an instance of `HttpFoundation\Response`
* for the success case, we would need to return an array

localheinz:
Thank you, @BackEndTea!

PR #1003, opened 2017-12-19
Fix: Use appropriate route when redirecting

localheinz:
This PR

* [x] uses the proper route name when redirecting

Blocks #1002.


localheinz:
Thank you, @BackEndTea!

PR #1004, opened 2017-12-19
Enhancement: Wrap tests in transactions

localheinz:
This PR

* [x] renames the marker interface `RequiresDatabaseReset` to `TransactionalTestCase` and wraps tests in transactions

Follows https://github.com/opencfp/opencfp/pull/991#issuecomment-352565555.

### Before

```
$ make integration

vendor/bin/phpunit --testsuite integration
PHPUnit 6.5.4 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.0.25
Configuration: /Users/am/Sites/opencfp/opencfp/phpunit.xml.dist

.................................................................................................................................................................                                                                                                                                             161 / 161 (100%)

Time: 42.77 seconds, Memory: 354.00MB

OK (161 tests, 935 assertions)
```

### After

```
$ make integration

vendor/bin/phpunit --testsuite integration
PHPUnit 6.5.4 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.0.25
Configuration: /Users/am/Sites/opencfp/opencfp/phpunit.xml.dist

.................................................................................................................................................................                                                                                                                                             161 / 161 (100%)

Time: 21.78 seconds, Memory: 354.00MB

OK (161 tests, 935 assertions)
```

PR #1001, opened 2017-12-19
Enhancement: Extract ViewAction from TalksController

localheinz:
This PR

* [x] extracts an `Admin\Talk\ViewAction` from the `Admin\TalksController`

Follows #912.

PR #996, opened 2017-12-18
Enhancement: Extract ViewAction from SpeakersController

localheinz:
This PR

* [x] extracts a `Reviewer\Speaker\ViewAction` from the `Reviewer\SpeakersController`

Follows #912.

PR #1005, opened 2017-12-19
Fix: Use TransactionalTestCase instead of removed RequiresDatabaseReset

localheinz:
This PR

* [x] uses `TransactionalTestCase` instead of `RequiresDatabaseReset`

Follows #959.
Follows #1001.
Follows #1004.

Blocks #996.
Blocks #1002.
Blocks #1003.

💁‍♂️ This should fix the broken build.


localheinz:
Thank you, @mdwheele!

PR #1007, opened 2017-12-20
Load Phinx configuration from Symfony container depending on environment

mdwheele:
This PR

* [x] Updates Phinx configuration to use `phinx.php` instead of `phinx.yml`.
* [x] Boots an instance of the `Kernel` into Phinx runtime; environment specified by `CFP_ENV` (`development` by default).
* [x] Queries database details from container parameters to dynamically generate Phinx configuration.
* [x] Updates `.travis.yml` configuration to use the environment variable rather than `-e testing`.
* [x] Updates `README.md`

This cuts down one configuration file we have to manage. The only trade-off is that we can't use `-e testing` option for `phinx migrate`. Instead, we should execute something like `CFP_ENV=production vendor/bin/migrate`. 

I like this change because it makes getting this up-and-running with Docker a little easier and also makes Phinx feel a little less "out of band" for first-time setup. I shouldn't have to enter the same database credentials twice *just* to get started. With this, I configure whatever environment I need, and it all kinda just works.

Inb4 TravisCI kills me...


mdwheele:
I have slain the Travis beast. 💪

mdwheele:
@localheinz You can ask me to make changes...

localheinz:
@mdwheele 

Ha, sorry!

README.md
```diff
@@ -229,13 +229,14 @@ mail:
 
 ### [Run Migrations](#run-migrations)
 
-This project uses [Phinx](http://phinx.org) to handle migrations. Be sure to copy the `phinx.yml.dist` file that is in the
-root directory for the project to `phinx.yml` and edit it to match your own database settings.
+This project uses [Phinx](http://phinx.org) to handle migrations. Configuration for Phinx is loaded from `phinx.php`.
```
localheinz: Can we also link to `phinx.php` here?

phinx.php
```diff
@@ -0,0 +1,38 @@
...
+ * the LICENSE file that was distributed with this source code.
+ *
+ * @see https://github.com/opencfp/opencfp
+ */
+
+use OpenCFP\Kernel;
+
+require_once __DIR__ . '/vendor/autoload.php';
+
+$kernel = new Kernel((string) \OpenCFP\Environment::fromServer($_SERVER), false);
```
localheinz: What do you think about importing `OpenCFP\Environment` in order to avoid unnecessary FQCNs?

phpunit.xml.dist
```diff
@@ -16,6 +16,7 @@
     <php>
         <!-- https://github.com/symfony/console/blob/v3.3.12/Terminal.php#L24-L36 -->
         <env name="COLUMNS" value="120"/>
+        <server name="CFP_ENV" value="testing"/>
         <ini name="memory_limit" value="512M"/>
```
localheinz: Can we keep these ordered alphabetically?

This has the advantage that it will be more obvious where to add new configuration items here.

.travis.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         - mysql -e "CREATE DATABASE $TRAVIS_DB" -uroot
         - cp phinx.yml.dist phinx.yml
```
localheinz: Since this file has been removed, can we remove this line and all other occurrences?

Running

```
$ git grep phinx.yml
```

yields

```
.gitignore:25:phinx.yml
.travis.yml:68:        - cp phinx.yml.dist phinx.yml
README.md:374:2. Copy the default `phinx.yml.dist` to `phinx.yml`
script/bootstrap:23:if [ ! -f "phinx.yml" ]; then
script/bootstrap:24:    echo "Missing phinx.yml, be sure to copy from phinx.yml.dist and update config"
script/test:14:vendor/bin/phinx --configuration=phinx.yml migrate -e testing
```

phpunit.xml.dist
mdwheele: Order the variable names? Or the XML element names?

```xml
<env name="COLUMNS" value="120"/>
<ini name="memory_limit" value="512M"/>
<server name="CFP_ENV" value="testing"/>
```

vs

```xml
<server name="CFP_ENV" value="testing"/>
<env name="COLUMNS" value="120"/>
<ini name="memory_limit" value="512M"/>
```

localheinz: I'd suggest to order the element names, or even better, just use the https://plugins.jetbrains.com/plugin/5919-lines-sorter plugin for PHPStorm (if you use it), and then press <kbd>Option</kbd> + <kbd>Shift</kbd> + <kbd>L</kbd>.

localheinz:
@mdwheele 

This looks good to me, but there's one thing I find a bit irritating: when running

```
$ make
```

then we can see

```
CFP_ENV=testing vendor/bin/phinx migrate
Phinx by CakePHP - https://phinx.org. 0.8.1

using config file ./phinx.php
using config parser php
using migration paths
 - /Users/am/Sites/opencfp/opencfp/migrations
warning no environment specified, defaulting to: testing
```

What do you think about applying the following patch to avoid the irritation:

```diff
diff --git a/.travis.yml b/.travis.yml
index 3b7ac0f1..74e928eb 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,7 @@ jobs:
         - sh .travis/setup-mail.sh
         - mysql -e "CREATE DATABASE $TRAVIS_DB" -uroot
         - cp config/testing.yml.dist config/testing.yml
-        - CFP_ENV=testing vendor/bin/phinx migrate
+        - CFP_ENV=testing vendor/bin/phinx migrate --environment testing
         - mysqldump -uroot $TRAVIS_DB > tests/dump.sql
         - if [[ "$WITH_COVERAGE" == "true" ]]; then code-climate-before-script; fi

diff --git a/Makefile b/Makefile
index 56e9ce15..005404fc 100644
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ cs: composer
 database: composer
        mysql -uroot -e "DROP DATABASE IF EXISTS cfp_test"
        mysql -uroot -e "CREATE DATABASE cfp_test"
-       CFP_ENV=testing vendor/bin/phinx migrate
+       CFP_ENV=testing vendor/bin/phinx migrate --environment testing
        mysqldump -uroot cfp_test > tests/dump.sql

 infection: composer database
```

mdwheele:
> What do you think about applying the following patch to avoid the irritation

@localheinz 

I've specified `--environment` to avoid the extra output.

localheinz:
Thank you, @mdwheele!

PR #1008, opened 2017-12-20
Fix: Consistently use dist extension

localheinz:
This PR

* [x] consistently uses the `.dist` extension for distributable configuration files


BackEndTea:
Using ```.dist.yml``` makes sure that IDEs etc know it is a yaml file, without having to specifically configure it. Using ```.yml.dist``` causes an IDE to not know what type of file it is.

So personally i'd prefer to not end in ```.dist```

localheinz:
@BackEndTea 

We can easily associate `.yml.dist` with YAML file type. Besides, we are using `phinx.yml.dist` and `phpunit.xml.dist` as well.

BackEndTea:
https://github.com/opencfp/opencfp/blob/589a84a3f641dc52da0e95853cb0c4e35f358c30/script/bootstrap#L28-L31

Could you update this as well

localheinz:
Nice catch, @BackEndTea!

BackEndTea:
Just waiting for travis to finish

localheinz:
Thank you, @BackEndTea!

PR #1011, opened 2017-12-20
Fix: remove server.php

BackEndTea:
This PR

* [x] Removes ```server.php```

Follows #992 

Now that we use the symfony web server there is no real reason to keep this file.


BackEndTea:
Thanks @chartjes

PR #980, opened 2017-12-18
Add the standard Symfony console to the project

derrabus:
This PR add the standard Symfony console to the project and lets it take over the `cache:clear` command.

Please test if that works out better for you.

The console has some helpful maintenance and debug commands, so it's probably good to have it from a developer perspective. We could merge the console with the `bin/opencfp` console, but I'm not sure if we should flood the end-user with framework-related commands. So it might be good to have to separate console applications.

Fixes #979. (kind of)

PR #1012, opened 2017-12-20
Removed old unused console application

derrabus:
This PR removes the old console application. Since we're using the standard Symfony console now, we don't need it anymore.

Follows #980.


localheinz:
Thank you, @derrabus!

PR #1009, opened 2017-12-20
Enhancement: Extract UpdatePasswordAction from ForgotController

localheinz:
This PR

* [x] extracts an `UpdatePasswordAction` from the `ForgotController`

Follows #912.

PR #1013, opened 2017-12-20
Fix: Rename class

localheinz:
This PR

* [x] renames a class

Similar to #982.
Blocks #1009.


classes/Http/Form/ResetFormType.php
```diff
@@ -19,7 +19,7 @@
 use Symfony\Component\Form\Extension\Core\Type\RepeatedType;
 use Symfony\Component\Form\FormBuilderInterface;
 
-class ResetForm extends AbstractType
+class ResetFormType extends AbstractType
```
localheinz: As the abstract class suggests, this isn't actually a form, but rather a form type, a definition of a form.

localheinz:
Thank you, @BackEndTea!

PR #1014, opened 2017-12-20
Fix: Rename argument

localheinz:
This PR

* [x] renames an argument (along with a field)

Blocks #1009.

### DIFF
--- a/resources/config/services/services.yml
+++ b/resources/config/services/services.yml
@@ -21,7 +21,7 @@ services:
       $applicationEmail: '%application.email%'
       $applicationEndDate: '%application.enddate%'
       $applicationTitle: '%application.title%'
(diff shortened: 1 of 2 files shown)


resources/config/services/services.yml
```diff
@@ -21,7 +21,7 @@ services:
       $applicationEmail: '%application.email%'
       $applicationEndDate: '%application.enddate%'
       $applicationTitle: '%application.title%'
-      $resetForm: '@OpenCFP\Http\Form\ForgotForm'
```
localheinz: The problem here is that we want to add a binding for the *actual* `$resetForm` in #1009, and as the name of the form type suggests, this is actually a forgot form (even if that name is a bit awkward).

localheinz:
Thank you, @chartjes!

PR #1015, opened 2017-12-20
Fix: Deprecate abstract Form class

localheinz:
This PR

* [x] marks the `abstract` `Form` class as deprecated

💁‍♂️ We should probably use Symfony forms instead.


localheinz:
Thank you, @mdwheele!

PR #1017, opened 2017-12-21
Since adding flash before default content, we do not need doubles

mdwheele:
This PR resolves this 🌶 

image

We recently added an include for the flash template before content on all templates extending `default.twig`. There were a few forgotten usages of flash that caused issues like this. I `grep`-ed for occurrences of `_flash.twig` and removed what made sense.\

Follows #900.


localheinz:
Thank you, @mdwheele!

PR #1018, opened 2017-12-21
Behat Tests

mdwheele:
This PR

* [x] Adds new dependencies for Behat, MinkExtension, Symfony2Extension, Selenium (i.e. all the things needed to do kernel and selenium based testing)
* [x] Configures two sessions: default (symfony2 kernel) and javascript (selenium-based) 
* [x] Tests that the homepage is visible without internal server errors
* [x] Tests success and form-validation scenarios for speaker signup

The point of this PR, overall, is to create a foundation for future work. That includes:

- [ ] Dealing with fixtures / database refresh in a sane way.
- [ ] Dealing with javascript-only features in an easy way (allow scenarios to be tagged as `@javascript`-only).
- [ ] Exemplar Gherkin that is feature-focused, not strictly browser-based in verbiage. 

❗️ Blocked by #1007 (will be able to simplify this PR a bit)


mdwheele:
The current state of this PR is "just enough wiring to get a passing suite". I'm going to spend some time adding more scenarios and fleshing out the `FeatureContext` with some ideas that will shape the expressiveness of Gherkin we can write. The Kernel driver... is wow...

BackEndTea:
👍 For the new test options.

tests/Features/signup.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Signup
+
+    Scenario: Speaker successfully signs up
+        Given I am on the signup page
+        When I fill in the following:
+            | email | speaker@example.com |
```
mheki: @mdwheele cool stuff, just one small note from the readability point of view.
You can add  more spaces in these "table" columns so it's easier to follow:
```
| email    | speaker@example.com |
| password | speaker             |

**Comment by BackEndTea** [comment]
@mdwheele 
Is this still WIP? and if so, what is still needed for this to be merged?

**Comment by mdwheele** [comment]
Yeah. The rest of it is on a laptop with a dead battery. I picked up the wrong charger before holiday and wot be back to the office until the 2nd. I'll send the rest then and we should be good to review. I've limited scope to Kernel driver for now because I don't have a great solution for rolling back the rest database (that isn't just using a dump file, which I'd like to get away from)

File: tests/Features/signup.feature (hunk 1 above)
Comment (mdwheele) [reply tests/Features/signup.feature]: Thanks @mheki! I ended up cooking this into the language itself rather than using the stock matchers from `MinkContext` (e.g. `When I fill in my account information`) captures the heart of behaviour without unnecessary detail

**Comment by mdwheele** [comment]
@BackEndTea 

This is ready for review.

File: tests/Features/Contexts/FeatureContext.php [hunk 2]
```diff
@@ -0,0 +1,216 @@
...
+     * @Given I am on the dashboard
+     */
+    public function iAmOnTheDashboard()
+    {
+        $this->visit('dashboard');
+    }
+
+    /**
+     * @When I submit a talk about :title
+     * @Given I have submitted a talk about :title
```
BackEndTea: Maybe this should be split up in two functions

`@When` for doing the action
`@Given` for simply updating the database

tests/Features/Contexts/FeatureContext.php
```diff
@@ -0,0 +1,216 @@
...
+     */
+    public function iSubmitATalkAbout($title)
```
BackEndTea: Can we add a type declaration here?

tests/Features/Contexts/FeatureContext.php
```diff
@@ -0,0 +1,216 @@
...
+     */
+    public function iShouldBeAbleToEditMyTalk()
+    {
+        $this->assertPageContainsText('Edit');
+    }
+
+    /**
+     * @When I delete the :title talk
+     */
+    public function iDeleteTheTalk($title)
```
BackEndTea: Can we add a type declaration here?

tests/Features/signup.feature
```diff
@@ -0,0 +1,13 @@
+Feature: A Speaker signs up
+
+    Scenario: Speaker successfully signs up
+        Given I am on the signup page
+        When I fill in my account information
+          And I agree to the code of conduct
+          And I press "Signup"
+        Then I should see "Welcome to OpenCFP!"
```
BackEndTea: I don't think we should test for seeing `OpenCFP` but rather something else that we can use to identify the page.

Since users updating OpenCFP for their own conference will get a failing test suite

tests/Features/homepage.feature
```diff
@@ -0,0 +1,6 @@
+Feature: Homepage
+
+    Scenario: It loads!
+        Given I am on the homepage
+        Then I should see "OpenCFP"
```
BackEndTea: I don't think we should test for seeing OpenCFP but rather something else that we can use to identify the page.

Since users updating OpenCFP for their own conference will get a failing test suite

tests/Features/Contexts/FeatureContext.php
```diff
@@ -0,0 +1,216 @@
...
+use Illuminate\Database\Capsule;
+use Illuminate\Database\Connection;
+use OpenCFP\Domain\Model\Talk;
+use OpenCFP\Domain\Model\User;
+use OpenCFP\Domain\Services\AccountManagement;
+use Phinx\Console\PhinxApplication;
+use Symfony\Component\Console\Input\StringInput;
+use Symfony\Component\Console\Output\NullOutput;
+
+final class FeatureContext extends MinkContext
```
BackEndTea: This is probably too much for this PR, but we should look into splitting up contexts where possible, so this class does not become too large to manage

.travis.yml
```diff
@@ -68,6 +68,7 @@ jobs:
         - if [[ "$WITH_COVERAGE" == "true" ]]; then xdebug-enable; fi
         - if [[ "$WITH_COVERAGE" == "true" ]]; then vendor/bin/phpunit --testsuite unit --coverage-clover build/logs/clover.xml; else vendor/bin/phpunit --testsuite unit; fi
         - if [[ "$WITH_COVERAGE" == "true" ]]; then xdebug-disable; fi
+        - vendor/bin/behat
```
BackEndTea: Can we add this command to the `Makefile` as well?
Something like
```make
behat: composer test-env
	vendor/bin/behat
```

And then also add it to the `make test` command

tests/Features/Contexts/FeatureContext.php
```diff
@@ -0,0 +1,216 @@
...
+    private function databaseConnection(): Connection
+    {
+        /** @var Capsule\Manager $capsule */
+        $capsule = $this->getContainer()->get(Capsule\Manager::class);
+
+        return $capsule->getConnection();
+    }
+
+    /**
+     * @Given I am on the signup page
```
BackEndTea: Maybe we can make these more generic into a 
`@Given i am on the :page page`

tests/Features/homepage.feature
```diff
@@ -0,0 +1,6 @@
...
+          And I should not see "Internal Server Error"
```
BackEndTea: Is there a way to automatically check this, and maybe for other errors as well, on every page visit?

tests/Features/Contexts/FeatureContext.php
mdwheele: I would recommend `SpeakerContext` and `AdminContext` and call it at that. I would prefer to make that the work of another PR as the point of this one was to get more-or-less a baseline.

tests/Features/Contexts/FeatureContext.php
mdwheele: My preference is to leave the step as-is because it describes a ubiquitous part of the application that *won't* change. The signup page will always be the signup page, but the URL could easily change. Maybe this could be parameterized to  reference a route name instead, but I certainly wouldn't want to tie Gherkin to details like a URL. Gherkin isn't the test, that's the step definition's purpose.

tests/Features/homepage.feature
mdwheele: We could change it to something other than OpenCFP, but my personal opinion is that this is *our* test suite dependent on values from *our* `testing.yml`. This "feature" is arguably a candidate for removal from the suite, in my opinion.

tests/Features/homepage.feature
mdwheele: I would stick with positive assertions. If you get an ISE, it's very likely a step definition for the affected scenario will fail. Additionally, we shouldn't check for such things in a "Homepage" feature (which I may remove instead).

I think adding an "Error Handling" feature that covers things like our 404, 500 handlers would be reasonable; especially if we're going to be improving those (e.g.: 500 error because "you forgot to configure the app" versus "uncaught exception", etc)

tests/Features/signup.feature
mdwheele: See my comments above about testing for presence of "OpenCFP".

As an alternative, the step could be `@Then I am greeted as a first-time speaker` that has a step definition that checks presence of message (using config from Kernel) and also checks that the speaker is prompted to fill out the rest of their profile. I think I like this better. Thoughts?

BackEndTea: `@Then I am greeted as a first-time speaker` looks like a good idea to me.

PR #1021, opened 2017-12-21
Enhancement: Turn TalkController into CreateProcessAction

localheinz:
This PR

* [x] turns the `TalkController` into a `CreateProcessAction`

### DIFF
--- a/classes/Http/Action/Talk/CreateProcessAction.php
+++ b/classes/Http/Action/Talk/CreateProcessAction.php
@@ -11,30 +11,29 @@ declare(strict_types=1);
  * @see https://github.com/opencfp/opencfp
  */
 
-namespace OpenCFP\Http\Controller;
+namespace OpenCFP\Http\Action\Talk;
 
 use HTMLPurifier;
 use OpenCFP\Domain\CallForPapers;
-use OpenCFP\Domain\Model\Talk;
-use OpenCFP\Domain\Services\Authentication;
-use OpenCFP\Http\Form\TalkForm;
-use OpenCFP\Http\View\TalkHelper;
+use OpenCFP\Domain\Model;
+use OpenCFP\Domain\Services;
+use OpenCFP\Http\Form;
+use OpenCFP\Http\View;
 use Swift_Mailer;
 use Swift_Message;
-use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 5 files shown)


classes/Http/Action/Talk/CreateProcessAction.php
```diff
@@ -133,64 +129,76 @@ public function processCreateAction(Request $request): Response
             'sponsor'     => $request->get('sponsor'),
             'user_id'     => $request->get('user_id'),
         ]);
-        $form->sanitize();
 
-        if ($form->validateAll()) {
```
localheinz: Inverted the condition to make it more obvious what the happy path is.

classes/Http/Action/Talk/CreateProcessAction.php
```diff
@@ -88,40 +97,27 @@ public function __construct(
...
 
-            return $this->redirectTo('dashboard');
+            $url = $this->urlGenerator->generate('dashboard');
+
+            return new HttpFoundation\RedirectResponse($url);
         }
 
         $user = $this->authentication->user();
 
-        $form = $this->getTalkForm([
```
localheinz: Renamed this method as it actually creates a form.

classes/Http/Action/Talk/CreateProcessAction.php
```diff
@@ -88,40 +97,27 @@ public function __construct(
...
+        $form = $this->createTalkForm([
```
localheinz: As the form is created here, we can't write unit tests that go further than the condition above.

BackEndTea:
Currently trying to submit a talk on your branch gives an ```Swift_TransportException``` When submitting a talk. However this also occurs on master, so i'm not sure if this is my fault or that master is broken

localheinz:
@BackEndTea 

Taking a look!

localheinz:
@BackEndTea 

I could reproduce the issue in `master`, but it can be fixed by properly configuring the mail adapter.

BackEndTea:
@localheinz 

I don't think that an error like this is intended behavior though. For example, how will the application respond if the email server is down?

localheinz:
@BackEndTea 

Ideally, emails should be send asynchronously, but I think this is out of scope of this PR and should be handled separately.

localheinz:
Thank you, @BackEndTea!

PR #1022, opened 2017-12-21
Enhancement: Turn SignupController into ProcessAction

localheinz:
This PR

* [x] turns the `SignupController` into a `ProcessAction`

Follows #912.


classes/Http/Action/Signup/ProcessAction.php
```diff
@@ -0,0 +1,142 @@
...
+    ) {
+        $this->authentication = $authentication;
+        $this->accounts       = $accounts;
+        $this->urlGenerator   = $urlGenerator;
+    }
+
+    public function __invoke(HttpFoundation\Request $request): HttpFoundation\Response
+    {
+        try {
+            $this->validate($request, [
```
localheinz: Can't write any unit tests as the validation facilities are created within this class.

localheinz:
Thank you, @chartjes!

Issue #1019, opened 2017-12-21
RFC: move all scripts to composer scripts


BackEndTea:
Currently there are multiple places where scripts for running tests etc are defined.

* ```Makefile```
* ```script/```
* ```.travis.yml```

I'd propose to define them in composer scripts, so we can run ```composer test``` or ```composer cs``` etc. 
We could then delete the ```Makefile```,  `script/`, and update `.travis.yml` to use the composer scripts. And of course update documentation. 

This has the following advantages
* Less files to maintain
* Allows windows users to use the commands (`make` isn't really a windows thing)
* Gives a bit more security that if tests pass on your system it will also pass on travis.

localheinz:
I don't care much whether we use `Makefile` or `composer `scripts as long as I have short and simple commands available to build locally.

In regard to some duplication between `.travis.yml` and `Makefile`: I can think of a single reason why that is fine, and that is that it is easier to see which step failed when there are  multiple steps defined (rather than `composer make` or `make`), and therefore I think it's worth paying the price of duplication.

I'm totally 👍 for replacing `/scripts` with either targets in `Makefile`, or `composer` scripts.

Maybe it's a good idea to do this in multiple steps?

For a start, we could move everything in `/scripts` into the `Makefile`, and then revisit.

For reference, also see http://fabien.potencier.org/symfony4-best-practices.html.

mdwheele:
I'm 100% on consolidating to one and making Travis do *exactly* what we do to set up.

That said, the current `Makefile` makes a assumptions about the environment and cannot be used for anything other than a local development setup. I run OpenCFP in a Docker environment that separates Selenium, MariaDB, Nginx and PHP-FPM into different containers. Current setup process assumes database is available via `localhost`, etc. Leads to me just doing my own thing; extra work. 

I would prefer something that is able to be parametrized (possibly via environment variables) which could be done no matter the direction; just voicing the concern. I am not suggesting that is within scope here, but is worth considering when picking a solution (i.e. it needs to be configurable without editing targets by hand).

For information, the `/scripts` concept was originally introduced based on standards GitHub adopted in their projects (https://github.com/github/scripts-to-rule-them-all). There *is* an argument to be made in having explicated ubiquitous targets for doing executing common application procedures. Pretty sure that can be done in a `Makefile` as well.

> I'd propose to define them in composer scripts, so we can run composer test or composer cs etc.

[...]

PR #1024, opened 2017-12-21
Fix: set up test config file before running database

BackEndTea:
This PR

* [x] fixes an issue with integration tests not running after upgrading to the latest version

Before this we tried to set up the database without a proper `testing.yml` file being present, and thus it would error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ coverage: composer
 cs: composer
 	vendor/bin/php-cs-fixer fix --verbose --diff
 
-database: composer
+database: test-env composer
 	mysql -uroot -e "DROP DATABASE IF EXISTS cfp_test"
 	mysql -uroot -e "CREATE DATABASE cfp_test"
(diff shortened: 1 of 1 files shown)


mdwheele:
Good catch @BackEndTea! This is the type of stuff that I hope will be resolved by #1019, no matter if we go `Makefile`, `composer` scripts or `scripts/*`.

Makefile
```diff
@@ -24,11 +24,13 @@ database: composer
 infection: composer database
 	vendor/bin/infection
 
-integration: composer database
-	if [ ! -f "config/testing.yml" ]; then cp config/testing.yml.dist config/testing.yml; fi
```
localheinz: How about removing the `test-env` target and instead moving this line to the `database` target instead?

BackEndTea: I had considered that, but i feel that setting up the test config file should not be explicitly part of `make database`.

mdwheele: I'm fine with it as-is. It's all going to be reconsidered for #1019 anyway.

PR #1025, opened 2017-12-21
Enhancement: Extract LogOutAction from SecurityController

localheinz:
This PR

* [x] extracts  a `LogOutAction` from the `SecurityController`

Follows #912.


localheinz:
Thank you, @chartjes!

PR #1026, opened 2017-12-21
Enhancement: Extract ChangePasswordProcessAction from ProfileController

localheinz:
This PR

* [x] extracts ` ChangePasswordProcessAction` from the `ProfileController`

Follows #912.

### DIFF
--- a/classes/Http/Action/Profile/ChangePasswordProcessAction.php
+++ b/classes/Http/Action/Profile/ChangePasswordProcessAction.php
@@ -1,0 +1,106 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Copyright (c) 2013-2017 OpenCFP
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ *
+ * @see https://github.com/opencfp/opencfp
+ */
+
+namespace OpenCFP\Http\Action\Profile;
+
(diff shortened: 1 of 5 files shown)


classes/Http/Action/Profile/ChangePasswordProcessAction.php
```diff
@@ -0,0 +1,106 @@
...
+        /**
+         * Okay, the logic is kind of weird but we can use the SignupForm
+         * validation code to make sure our password changes are good
+         */
+        $formData = [
+            'password'  => $request->get('password'),
+            'password2' => $request->get('password_confirm'),
+        ];
+
+        $form = new Form\SignupForm(
```
localheinz: Because we are creating the form here instead of injecting it, we cannot write any unit tests.

localheinz:
@BackEndTea 

My take here is that I'm just extracting actions, and reuse existing integrations tests as far as they are available, and add only unit tests. I added this single integration test here only because there hasn't been an integration test yet for the action at all.

I don't mind adding integration tests, though.

localheinz:
@BackEndTea 

Added a few more tests, but can't add a test for the success case because of how we currently mock the authentication. Happy to modify this, though, also see #1037.

localheinz:
Thank you, @BackEndTea!

PR #1027, opened 2017-12-21
Fix: Add test-env to phony targets

localheinz:
This PR

* [x] adds `test-env` as a phony target

Follows #1024.


localheinz:
Thank you, @BackEndTea!

PR #1028, opened 2017-12-21
Enhancement: Extract ResetProcessAction from ForgotController

localheinz:
This PR

* [x] extracts a `ResetProcessAction` from the `ForgotController`

Follows #912.

### DIFF
--- a/tests/Integration/Http/Action/Forgot/ResetProcessActionTest.php
+++ b/tests/Integration/Http/Action/Forgot/ResetProcessActionTest.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Copyright (c) 2013-2017 OpenCFP
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ *
+ * @see https://github.com/opencfp/opencfp
+ */
+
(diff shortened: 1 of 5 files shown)


tests/Integration/Http/Action/Forgot/ResetProcessActionTest.php
```diff
@@ -0,0 +1,44 @@
...
+final class ResetProcessActionTest extends WebTestCase implements TransactionalTestCase
+{
+    public function testRendersResetPasswordFormIfFormIsNotSubmitted()
+    {
+        $resetCode = $this->faker()->sha256;
+
+        /** @var Model\User $user */
+        $user = factory(Model\User::class)->create()->first();
+
+        $client = $this->createClient();
```
localheinz: Moving forward, shall we use the client directly, or the methods `post()`, `get()`, etc?

chartjes: Hrm, I wonder if it simply comes down to personal preference UNLESS we want things to be consistent. If we're not using post() or get() elsewhere, we should stick with that.

Personally I'm okay with `$client->request()`

localheinz:
Thank you, @chartjes!

PR #1029, opened 2017-12-21
Fix: Reduce visibility and mark as final

localheinz:
This PR

* [x] reduces the visibility of `public` methods in `WebTestCase` to `protected`, and marks most `protected` methods as `final`

### DIFF
--- a/tests/Integration/WebTestCase.php
+++ b/tests/Integration/WebTestCase.php
@@ -84,7 +84,7 @@ abstract class WebTestCase extends KernelTestCase
         parent::tearDown();
     }
 
-    protected static function getKernelClass()
+    final protected static function getKernelClass()
     {
         return Kernel::class;
     }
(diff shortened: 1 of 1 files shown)


tests/Integration/WebTestCase.php
```diff
@@ -115,12 +115,12 @@ private static function runBeforeClassTraits()
         }
     }
 
-    protected function createClient(): Client
+    final protected function createClient(): Client
     {
         return $this->container->get('test.client');
     }
 
-    public function call(string $method, string $uri, array $parameters = [], array $cookies = [], array $files = [], array $server = [], string $content = null): Response
```
localheinz: None of these methods need to be `public`.

tests/Integration/WebTestCase.php
```diff
@@ -115,12 +115,12 @@ private static function runBeforeClassTraits()
         }
     }
 
-    protected function createClient(): Client
+    final protected function createClient(): Client
```
localheinz: Hardly any of these `protected` methods are intended to be overridden by extending classes.

localheinz:
Thank you, @chartjes!

PR #1030, opened 2017-12-21
Fix: Use TransactionalTestCase

localheinz:
This PR

* [x] uses the `TransactionalTestCase` interface instead of the `RefreshDatabase` trait
* [x] removes the `RefreshDatabase` trait

PR #1031, opened 2017-12-21
Fix: Refresh database before and after class

localheinz:
This PR

* [x] refreshes the database before **and** after the class

Blocks #1030.

💁‍♂️ I have been scratching my head because tests in #1030 are failing. Then it dawned on me that of course, when we have integration tests run from a test class that uses the `RefreshDatabase` trait, and the next test class does not use the `RefreshDatabase` trait, but instead implements the `TransactionalTestCase`, then the database won't be fresh in the latter, and tests might fail (as they currently do in #1030). So unfortunately, until we remove the `RefreshDatabase` trait (working on it), we'll have to refresh the database twice.

### DIFF
--- a/tests/Integration/WebTestCase.php
+++ b/tests/Integration/WebTestCase.php
@@ -63,7 +63,12 @@ abstract class WebTestCase extends KernelTestCase
 
     public static function setUpBeforeClass()
     {
-        self::runBeforeClassTraits();
+        self::runBeforeAndAfterClassTraits();
+    }
+
+    public static function tearDownAfterClass()
+    {
+        self::runBeforeAndAfterClassTraits();
     }
 
     protected function setUp()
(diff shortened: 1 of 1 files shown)


mdwheele:
👍 Are you focused on removing `RefreshDatabase` and want to continue down that path? I could help out some tonight. Just don't want to submit the same work twice :) If it's something we're hitting-as-we-go, perhaps an issue to revert the diff here (once `RefreshDatabase` is gone) would be worthwhile?

localheinz:
@mdwheele 

Probably going to bed soon, but thanks a lot! Let me push some of the changes in a moment, then you can take over if you like, what do you think?

mdwheele:
@localheinz 

👍 Sounds good!

localheinz:
@mdwheele 

Pushed some changes into #1030, feel free to take over!

Good night! 🌝

localheinz:
Thank you, @chartjes and @mdwheele!

PR #1034, opened 2017-12-22
Fix: use assets path

BackEndTea:
This PR

* [x] Uses the assets function in the views, to make this easier to configure


localheinz:

PR #1035, opened 2017-12-22
Clean and split up Path class

BackEndTea:
This PR

* [x] splits the path class in two
* [x] removes all unused functions
 
Blocked by #1034


localheinz:
Thank you, @BackEndTea!

PR #1036, opened 2017-12-22
Fix: Increase memory limit

localheinz:
This PR

* [x] increases the memory limit for running tests

Blocks #1030.

💁‍♂️ Running tests fails there because the current memory limit of `512M` is exceeded a bit.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -16,7 +16,7 @@
     <php>
         <!-- https://github.com/symfony/console/blob/v3.3.12/Terminal.php#L24-L36 -->
         <env name="COLUMNS" value="120"/>
-        <ini name="memory_limit" value="512M"/>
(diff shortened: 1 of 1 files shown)


mdwheele:
> Gonna accept this for now, but i think we should check what is causing us to use this much memory during tests.

Hahaha, yes. I can tell you that the static methods on `RefreshDatabase` are a source of it. Since we're moving in the `TransactionalTestCase` direction, perhaps that will be alleviated. Otherwise, I bet the "easiest" thing to do would be to set up blackfire.io for some test runs and go from there. The Symfony Profiler might also be good enough for what we need. I've never used it. 

I've been in an xdebug + qcachegrind wasteland for most of my life.

localheinz:
Thank you, @BackEndTea!

PR #1037, opened 2017-12-22
Fix: Don't assume a user identifier

localheinz:
This PR

* [x] removes default values for identifiers when setting up signed in users in tests


localheinz:
Thank you, @chartjes!

PR #1039, opened 2017-12-22
Fix: Rename classes

localheinz:
This PR

* [x] renames two classes

### DIFF
--- a/tests/Helper/MockAuthentication.php
+++ b/tests/Helper/MockAuthentication.php
@@ -17,7 +17,7 @@ use OpenCFP\Domain\Services\Authentication;
 use OpenCFP\Domain\Services\NotAuthenticatedException;
 use OpenCFP\Infrastructure\Auth\UserInterface;
 
-final class MockableAuthenticator implements Authentication
+final class MockAuthentication implements Authentication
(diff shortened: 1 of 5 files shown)


tests/Helper/MockAuthentication.php
```diff
@@ -17,7 +17,7 @@
 use OpenCFP\Domain\Services\NotAuthenticatedException;
 use OpenCFP\Infrastructure\Auth\UserInterface;
 
-final class MockableAuthenticator implements Authentication
```
localheinz: The actual implementation of the `Authentication` interface is called [`SentinelAuthentication`](https://github.com/opencfp/opencfp/blob/844710aa1ea60a761696007741c7d703fdf89ecf/classes/Infrastructure/Auth/SentinelAuthentication.php), so `Authentication` here makes more sense, too.

Also, it's not actually `Mockable`, as the class itself is `final`, so `MockAuthentication` would be more appropriate, too, I think.

localheinz:
@chartjes 

Should be fixed in a few minutes!

localheinz:
Thank you, @chartjes!

PR #1040, opened 2017-12-22
Enhancement: added "pretty print" helper

chartjes:
This PR

* [x] Adds a 'pretty printer' for PHPUnit results to allow better visibility into what tests are being run


localheinz:
Thank you, @chartjes!

PR #1041, opened 2017-12-22
Enhancement: Turn SecurityController into LogInAction

localheinz:
This PR

* [x] turns the `SecurityController` into a `LoginAction`

Follows #912.


localheinz:
Thank you, @chartjes!

PR #1045, opened 2017-12-22
Enhancement: Extract RateAction from TalksController

localheinz:
This PR

* [x] extracts an `Admin\Talk\RateAction` from the `Admin\TalksController`

Follows #912.


classes/Http/Controller/Admin/TalksController.php
```diff
@@ -36,23 +35,6 @@ public function __construct(TalkHandler $talkHandler, Twig_Environment $twig, Ur
         parent::__construct($twig, $urlGenerator);
     }
 
-    public function rateAction(Request $request): Response
-    {
-        try {
-            $this->validate($request, [
```
BackEndTea: It looks like this was the last place that used the validate functio in the BaseController. Maybe we can remove it from the BaseController now

tests/Unit/ProjectCodeTest.php
```diff
@@ -47,6 +47,7 @@ public function testProductionClassesHaveUnitTests()
                 Http\Action\Admin\Talk\IndexAction::class,
                 Http\Action\Reviewer\DashboardAction::class,
                 Http\Action\Reviewer\Speaker\ViewAction::class,
+                Http\Action\Admin\Talk\RateAction::class,
```
BackEndTea: It looks like this action won't be too hard to unit test, maybe we can add some

classes/Http/Action/Admin/Talk/RateAction.php
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    public function __invoke(HttpFoundation\Request $request): HttpFoundation\Response
+    {
+        try {
+            $this->validate($request, [
+                'rating' => 'required|integer',
+            ]);
+        } catch (ValidationException $exception) {
+            return new HttpFoundation\Response();
```
derrabus: I know, this is the same behavior as in the controller, but I was thinking, we should maybe send a `400` status here instead of `200`. WDYT?

classes/Http/Action/Admin/Talk/RateAction.php
```diff
@@ -0,0 +1,89 @@
...
+     * @param array                  $messages
+     * @param array                  $customAttributes
+     *
+     * @throws ValidationException
+     */
+    private function validate(HttpFoundation\Request $request, $rules = [], $messages = [], $customAttributes = [])
+    {
+        $data = $request->query->all() + $request->request->all() + $request->files->all();
+
+        $validation = new Validation\Factory(
```
derrabus: I think, this should be a service.

classes/Http/Action/Admin/Talk/RateAction.php
```diff
@@ -0,0 +1,89 @@
...
+            $rules,
+            $messages,
+            $customAttributes
+        );
+
+        if ($validator->fails()) {
+            throw ValidationException::withErrors(array_flatten($validator->errors()->toArray()));
+        }
+
+        unset($validation, $validator);
```
derrabus: No need to unset anything because the scope of these variables is about to die.

tests/Unit/ProjectCodeTest.php
localheinz: Ha, I'd prefer to test this when we have extracted the validator into a service, otherwise I'd just be testing the validator mostly.

classes/Http/Action/Admin/Talk/RateAction.php
derrabus: Agreed.

classes/Http/Action/Admin/Talk/RateAction.php
localheinz: Just copied from the `BaseController`, it's exactly the same in `Reviewer\Talk\RateAction`. Will tackle this later, @derrabus.

BackEndTea:
Im on my phone so i cant merge this atm.

localheinz:
Thank you, @BackEndTea!

PR #1048, opened 2017-12-23
Enhancement: Extract PromoteAction from SpeakersController

localheinz:
This PR

* [x] extracts an `Admin\Speaker\PromoteAction` from the `Admin\SpeakersController`

Follows #912.


classes/Http/Action/Admin/Speaker/PromoteAction.php
```diff
@@ -0,0 +1,79 @@
...
+            }
+
+            $this->accountManagement->promoteTo($user->getLogin(), $role);
+
+            $request->getSession()->set('flash', [
+                'type'  => 'success',
+                'short' => 'Success',
+                'ext'   => '',
+            ]);
+        } catch (\Exception $e) {
```
localheinz: This is a rather large `try`-and-`catch` block, and I would suggest to split it up in a separate PR, where we then catch specific exceptions. The unit tests added in this PR already reflect the different cases, only need adjustments in regard to flash messages.

The unit tests added here already reflect the different exect

localheinz:
Thank you, @chartjes!

PR #1050, opened 2017-12-23
Fix: Split try-and-catch block

localheinz:
This PR

* [x] splits a `try`-and-`catch` block

Follows https://github.com/opencfp/opencfp/pull/1048#discussion_r158587123.


localheinz:
Thank you, @chartjes!

PR #1051, opened 2017-12-24
Fix: No need to install SQLite on Travis

localheinz:
This PR

* [x] removes configuration to install SQLite on Travis


localheinz:
Thank you, @BackEndTea!

PR #1053, opened 2017-12-28
Removed Silex-compatible service aliases

derrabus:
This PR removes Silex-compatible service aliases.

In order to keep the diff of #895 low, I had added two service aliases, so I didn't have to adjust too many test cases when switching between Silex and Symfony. Now that Silex is gone for good, we can safely remove those aliases and use the Symfony service IDs in those tests.

Follows #895.


localheinz:
Thank you, @derrabus!

PR #1054, opened 2017-12-28
Upgraded Phinx to 0.9.2

derrabus:
This PR upgrades Phinx to 0.9.2. The main difference for us is that the new release is compatible with Symfony 4.

Related to #618.


localheinz:
Thank you, @derrabus!

PR #1055, opened 2017-12-28
Upgraded Twig to version 2

derrabus:
This PR upgrades Twig to the 2.x branch. Hello future! 😎

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
     },
     "require": {
         "php": ">=7",
-        "aptoma/twig-markdown": "^2",
+        "aptoma/twig-markdown": "^3.0.0",
         "cartalyst/sentinel": "^2.0.17",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -41,7 +41,6 @@
         "symfony/validator": "^3.4",
         "symfony/yaml": "^3.4",
         "twig/extensions": "^1.5",
-        "twig/twig": "1.*",
```
localheinz: Can we make the dependency on `twig/twig` explicit?

derrabus: Sure, but we did more or less the same with Swiftmailer (depend on the bundle instead of bundle + library).

Do you want me to change it?

localheinz: I think it would be nice to keep it as explicit dependency.

localheinz:
Thank you, @derrabus!

PR #1056, opened 2017-12-28
Load WebServerBundle in development only

derrabus:
This PR moves the registration of `WebServerBundle` to the development environment section.

It should be possible to run the application in production environment with only production dependencies. This is currently not the case because the kernel tries to register the `WebServerBundle` which is a dev dependency.

Since you probably don't want to run the production system on php's built-in webserver, the bundle is now only available in the development environment.

Follows #992.


localheinz:
Thank you, @derrabus!

PR #1057, opened 2017-12-28
Use TwigBundle's error handler

derrabus:
This PR reduces the functionality of the `ExceptionListener` class, so HTML error pages are rendered by the exception handler provided by TwigBundle.

The advantage of this approach is that TwigBundle renders a helpful exception page with a stack trace in our development environment.

Reference: https://symfony.com/doc/3.4/controller/error_pages.html

For example, if I throw an exception in our `DashboardAction`, I see the following page:

**development**
bildschirmfoto 2017-12-28 um 21 44 36

**production**
bildschirmfoto 2017-12-28 um 21 44 23

By the way, we should reconsider the remaining logic of the `ExceptionListener` class. For JSON errors, it currently prints the exception message, which is probably information we would not want to hand out in production mode.


chartjes:
Could we have a better 500 error page to show to folks? What we have is terrible

classes/Infrastructure/Event/ExceptionListener.php
```diff
@@ -46,40 +45,25 @@ public function onKernelException(GetResponseForExceptionEvent $event)
...
+        $headers = [];
 
-        if (\array_key_exists($code, $templates)) {
-            $template = $templates[$code];
+        if ($exception instanceof HttpExceptionInterface) {
+            $code    = $exception->getStatusCode();
+            $headers = $exception->getHeaders();
         }
 
-        $message = $this->twig->render($template);
```
localheinz: Since we aren't using `$twig` here anymore, can we remove the field as well as the constructor?

See https://github.com/opencfp/opencfp/blob/04086a1bea1eb16816291aae7137603dacd15806/classes/Infrastructure/Event/ExceptionListener.php#L27-L35.

classes/Infrastructure/Event/ExceptionListener.php
```diff
@@ -46,40 +45,25 @@ public function onKernelException(GetResponseForExceptionEvent $event)
...
-            Response::HTTP_FORBIDDEN    => 'error/403.twig',
-            Response::HTTP_NOT_FOUND    => 'error/404.twig',
-        ];
+        $event->setResponse($this->renderResponse($exception));
+    }
 
-        $code = $exception instanceof HttpExceptionInterface ? $exception->getStatusCode() : 500;
+    private function renderResponse(\Throwable $exception): Response
+    {
+        $code    = 500;
```
localheinz: While it hasn't been used before, can we take the opportunity here to use the `HttpFoundation\Response::HTTP_INTERNAL_SERVER_ERROR` constant instead?

classes/Infrastructure/Event/ExceptionListener.php
```diff
@@ -46,40 +45,25 @@ public function onKernelException(GetResponseForExceptionEvent $event)
...
+            $code    = $exception->getStatusCode();
+            $headers = $exception->getHeaders();
         }
 
-        $message = $this->twig->render($template);
-
-        return new Response($message, $code);
+        return new JsonResponse([
+            'error' => $exception->getMessage(),
+        ], $code, $headers);
```
localheinz: Not a big deal, but at first glance it's not quite apparent that we actually pass three arguments to the constructor of `JsonResponse`, maybe we can adjust the wrapping a bit to make that more obvious?

```diff
diff --git a/classes/Infrastructure/Event/ExceptionListener.php b/classes/Infrastructure/Event/ExceptionListener.php
index a1581611..b51c1b2e 100644
--- a/classes/Infrastructure/Event/ExceptionListener.php
+++ b/classes/Infrastructure/Event/ExceptionListener.php
@@ -62,8 +62,12 @@ class ExceptionListener implements EventSubscriberInterface
             $headers = $exception->getHeaders();
         }

-        return new JsonResponse([
-            'error' => $exception->getMessage(),
-        ], $code, $headers);
+        return new JsonResponse(
+            [
+                'error' => $exception->getMessage(),
+            ],
+            $code,
+            $headers
+        );
     }
 }
```

tests/Integration/WebTestCase.php
```diff
@@ -94,7 +94,7 @@ protected function tearDown()
         return Kernel::class;
     }
 
-    private function refreshContainer()
+    protected function refreshContainer()
```
localheinz: This change and the change in `ExceptionListenerTest` appear to be necessary, but I have to admit they *smell* a bit. Not sure of how that could be improved at the moment, but maybe we can take a look some other time.

classes/Infrastructure/Event/ExceptionListener.php
derrabus: Good catch, I'm removing it.

tests/Integration/WebTestCase.php
derrabus: I agree.

localheinz:
Thank you, @derrabus!

PR #1058, opened 2017-12-29
Don't override the global app template variable

derrabus:
This PR removes our global `app` template variable.

The reason for this change is that Symfony uses that variable to provide access to certain objects like the environment, the current request or whether the application runs in debug mode. If we override it like we currently do, we cannot make use of that functionality.

Reference: https://symfony.com/doc/3.4/templating/app_variable.html

Furthermore, the information that is currently being queried from that variables is already redundantly present in another global variable, `site`. Our `app` contains more information than `site`, but I could not find a twig line where information from `app` is accessed that `site` doesn't also contain. So, from my point of view, if safe to redirect all access to the `site` variable and drop `app`.

### DIFF
--- a/classes/Kernel.php
+++ b/classes/Kernel.php
@@ -113,7 +113,6 @@ final class Kernel extends SymfonyKernel
 
     private function bindConfiguration(ContainerBuilder $container, array $config)
     {
-        $container->setParameter('config.all', $config);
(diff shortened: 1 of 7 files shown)


localheinz:
Thank you, @derrabus!

PR #1060, opened 2018-01-02
Fix: make infection 

BackEndTea:
This PR

* [x] updates the `make infection` command 

Due to the new printer infection was not working correctly, this PR forces phpunit to use the default printer for infection

Follows #1040.


localheinz:
Thank you, @BackEndTea!

PR #1062, opened 2018-01-02
Enhancement: Make cache

BackEndTea:
This PR

* [x] Adds the `make cache` command, which clears testing and development cache
* [x] Has `make integration` run the cache command before executing

Unit tests shouldn't interact with the cache, so no need to clear it before those, since it is quite slow to run these commands.

This should make sure we don't run our tests against an old cache


localheinz:
Thank you, @BackEndTea!

PR #1063, opened 2018-01-02
Fix: Only report coverage after success

BackEndTea:
This PR

* [x] Changes travis to only on a  success.


localheinz:
Thank you, @BackEndTea!

PR #1065, opened 2018-01-05
Removed unused Twig dependency

derrabus:
This PR removes the unused dependency on `Twig_Environment` from the `OpenCFP\Http\Action\Signup\ProcessAction` class.


derrabus:
Travis failure is unrelated.

PR #1043, opened 2017-12-22
Enhancement: Turn TalksController into RateAction

localheinz:
This PR

* [x] turns the `Reviewer\TalksController` into a `Reviewer\Talk\RateAction`

Follows #912.

Issue #1023, opened 2017-12-21
Replace home-made and Illuminate validation with Symfony validation


localheinz:
Currently we are using a bit of 

* home-made validation
* validation using Illuminate
* no validation

Ideally, we should be consistently using one kind of validation, and especially since we have turned this into a Symfony application, I suggest we use [`symfony/validator`](https://github.com/symfony/validator).

Related to #1020.

BackEndTea:
> validation using Illuminate

This one is only used in the base controller, so once we extract all controllers into actions this *should* be resolved.

On that note, the only place we use `illuminate/validaton`, `illuminate/container` or `illuminate/events` is in the base controller in the validation function. So once refactor that out we can drop those 3 dependencies.

mdwheele:
Oh, and there is 100% an argument for "We're using Symfony, we should use Symfony". 

😁

PR #1066, opened 2018-01-05
Moved request validation into a service

derrabus:
This PR extracts the `validate()` method, that has been copied into three action classes, into a separate class that is loaded as a service.

Follows: #1045, #1043, #1022
Relates to: #1023

### DIFF
--- a/classes/Http/Action/Reviewer/Talk/RateAction.php
+++ b/classes/Http/Action/Reviewer/Talk/RateAction.php
@@ -13,12 +13,9 @@ declare(strict_types=1);
 
 namespace OpenCFP\Http\Action\Reviewer\Talk;
 
-use Illuminate\Container\Container;
-use Illuminate\Filesystem\Filesystem;
-use Illuminate\Translation;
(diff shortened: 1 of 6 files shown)


classes/Http/Action/Reviewer/Talk/RateAction.php
```diff
@@ -28,15 +25,21 @@
...
+        $this->talkHandler      = $talkHandler;
+        $this->requestValidator = $requestValidator;
     }
 
     public function __invoke(HttpFoundation\Request $request): HttpFoundation\Response
     {
         try {
-            $this->validate($request, [
+            $this->requestValidator->validate($request, [
                 'rating' => 'required|integer',
```
derrabus: I'm unsure if we really need to throw a sophisticated validator at this problem, I must admit.

BackEndTea: We can probably just use a simpler check for this, but its good for this PR.

PR #1067, opened 2018-01-05
Enhancement: Configure no_extra_consecutive_blank_lines fixer

localheinz:
This PR

* [x] configures the `no_extra_consecutive_blank_lines` fixer 
* [x] runs `make cs`


classes/Domain/Talk/TalkFilter.php
```diff
@@ -71,25 +71,18 @@ public function getFilteredTalks(int $adminUserId, $filter = null)
...
-
             case 'plusone':
                 return $this->talk->ratedPlusOneBy($adminUserId);
-
             case 'viewed':
                 return $this->talk->viewedBy($adminUserId);
-
             case 'favorited':
                 return $this->talk->favoritedBy($adminUserId);
-
```
derrabus: I kinda like blank lines before `case`/`default`, I must say. 😕

localheinz: I don't mind removing the configuration for `case` statements, as I personally never use `switch` statements.

localheinz: This case is actually handled by configuring the `return` token: it removes extra consecutive blank lines after a `return` statement.

localheinz:
Thank you, @BackEndTea!

Issue #1068, opened 2018-01-09
RFC: Use correct HTTP response codes


BackEndTea:
# Suggestion

Lets use Proper HTTP Codes for the responses.

localheinz:
@derrabus 

Absolutely! Noticed this a lot when dissecting controllers into actions, and have this on my list. Probably good to start working on this on already dissected actions.

PR #1069, opened 2018-01-09
Add http codes to the talk create process action

BackEndTea:
This PR

* [x] Adds Proper HTTP response codes to the `CreateProcessAction`
* [x] Updates its test to explicitly check for correct HTTP codes

Related to #1068

### DIFF
--- a/classes/Http/Action/Talk/CreateProcessAction.php
+++ b/classes/Http/Action/Talk/CreateProcessAction.php
@@ -157,7 +157,7 @@ final class CreateProcessAction
                 'flash'          => $request->getSession()->get('flash'),
             ]);
 
-            return new HttpFoundation\Response($content);
+            return new HttpFoundation\Response($content, HttpFoundation\Response::HTTP_BAD_REQUEST);
         }
 
         $talk = Model\Talk::create(\array_merge($form->getCleanData(), [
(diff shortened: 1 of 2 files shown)


classes/Http/Action/Talk/CreateProcessAction.php
```diff
@@ -177,7 +177,7 @@ public function __invoke(HttpFoundation\Request $request): HttpFoundation\Respon
 
         $url = $this->urlGenerator->generate('dashboard');
 
-        return new HttpFoundation\RedirectResponse($url);
+        return new HttpFoundation\RedirectResponse($url, HttpFoundation\Response::HTTP_CREATED);
```
derrabus: In theory, you can perform a redirect with a `201` status. But the standard says that the location header should point to the created resource. imho, `302` is the better choice here.

I wouldn't use `201` on form processing, only for APIs.

classes/Http/Action/Talk/CreateProcessAction.php
```diff
@@ -157,7 +157,7 @@ public function __invoke(HttpFoundation\Request $request): HttpFoundation\Respon
                 'flash'          => $request->getSession()->get('flash'),
             ]);
 
-            return new HttpFoundation\Response($content);
+            return new HttpFoundation\Response($content, HttpFoundation\Response::HTTP_BAD_REQUEST);
```
derrabus: I would not throw a `400` when processing forms. Browsers might misinterpret that.

BackEndTea:
Closing this after recieved feedback

PR #1071, opened 2018-01-09
Enhancement: Pass model as constructor argument

BackEndTea:
This PR

* [x] Passes the Talk Model to the constructor of the HomePageAction, instead of statically accessing it

This should make the code base more testable since we can easily mock models.

Let me know if we want to do this for the other actions, or if this is a bad idea

### DIFF
--- a/classes/Domain/Model/Talk.php
+++ b/classes/Domain/Model/Talk.php
@@ -26,6 +26,9 @@ use Illuminate\Database\Eloquent\Relations\HasMany;
  * @method Builder ratedPlusOneBy(int $userId)
  * @method Builder viewedBy(int $userId)
  * @method Builder favoritedBy(int $userId)
+ *
+ * This is needed for Prophecy to mock the method, because it can't find it otherwise
+ * @method count()
  */
 class Talk extends Eloquent
 {
--- a/classes/Http/Action/Page/HomePageAction.php
+++ b/classes/Http/Action/Page/HomePageAction.php
@@ -23,9 +23,15 @@ final class HomePageAction
      */
     private $showSubmissionCount;
 
-    public function __construct(bool $showSubmissionCount)
+    /**
+     * @var Talk
+     */
+    private $talk;
+
+    public function __construct(bool $showSubmissionCount, Talk $talk)
     {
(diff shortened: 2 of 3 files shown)


classes/Domain/Model/Talk.php
```diff
@@ -26,6 +26,9 @@
  * @method Builder ratedPlusOneBy(int $userId)
  * @method Builder viewedBy(int $userId)
  * @method Builder favoritedBy(int $userId)
+ *
+ * This is needed for Prophecy to mock the method, because it can't find it otherwise
+ * @method count()
```
BackEndTea: If https://github.com/phpspec/prophecy/pull/339 passes we can remove this, but for now, every function that isn't directly on the model itself has to be added like this.

derrabus:
@localheinz suggested switching to Doctrine to improve testability: https://github.com/opencfp/opencfp/issues/618#issuecomment-352252107

If this is still under consideration, injecting the models everywhere could be unnecessary effort.

localheinz:
@chartjes @mdwheele 

What do you think about moving to Doctrine?

chartjes:
We just moved from Spot to Illuminate so I am a little worried about moving from Illuminate to Doctrine. I don't object to Doctrine on principal but are want some other gains that we can get other than "making things more testable"

IIRC Sentinel has very easy integration with Illumination -- will it be a lot of work to switch that to Doctrine?

BackEndTea:
@chartjes Sentinel provides some interfaces for creating your own implementations. 

If we want to move to Doctrine i can take a look at writing those

chartjes:
@BackEndTea I would like you to look into this please

chartjes:
@BackEndTea @localheinz What's the status of this? Are we going to put the effort into moving to Doctrine?

localheinz:
@chartjes 

I’m quite busy at the moment and can’t spare the time to invest for working on this, unfortunately.

chartjes:
@localheinz @derrabus @BackEndTea Can I close this PR then?

BackEndTea:
Yeah its okay to close this, i've been working on other projects lately, and don't really have the time to add doctrine here, sorry.

PR #1074, opened 2018-01-09
Remove redirectBack

BackEndTea:
This PR 

* [x] removes the `redirectBack` function from the BaseController

Its a left over from refactoring controllers into actions


BackEndTea:
@derrabus You are correct. I assumed it was used for the API, but was introduced during the implementation of the new front-end.

localheinz:

PR #1082, opened 2018-01-20
Set default charset for tables to utf-mb4

heiglandreas:
That way *all* Unicode-Characters can find their way into the CFP and not only those of the first Unicode-Plane. Which is missing f.e. the emojis… ;)

When this PR is merged all **new** instances will default to full unicode-support on MySQL. All future migrations need to take into account that indexes on `string` fields need the option `['limit' => 191]` to limit the keysize to the first 191 characters of the VARCHAR-field. That only applies to MySQL and will be ignored for other database-engines.

All **existing** instances might still run without full unicode-support on MySQL but all *new* tables will support full unicode. To update an **existing** MySQL-based instance to full unicode-support the users will need to follow the instructions (still to follow) in the README.md to make their installation unicode-compliant.


derrabus:
Good idea. :+1:

I think, you'll need to adjust the configuration of the EloquentBundle as well.

Also, I'm not that familiar with Phinx: Would this change make Phinx migrate all tables to `utf8mb4` or do we need to add a migration for this.

derrabus:
Also, you changes to the indexes probably caused the build failure on Travis.

heiglandreas:
There needs to be a migration as well as the current setting will only start to become effective on tables created after the change was introduced. So old databases might need to be migrated. Though a migration like that is then limited to mysql-like databases. I'd rather add an instruction on how to change that for a running database so the user can decide whether they want to do that or not. 

Or can we asume that *only* MySql is used?

derrabus:
> Or can we asume that only MySql is used?

It's currently the only supported database system, afaik.

BackEndTea:
If im correct we also want to support postgress 

Cc @chartjes

derrabus:
> If im correct we also want to support postgress

Okay, but right now, you don't. So for a migration we can assume that *if* there is an old database, it's a MySQL.

chartjes:
@derrabus There are plans to make it work with Postgres but they are still a ways off, no need to hold up @heiglandreas' PR in this case

PR #1085, opened 2018-01-20
Enhancement: Normalize composer.json

localheinz:
This PR

* [x] requires `localheinz/composer-normalize`
* [x] runs `composer normalize --dry-run` on Travis
* [x] normalizes `composer.json`

💁‍♂️ For reference, see 

* https://github.com/localheinz/composer-normalize
* https://localheinz.com/blog/2018/01/15/normalizing-composer.json/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ jobs:
         - composer install
 
       script:
+        - composer normalize --dry-run
         - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then COMMIT_RANGE=$TRAVIS_COMMIT_RANGE; else COMMIT_RANGE="HEAD~..HEAD"; fi;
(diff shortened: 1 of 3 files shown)


derrabus:
If we take that road, I'd suggest to make sure that composer files remain normalized. I have opened a feature request localheinz/composer-normalize#29 for that purpose.

localheinz:
@derrabus 

Because of https://github.com/localheinz/composer-normalize/pull/30 I require `php:^7.1`, so for now we can't use `localheinz/composer-normalize` as a direct development dependency, unless we drop support for PHP 7.0.

derrabus:
@localheinz I see. But you could do the validation on the php 7.1 Travis run only.

chartjes:
Thanks @localheinz and @derrabus

localheinz:
Thank you, @chartjes!

PR #1086, opened 2018-01-20
Enhancement: Extract EditAction from ProfileController

localheinz:
This PR

* [x] extracts a `Profile\EditAction` from the `ProfileController`

Follows #912.


localheinz:
Thank you, @chartjes!

Issue #1083, opened 2018-01-20
Ask only for the username for Joind.in-Link 


heiglandreas:
# Feature-Request:

Currently users can add a link to their joind.in profile. For that they need to enter the complete URL.

In my opinion it would be easier to let users enter their joind.in-username and to create the link  for them. That would then be the same process as for twitter.

localheinz:
Sounds good to me, @heiglandreas!

derrabus:
Migrating this could be ugly if users have already used that field for URLs other than a joind.in profile link.

chrisforrence:
@derrabus Fortunately (?), it looks like the validation for the URL had been restricted to joind.in links anyways, so users wouldn't have been able to put in other URLs

PR #1087, opened 2018-01-20
Fix: Remove @ from Twitter link and title

localheinz:
This PR

* [x] removes `@` from links and titles where they don't make any sense

Hardly related to #1083.

### Before

screen shot 2018-01-20 at 21 15 15

### After

screen shot 2018-01-20 at 21 16 17


localheinz:
Thank you, @BackEndTea!

PR #1088, opened 2018-01-21
Enhancement: Turn ProfileController into ProcessAction

localheinz:
This PR

* [x] turns the `ProfileController` into a `Profile\ProcessAction`

Follows #912.


localheinz:
Thank you, @chartjes!

PR #1092, opened 2018-01-28
Enhancement: Leverage Makefile rule syntax

localheinz:
This PR

* [x] leverages the `Makefile` rule syntax

💁‍♂️ For reference, see [comment on Reddit](https://www.reddit.com/r/PHP/comments/7sltmh/makefile_for_lazy_developers/dt5rba0/) by @nicwortel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: asset cache composer coverage cs database infection integration it stan test test-env unit
+.PHONY: asset cache coverage cs database infection integration it stan test test-env unit
 
 it: cs test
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,31 +1,28 @@
...
+.PHONY: asset cache coverage cs database infection integration it test test-env unit
 
 it: cs test
 
 asset:
 	yarn install
 	yarn run production
 
-cache:
+cache: vendor
```
localheinz: This target should probably depend on `vendor` as well as it would fail in the worst case (no dependencies installed).

localheinz:
@BackEndTea 

Rebased!

localheinz:
Thank you, @BackEndTea!

PR #1093, opened 2018-01-28
Enhancement: Require phpstan/phpstan

localheinz:
This PR

* [x] requires `phpstan/phpstan` 
* [x] adds a `Makefile` target

💁‍♂️ I think that as a start it could be useful to run 

```
$ make stan
```

locally.


localheinz:
Nice how it fails even on the basic level for anything related to Laravel.

chartjes:
@localheinz Maybe we could cut back on the Laravel bashing and look at what phpstan tells us and decide if it's worth doing.

PR #1104, opened 2018-02-15
Fix: Unknown target

localheinz:
This PR

* [x] uses the `vendor` target instead of the removed `composer` target

Follows #1092.


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #1103, opened 2018-02-15
Make sure login is not too thin on mobile

Nyholm:
Before: 
screen shot 2018-02-15 at 10 18 40

After: 
screen shot 2018-02-15 at 10 18 36

PR #1105, opened 2018-02-15
Fix: Compile assets

localheinz:
This PR

* [x] compiles assets

Follows #1103.


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #1107, opened 2018-02-18
Fix: Drop support for PHP 7.0

localheinz:
This PR

* [x] drops support for PHP 7.0

For reference, see http://php.net/supported-versions.php:

screen shot 2018-02-18 at 23 08 58

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,11 +6,11 @@
         "preferred-install": "dist",
         "sort-packages": true,
         "platform": {
-            "php": "7.0.25"
+            "php": "7.1.3"
         }
     },
     "require": {
-        "php": ">=7",
+        "php": "^7.1",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -4,13 +4,10 @@
     "license": "MIT",
     "config": {
         "preferred-install": "dist",
-        "sort-packages": true,
-        "platform": {
-            "php": "7.0.25"
```
derrabus: You should keep the platform section. Just raise it to a 7.1 version, e.g. 7.1.3 (which is the minimum for Symfony 4 and Laravel 5.6).

localheinz: Thank you, @derrabus, updating!

localheinz:
Thank you, @chartjes!

PR #1122, opened 2018-04-02
Fix #1120: Add note that command "bin/console server:start" is only available in application mode development

### LINKED ISSUE #1120: "bin/console server:start" (in production env) throws "There are no commands defined in the "server" namespace."

1. `git clone https://github.com/opencfp/opencfp.git`
2. `cd opencfp && script/setup`
3. `cp config/production.yml.dist  config/production.yml`
4. `export CFP_ENV=production`
5.  `bin/console server:start`

# Expected Result

* A running opencfp with the PHP internal web server as described here in the README: https://github.com/opencfp/opencfp#php-built-in-web-server

# Actual Result

Error message
```
  There are no commands defined in the "server" namespace.
```

# Possible solution

After applying
```
export CFP_ENV=development
cp config/development.yml.dist config/development.yml
bin/console server:start
```

it works as expected.
Maybe a small note into the README that this only works in development mode might be useful for starters/beginners with OpenCFP.

On the Symfony docs [How to Use PHP's built-in Web Server](https://symfony.com/doc/current/setup/built_in_web_server.html) there is a warning like

> The built-in web server is meant to be run in a controlled environment. It is not designed to be used on public networks.

which make sense. I agree. But I found this warning only after clicking the link and it was not obvious to me at first.
So I would not consider this issue as a bug. More as a small improvement to make this app more accessible to less experienced users with PHP (who are not able to figure this out).

PS: Thanks for this application :)

andygrunwald:
This PR

* [x] Adds a small note that the command "bin/console server:start" is only available in application mode development

Fixes #1120

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ $ bin/console server:start
 ```
 
 The server uses port `8000`. This is a quick way to get started doing development on OpenCFP itself.
+Important note: The command `server:start` is only available when the application environment (`CFP_ENV`) is set to `development`.
 
 Details on how to use this console command can be found at Symfony's documentation for [using the built-in web server](http://symfony.com/doc/current/setup/built_in_web_server.html).
 


localheinz:
Thank you, @andygrunwald!

PR #1124, opened 2018-04-04
Add "/cache/env" and "/log" into writable folders

andygrunwald:
This PR

* [x] Adds documentation that two more folders needs to be writable for a production env


chartjes:
Thanks @andygrunwald

Issue #1119, opened 2018-03-29
joind.in field rejects valid URL


kevinsmith:
# Steps required to reproduce the problem

1. When editing your profile, enter `https://joind.in/user/_KevinSmith` into the joind.in field, which [as you can see](https://joind.in/user/_KevinSmith) is a valid URL
2. Click update profile button
3. Get the error "You did not enter a valid joind.in URL"

# Expected Result

* URL is accepted

# Actual Result

* URL is rejected as invalid

PR #1129, opened 2018-04-21
Updates the regex for joind.in to better match valid characters

chrisforrence:
This PR addresses #1119.

The joind.in username validation is relatively [permissive](https://github.com/joindin/joindin-web2/blob/master/app/src/User/RegisterFormType.php#L42), only checking that it's not blank and it's a maximum of 100 characters. Without spinning up a joind.in instance, I wasn't comfortable testing their site for possible usernames (emoji, other special characters), hence why I'm still relatively restrictive in this PR.

### DIFF
--- a/classes/Http/Form/SignupForm.php
+++ b/classes/Http/Form/SignupForm.php
@@ -201,9 +201,9 @@ class SignupForm extends Form
         return $validationResponse;
     }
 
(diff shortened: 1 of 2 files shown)


chartjes:
Thanks @chrisforrence

PR #1141, opened 2018-05-04
Feature/docker dev

zaherg:
This PR has the following:

* [x] remove xDebug from the docker image, and use a base image which does not use xdebug, this made the image even smaller in size (270MB)
* [x] add a new `DOCKER.md` file, which explains how we can utilize the docker image for development.
* [x] add a persistent volume for MySQL so that the tables won't be deleted after each build.


chartjes:
Thank you @linuxjuggler

linuxjuggler:
most welcome @chartjes, but I was wondering if you are thinking of hosting an image on docker hub for people to use instead of building it themselves

PR #1147, opened 2018-05-21
GDPR Delete User functionality

chartjes:
This PR

* [x] Adds functionality to allow a user to delete all their information from OpenCFP


resources/views/user/delete.twig
```diff
@@ -0,0 +1,16 @@
...
+
+{% block scripts %}
+    {{ parent() }}
+{% endblock %}
+
+{% block content %}
+    <h2 class="headline">Delete My Profile</h2>
+    <p>
+        By clicking on the button below you all data associated with you account will be
+        deleted from the OpenCFP instance and then you will be logged out.
```
localheinz: This should probably be

```diff
diff --git a/resources/views/user/delete.twig b/resources/views/user/delete.twig
index e7c7f2c4..ae937e46 100644
--- a/resources/views/user/delete.twig
+++ b/resources/views/user/delete.twig
@@ -7,10 +7,10 @@
 {% block content %}
     <h2 class="headline">Delete My Profile</h2>
     <p>
-        By clicking on the button below you all data associated with you account will be
+        By clicking on the button below all data associated with your account will be
         deleted from the OpenCFP instance and then you will be logged out.
     </p>
     <form method="post" action="{{ url('user_delete_process') }}">
         <button type="submit" class="btn btn-brand mt-2 ml-4">Delete My Profile</button>
     </form>
-{% endblock %}
\ No newline at end of file
+{% endblock %}
```

Also note missing empty line at end of file.

resources/views/dashboard.twig
```diff
@@ -55,6 +55,7 @@ function deleteTalk(tid) {
         <div>
             <a class="btn btn-white-o mr-2" href="{{ url('password_edit') }}">Change Password</a>
             <a class="btn btn-white-o mr-2" href="{{ url('user_edit', { id: user.id }) }}">Edit Profile</a>
+            <a class="btn btn-white-o mr-2" href="{{ url('user_delete') }}">Delete Profile</a>
```
localheinz: When I freshly signed up, the dashboard shows the following:

screen shot 2018-05-22 at 08 36 15

Also see https://github.com/opencfp/opencfp/blob/2d3ea6a061ac31ae0c73090d02a3e4edae73177f/resources/views/dashboard.twig#L25-L32

That is, unless newly signed-up users fill out their personal details, they cannot see the link tot the *Delete Profile* page. What do you think about adjusting `dashboard.twig` to also show the link when the profile is not filled out yet?

tests/Integration/Http/Action/Profile/ProcessDeleteActionTest.php
```diff
@@ -0,0 +1,47 @@
...
+    public function deletedUserHasNoRecords()
+    {
+        $speaker = factory(User::class, 1)->create()->first();
+        factory(Talk::class, 1)->create()->first();
+
+        $response = $this
+            ->asLoggedInSpeaker($speaker->id)
+            ->post('/profile/delete');
+        $this->assertResponseIsRedirect($response);
+        $this->assertResponseBodyContains('Redirecting to /', $response);
```
localheinz: Shall we assert here that neither the user nor any of their talks exist here anymore?

Also, I have taken a look at the database, and the following tables also contain references to `users`:

* `favorites`
* `persistences`
* `reminders`
* `talk_comments`
* `talk_meta`
* `throttle`

What do you think about adding more tests to make sure all of these are deleted as well (unless we want to keep them, but then, somehow anonymized)?

src/Http/Action/Profile/ProcessDeleteAction.php
```diff
@@ -0,0 +1,50 @@
...
+        Services\Authentication $authentication,
+        Routing\Generator\UrlGeneratorInterface $urlGenerator
+    ) {
+        $this->authentication = $authentication;
+        $this->urlGenerator   = $urlGenerator;
+    }
+
+    public function __invoke(HttpFoundation\Request $request): HttpFoundation\Response
+    {
+        $user = User::find($this->authentication->user()->getId());
```
localheinz: Before we attempt to find a user here, should we make sure that the current user is actually authenticated?

src/Http/Action/Profile/ProcessDeleteAction.php
```diff
@@ -0,0 +1,50 @@
...
+        $user->delete();
```
localheinz: Before we delete a user here, should we make sure that we actually found a user?

src/Http/Action/Profile/ProcessDeleteAction.php
```diff
@@ -0,0 +1,50 @@
...
+        $this->authentication->logout();
+        $url = $this->urlGenerator->generate('homepage');
```
localheinz: What do you think about adding a flash message here to inform the user that their data has been deleted?

localheinz:
@chartjes 

Quick note, as suggested by someone earlier, I have created a [GDPR](https://github.com/opencfp/opencfp/milestone/1) milestone, maybe we can remove the [GDPR](https://github.com/opencfp/opencfp/labels/GDPR) label, then?

tests/Integration/Http/Action/Profile/ProcessDeleteActionTest.php
localheinz: I think we probably need two tests (one for each user role) with each user having added as much data as they can currently add, and then assert that after deletion nothing is there.

chartjes:
@localheinz Implemented most of what you commented on, please take a look

resources/views/dashboard.twig
```diff
@@ -26,8 +26,11 @@ function deleteTalk(tid) {
     <div class="text-center">
         <i class="fa fa-user-circle text-soft fa-5x"></i>
         <h1 class="font-serif font-thin text-5xl">Welcome to {{ site.title }}!</h1>
-        <p class="text-soft mb-8">It looks like this is your first time here ...</p>
+        <p class="text-soft mb-8">It looks like this is your first time here...</p>
         <a href="{{ url('user_edit', { id: user.id }) }}" class="btn btn-brand">Fill out your profile</a>
+        <br><br>
+        <p class="text-soft mb-8">Click below to delete your account without going any further</p>
+        <a href="{{ url('user_delete') }}" class="btn btn-brand">Delete your account</a>
```
localheinz: Minor nitpick, but for consistency, can we use the same order of attributes as on the link below?

Feel free to use this patch here:

```diff
diff --git a/resources/views/dashboard.twig b/resources/views/dashboard.twig
index cbc69eb5..9221a773 100644
--- a/resources/views/dashboard.twig
+++ b/resources/views/dashboard.twig
@@ -30,7 +30,7 @@ function deleteTalk(tid) {
         <a href="{{ url('user_edit', { id: user.id }) }}" class="btn btn-brand">Fill out your profile</a>
         <br><br>
         <p class="text-soft mb-8">Click below to delete your account without going any further</p>
-        <a href="{{ url('user_delete') }}" class="btn btn-brand">Delete your account</a>
+        <a class="btn btn-brand" href="{{ url('user_delete') }}">Delete your account</a>
     </div>
     {% else %}
     <div class="bg-brand p-8 text-white flex items-center justify-between">
```

localheinz: On the other hand, we use `btn btn-brand` as classes for the link here, but `btn btn-white-o mr-2` below - to the same effect. Maybe we should use semantic classes for these (no need to change anything now).

resources/views/user/delete.twig
```diff
@@ -0,0 +1,18 @@
...
+{% block content %}
+    <div class="text-center">
+            <h2 class="headline">Delete My Profile</h2>
+            <p>
+                By clicking on the button below all data associated with your account<br>
+                will be deleted from this OpenCFP install and then you will be logged out.
+            </p>
+            <form method="post" action="{{ url('user_delete_process') }}">
+                <button type="submit" class="btn btn-brand mt-2 ml-4">Delete My Profile</button>
+            </form>
```
localheinz: The indentation here is a bit off, how do you feel about outdenting once?

resources/views/user/delete.twig
```diff
@@ -0,0 +1,18 @@
...
+
+{% block scripts %}
+    {{ parent() }}
+{% endblock %}
+
+{% block content %}
+    <div class="text-center">
+            <h2 class="headline">Delete My Profile</h2>
+            <p>
+                By clicking on the button below all data associated with your account<br>
```
localheinz: I feel that the usage of `<br>` tag here is for cosmetic reasons - maybe we would be better of by using a container for the entire block that limits the width of what is contained?

src/Domain/Model/User.php
```diff
@@ -34,6 +34,21 @@ public function meta(): HasMany
...
+    {
+        return $this->hasMany(Persistence::class, 'user_id');
+    }
+
+    public function reminders(): HasMany
+    {
+        return $this->hasMany(Reminder::class);
+    }
+
+    public function throttle(): HasMany
```
localheinz: For consistency, should we use plural here?

src/Domain/Model/User.php
```diff
@@ -99,6 +114,18 @@ public function delete(): bool
...
+                throw new \Exception('Unable to delete persistence records of user');
+            }
+        });
+
+        $this->reminders()->get()->each(function ($item) {
+            if (!$item->delete()) {
+                throw new \Exception('Unable to delete reminder records of user');
+            }
+        });
+
```
localheinz: Should we also delete all throttles here?

src/Http/Action/Profile/ProcessDeleteAction.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $user = User::find($this->authentication->user()->getId());
+
+        if (!$user) {
+            $url = $this->urlGenerator->generate('login');
+
+            return new HttpFoundation\RedirectResponse($url);
+        }
+
+        $user->delete();
```
localheinz: Since we throw exceptions in `delete()`, what do you think about wrapping this in a `try`-and-`catch` block to ensure we at least fail with a pleasant error message?

src/Domain/Model/User.php
```diff
@@ -99,6 +114,18 @@ public function delete(): bool
                 }
             });
 
+        $this->persistences()->get()->each(function ($item) {
```
localheinz: Can we add type declarations here?

```diff
diff --git a/src/Domain/Model/User.php b/src/Domain/Model/User.php
index eaf08c77..81016ff7 100644
--- a/src/Domain/Model/User.php
+++ b/src/Domain/Model/User.php
@@ -114,13 +114,13 @@ class User extends Eloquent
                 }
             });

-        $this->persistences()->get()->each(function ($item) {
+        $this->persistences()->get()->each(function (Persistence $item): void {
             if (!$item->delete()) {
                 throw new \Exception('Unable to delete persistence records of user');
             }
         });

-        $this->reminders()->get()->each(function ($item) {
+        $this->reminders()->get()->each(function (Reminder $item): void {
             if (!$item->delete()) {
                 throw new \Exception('Unable to delete reminder records of user');
             }
```

tests/Unit/ProjectCodeTest.php
```diff
@@ -39,6 +39,9 @@ public function productionClassesHaveUnitTests()
                 Domain\Model\Airport::class,
                 Domain\Model\Eloquent::class,
                 Domain\Model\Favorite::class,
+                Domain\Model\Persistence::class,
+                Domain\Model\Reminder::class,
+                Domain\Model\Throttle::class,
```
localheinz: Looks like these have become a bit out of order, maybe apply this patch?

```diff
diff --git a/tests/Unit/ProjectCodeTest.php b/tests/Unit/ProjectCodeTest.php
index 491af09b..8bb65d57 100644
--- a/tests/Unit/ProjectCodeTest.php
+++ b/tests/Unit/ProjectCodeTest.php
@@ -41,27 +41,27 @@ final class ProjectCodeTest extends Framework\TestCase
                 Domain\Model\Favorite::class,
                 Domain\Model\Persistence::class,
                 Domain\Model\Reminder::class,
-                Domain\Model\Throttle::class,
                 Domain\Model\Talk::class,
                 Domain\Model\TalkComment::class,
                 Domain\Model\TalkMeta::class,
+                Domain\Model\Throttle::class,
                 Domain\Model\User::class,
                 Domain\Services\ProfileImageProcessor::class,
                 Domain\Talk\TalkFormatter::class,
                 Domain\Talk\TalkHandler::class,
                 Http\Action\Admin\DashboardAction::class,
                 Http\Action\Admin\Talk\IndexAction::class,
+                Http\Action\Admin\Talk\RateAction::class,
                 Http\Action\Profile\ChangePasswordProcessAction::class,
                 Http\Action\Profile\DeleteAction::class,
                 Http\Action\Profile\ProcessDeleteAction::class,
                 Http\Action\Reviewer\DashboardAction::class,
                 Http\Action\Reviewer\Speaker\IndexAction::class,
                 Http\Action\Reviewer\Speaker\ViewAction::class,
-                Http\Action\Admin\Talk\RateAction::class,
                 Http\Action\Reviewer\Talk\IndexAction::class,
                 Http\Action\Reviewer\Talk\RateAction::class,
-                Http\Action\Signup\ProcessAction::class,
                 Http\Action\Signup\PrivacyAction::class,
+                Http\Action\Signup\ProcessAction::class,
                 Http\Controller\Admin\ExportsController::class,
                 Http\Controller\Admin\SpeakersController::class,
                 Http\Controller\Admin\TalksController::class,
```

src/Domain/Model/Throttle.php
```diff
@@ -0,0 +1,41 @@
...
+    public function user(): BelongsTo
+    {
+        return $this->belongsTo(User::class, 'user_id');
+    }
+
+    public function setUpdatedAt($value)
+    {
+        /**
+         * This is the dirty way to tell Illuminate that we don't have an updated at field
+         * while still having a created_at field.
```
localheinz: Ha, but we don't have a `created_at` field, do we?

b3qijlfccaejlwm jpg-large

https://twitter.com/nzkoz/status/538892801941848064

😆

src/Domain/Model/User.php
chartjes: The table is called throttle so I tried to keep it consistent...

resources/views/user/delete.twig
chartjes: I will need some help from someone with better layout skills to do this for me right now

src/Domain/Model/User.php
chartjes: Will add that in

PR #1150, opened 2018-06-06
Using show_contrib_banner parameter on splash layout

jschaedl:
Using site.show_contrib_banner parameter on splash layout for enable/disable fork me banner on signup, login and forgot_password pages.


chartjes:
Thank you @jschaedl

PR #1152, opened 2018-06-13
Fix: Default to production when environment could not be determined from environment variable

### LINKED ISSUE #1151: Environment should default to production
In the default installation, the environment is initialized as

```php
$environment = Environment::fromServer($_SERVER);
```
which returns `Environment::TYPE_DEVELOPMENT` if no environment var is present.

Returning `Environment::TYPE_PRODUCTION` might be safer as a default.

localheinz:
This PR

* [x] adds a missing test for `Environment::fromServer()`
* [x] defaults to production environment if the environment could not be determined from the environment variable

Fixes #1151.

💁‍♂️ Apologies for the clunky work - on vacation and working from an iPad.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -84,7 +84,7 @@ class Environment
      */
     public static function fromServer(array $server): self
     {
-        $type = $server['CFP_ENV'] ?? self::TYPE_DEVELOPMENT;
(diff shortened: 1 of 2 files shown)


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

PR #1155, opened 2018-08-24
remove paragonie/random_compat

smoench:
random_bytes() and random_int() are integrated with PHP 7


BackEndTea:
On that note we can also replace the symfony/polyfill-php* that are less then 71

smoench:
@BackEndTea done :)

PR #1159, opened 2018-09-01
Switched out speaker photo upload functionality to use flysystem

carnage:
This PR

* [x] Adds dependency on Flysystem
* [x] Uses Flysystem to handle photo uploads
* [x] Defines a default filesystem which matches the current upload directory for BC 

This may not seem overly useful on it's own as you can only use the Local adaptor, however it is possible with a bit of nginx/apache magic to use both S3 and GCS as backing stores by rewriting urls, for the time being, this is left as an exercise for the user.

This probably breaks tests and hasn't been fully tested yet. Opening early PR for feedback.


carnage:
I've not had the time to fully test this and there aren't any automated tests which cover this area. If someone else was to verify it's working, I'd be happy for it to be merged if someone else tested it out first, otherwise it will have to wait till I've solved the other issues for the kubernetes deployment I'm working on.

PR #1160, opened 2018-09-05
phinx: CFP_ENV is passed via enviroment variable instead of $_SERVER

smoench:
In `script/test` the `CFP_ENV` is passed via envirmoment variable. `phinx.php` should only be used on cli anyway.

### DIFF
--- a/phinx.php
+++ b/phinx.php
@@ -15,7 +15,7 @@ use OpenCFP\Kernel;
 
 require_once __DIR__ . '/vendor/autoload.php';
 
-$kernel = new Kernel((string) Environment::fromServer($_SERVER), false);
+$kernel = new Kernel((string) Environment::fromServer(\getenv()), false);
 $kernel->boot();
 
 $container = $kernel->getContainer();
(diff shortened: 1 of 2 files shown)


phinx.php
```diff
@@ -15,7 +15,7 @@
 
 require_once __DIR__ . '/vendor/autoload.php';
 
-$kernel = new Kernel((string) Environment::fromServer($_SERVER), false);
+$kernel = new Kernel((string) Environment::fromServer(\getenv()), false);
```
localheinz: @smoench 

I think we need to set an environment variable in [`.travis.yml`](https://github.com/opencfp/opencfp/blob/4430b2d56bd9f2c1640e313d9c104f731a2ae406/.travis.yml#L6), now, so we can properly determine the environment on Travis.

Can you adjust, please?

It's already set in [`Makefile`](https://github.com/opencfp/opencfp/blob/4430b2d56bd9f2c1640e313d9c104f731a2ae406/Makefile#L22).

localheinz:
Thank you, @smoench!

PR #1162, opened 2018-09-16
Add an option to trust headers set by reverse proxies

carnage:
eg Cloudflare, kubernetes ingresses etc

Going to tidy up and test this tomorrow, putting in a PR as a reminder

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ mail:
 ```
 ### [Running behind a trusted proxy](#run-trusted-proxy)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -252,6 +252,10 @@ mail:
     encryption: tls
     auth_mode: ~
 ```
+### [Running behind a trusted proxy](#run-trusted-proxy)
+
+If you are running open cfp behing a proxy server which adds X-Forwarded-For headers (this could be a cloud based load balancer or a service such as cloudflare) you will need to set the environment variable TRUST_PROXIES to true this will ensure that opencfp trusts the headers set by these proxies for the original IP address and ssl mode. Setting this will trust these headers regardless of where the original request originates, so it's advisable to either lock down your instance so that only the trusted proxy can access it, or modifiy the list of trusted proxies in the index.php file to only include the ip addresses of your proxies.
```
chartjes: The project name is OpenCFP

drbyte: Also: `modifiy` should be `modify`

localheinz: Apologies for chiming it, but I think it also should be `Cloudflare`, see https://www.cloudflare.com.

localheinz:
@carnage 

Can you take another look, please?

carnage:
Yep, just haven't had time yet

PR #1163, opened 2018-09-21
Enable SSO via OpenCFP Central

chartjes:
This PR

* [x] Adds the configurable ability to use OpenCFP Central as a SSO option


chartjes:
Ooops, not done yet

PR #1164, opened 2018-09-30
Ch sso

chartjes:
This PR

* [x] adds the ability to use OpenCFP Central as a single-sign-on source

### DIFF
--- a/config/development.yml.dist
+++ b/config/development.yml.dist
@@ -68,3 +68,12 @@ security:
 reviewer:
   users:
     -
+
+opencfpcentral:
+  sso: off
+  clientId:
+  clientSecret:
+  authorizeUrl: http://localhost/oauth/authorize?
+  redirectUri: http://localhost/redirect
+  resourceUri: https://www.opencfpcentral.com/api/user
+  tokenUrl: https://www.opencfpcentral.com/oauth/token
--- a/config/development.yml.dist
+++ b/config/development.yml.dist
@@ -73,7 +73,7 @@ opencfpcentral:
   sso: off
   clientId: 0
   clientSecret: 0
-  authorizeUrl: http://localhost/oauth/authorize?
+  authorizeUrl: https://www.opencfpcentral.com/oauth/authorize?
   redirectUri: http://localhost/redirect
   resourceUri: https://www.opencfpcentral.com/api/user
   tokenUrl: https://www.opencfpcentral.com/oauth/token
--- a/config/development.yml.dist
+++ b/config/development.yml.dist
@@ -71,8 +71,8 @@ reviewer:
 
 opencfpcentral:
   sso: off
-  clientId:
-  clientSecret:
+  clientId: 0
+  clientSecret: 0
   authorizeUrl: http://localhost/oauth/authorize?
   redirectUri: http://localhost/redirect
   resourceUri: https://www.opencfpcentral.com/api/user
--- a/config/development.yml.dist
+++ b/config/development.yml.dist
@@ -74,6 +74,6 @@ opencfpcentral:
   clientId: 0
   clientSecret: 0
   authorizeUrl: https://www.opencfpcentral.com/oauth/authorize?
-  redirectUri: http://localhost/redirect
+  redirectUri: http://localhost/sso/redirect
   resourceUri: https://www.opencfpcentral.com/api/user
   tokenUrl: https://www.opencfpcentral.com/oauth/token
--- a/resources/config/routing.yml
+++ b/resources/config/routing.yml
@@ -172,9 +172,8 @@ talk_view:
 login:
   path: /login
   methods: [GET]
-  controller: FrameworkBundle:Template:template
-  defaults:
-    template: security/login.twig
+  controller: OpenCFP\Http\Action\Security\ShowLogInAction
+
 
 login_check:
   path: /login
@@ -257,4 +256,14 @@ privacy:
    path: /privacy
    methods: [GET]
    controller: OpenCFP\Http\Action\Signup\PrivacyAction
+
+central:
+  path: /central
+  methods: [GET]
+  controller: OpenCFP\Http\Action\Security\CentralAction
+
+redirect:
+  path: /redirect
+  methods: [GET]
+  controller: OpenCFP\Http\Action\Security\RedirectAction
 
--- a/resources/config/routing.yml
+++ b/resources/config/routing.yml
@@ -260,10 +260,10 @@ privacy:
 sso:
   path: /sso
   methods: [GET]
-  controller: OpenCFP\Http\Action\Security\CentralAction
+  controller: OpenCFP\Http\Action\Security\SsoAction
 
 sso_redirect:
   path: /sso/redirect
   methods: [GET]
-  controller: OpenCFP\Http\Action\Security\RedirectAction
+  controller: OpenCFP\Http\Action\Security\SsoRedirectAction
 
--- a/resources/config/routing.yml
+++ b/resources/config/routing.yml
@@ -257,13 +257,13 @@ privacy:
    methods: [GET]
    controller: OpenCFP\Http\Action\Signup\PrivacyAction
 
-central:
-  path: /central
+sso:
+  path: /sso
   methods: [GET]
   controller: OpenCFP\Http\Action\Security\CentralAction
 
-redirect:
-  path: /redirect
+sso_redirect:
+  path: /sso/redirect
   methods: [GET]
   controller: OpenCFP\Http\Action\Security\RedirectAction
 
--- a/resources/views/security/login.twig
+++ b/resources/views/security/login.twig
@@ -13,10 +13,13 @@
       <input id="email" type="email" name="email" value="{% if email %}{{ email | escape }}{% endif %}" class="w-full" placeholder="you@example.com">
       <label for="password" class="text-sm text-white">Password</label>
       <input id="password" type="password" name="password" class="w-full" placeholder="•••••••">
-      <p class="text-right my-4"><a href="{{ url('forgot_password') }}" class="text-white text-sm">Forgot password?</a></p>
+      <p class="text-right my-4"><a href="{{ url('forgot_password') }}" class="text-white text-sm">Forgot your password? {{ sso }}</a></p>
       <button type="submit" class="btn btn-brand w-full">Login</button>
     </form>
-
-    <p class="text-light text-center text-sm my-4">New user? <a href="{{ url('user_new') }}" class="text-brand text-bold">Signup Now</a></p>
+    <br>
+    {% if sso == 'on' %}
+    <a href="/central"><button class="btn btn-brand w-full">Login using OpenCFP Central</button></a>
+    {% endif %}
+    <p class="text-light text-center text-sm my-4">New user? <a href="{{ url('user_new') }}" class="text-brand text-bold">Sign up Now</a></p>
   </div>
 {% endblock %}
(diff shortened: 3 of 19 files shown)


config/development.yml.dist
```diff
@@ -68,3 +68,12 @@ security:
 reviewer:
   users:
     -
+
+opencfpcentral:
```
localheinz: When first running

```
$ make
```

after checkout, the process failed with

```
  The service "OpenCFP\Http\Action\Security\CentralAction" has a dependency on a non-existent parameter "opencfpcentral.clientId".
```

Maybe we should extract parameters and use [`incenteev/parameter-handler`](https://github.com/Incenteev/ParameterHandler)? This is clearly out of scope of this PR, though.

resources/views/security/login.twig
```diff
@@ -13,10 +13,13 @@
       <input id="email" type="email" name="email" value="{% if email %}{{ email | escape }}{% endif %}" class="w-full" placeholder="you@example.com">
       <label for="password" class="text-sm text-white">Password</label>
       <input id="password" type="password" name="password" class="w-full" placeholder="•••••••">
-      <p class="text-right my-4"><a href="{{ url('forgot_password') }}" class="text-white text-sm">Forgot password?</a></p>
+      <p class="text-right my-4"><a href="{{ url('forgot_password') }}" class="text-white text-sm">Forgot your password? {{ sso }}</a></p>
```
localheinz: Not sure, why are we rendering `{{ sso }}` here?

screen shot 2018-10-03 at 09 16 55

src/Http/Action/Security/CentralAction.php
```diff
@@ -0,0 +1,94 @@
...
+        string $redirectUri,
+        string $authorizeUrl,
+        string $sso
+    ) {
+        $this->authentication = $authentication;
+        $this->urlGenerator   = $urlGenerator;
+        $this->clientId       = $clientId;
+        $this->redirectUri    = $redirectUri;
+        $this->authorizeUrl   = $authorizeUrl;
+        $this->sso            = $sso;
```
localheinz: This field is actually never used, can we remove it, as well as the dependency on the `$sso` parameter?

src/Http/Action/Security/LogInAction.php
```diff
@@ -30,6 +30,9 @@
      */
     private $twig;
 
+    /** @var string */
+    private $sso;
```
localheinz: This field is never used, can we remove it, or revert all of the changes to this file?

resources/config/routing.yml
```diff
@@ -258,3 +257,13 @@ privacy:
    methods: [GET]
    controller: OpenCFP\Http\Action\Signup\PrivacyAction
 
+central:
+  path: /central
+  methods: [GET]
+  controller: OpenCFP\Http\Action\Security\CentralAction
+
+redirect:
+  path: /redirect
```
localheinz: These paths seem a bit generic, perhaps it makes sense to rename a few of the paths?

* `/opencfp-central`
* `/opencfp-central/redirect`

or

* `/sso`
* `/sso/redirect`

What do you think?

In the same light, perhaps renaming the actions also makes sense?

resources/config/routing.yml
```diff
@@ -258,3 +257,13 @@ privacy:
    methods: [GET]
    controller: OpenCFP\Http\Action\Signup\PrivacyAction
 
+central:
+  path: /central
```
localheinz: When navigating to `/central`, after a redirect I end up here:

screen shot 2018-10-03 at 10 24 43

What am I missing here?

Also see the `opencfpcentral.authorizeUrl` parameter.

src/Http/Action/Security/RedirectAction.php
```diff
@@ -0,0 +1,158 @@
...
+        /**
+         * What do we do with the access token?
+         * Check if the user exists
+         * -> if they don't, generate a random password and then create and activate an account
+         * manually log them in
+         * redirect them to their dashboard
+         */
+        $details = \json_decode((string) $response->getBody(), true);
+
+        $user_response = $this->httpClient->get($this->resourceUri, [
```
localheinz: Should we use `camelCase` here and elsewhere for variable names?

src/Http/Action/Security/ShowLogInAction.php
```diff
@@ -0,0 +1,73 @@
...
+         * Otherwise show the login page
+         */
+        try {
+            $this->authentication->authenticate(
+                $request->get('email'),
+                $request->get('password')
+            );
+            $url = $this->urlGenerator->generate('dashboard');
+
+            return new HttpFoundation\RedirectResponse($url);
```
localheinz: Can we move this and the previous statement to after the `try`-and-`catch` block?

I suppose the only thing that can go wrong here and that we care about is the authentication.

config/development.yml.dist
chartjes: Make now passes for everything I've done here (saw similar errors but thanks for pointing out I forgot to run it) and I am open to using that parameter-handler but in a follow-up PR once someone other than me can get SSO up-and-running for themselves.

resources/views/security/login.twig
chartjes: Removed that, must have been left over from my debugging and testing of the feature via manual testing.

src/Http/Action/Security/CentralAction.php
chartjes: Removed

src/Http/Action/Security/LogInAction.php
chartjes: Removed

resources/config/routing.yml
chartjes: I have made those suggested changes, need to also update the wording on the OpenCFP Central web site to reflect this.

resources/config/routing.yml
chartjes: I have updated all the config files to use the correct authorizeUrl value

src/Http/Action/Security/ShowLogInAction.php
chartjes: I don't see a reason to move it. Convince me!

localheinz: None of

```php
$url = $this->urlGenerator->generate('dashboard');

return new HttpFoundation\RedirectResponse($url);
```

will throw an `AuthenticationException`. 

Instead, these two statements constitute the success case on the happy path. By moving them out of the `try` block we make it more clear that they are on that happy path. 

Still, an exception (albeit not an `AuthenticationException`) might be thrown, perhaps when a URL with the `dashboard` identifier could not be generated, but that is clearly out of the scope of handling the authentication.

Does that work?

src/Http/Action/Security/SsoAction.php
```diff
@@ -0,0 +1,91 @@
...
+    /** @var string */
+    private $authorizeUrl;
+
+    /**
+     * @var Routing\Generator\UrlGeneratorInterface
+     */
+    private $urlGenerator;
+
+    /** @var string */
+    private $sso;
```
localheinz: Field is still not used, can we remove it?

src/Http/Action/Security/SsoAction.php
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * LoginAction constructor.
+     *
+     * @param Services\Authentication                 $authentication
+     * @param Routing\Generator\UrlGeneratorInterface $urlGenerator
+     * @param int                                     $clientId
+     * @param string                                  $redirectUri
+     * @param string                                  $authorizeUrl
+     */
```
localheinz: What do you think about removing the DocBlock, as it only repeats information that is already present from the type declarations of the method arguments? Also, following the rename of the action, the short description has become misleading.

src/Http/Action/Security/SsoRedirectAction.php
```diff
@@ -0,0 +1,158 @@
...
+     * @param Sentinel                                $sentinel
+     * @param AccountManagement                       $accounts
+     * @param Routing\Generator\UrlGeneratorInterface $urlGenerator
+     * @param int                                     $clientId
+     * @param string                                  $clientSecret
+     * @param string                                  $redirectUri
+     * @param string                                  $resourceUri
+     * @param string                                  $tokenUrl
+     * @param Client                                  $httpClient
+     */
```
localheinz: See comment above about suggesting to remove the DocBlock.

src/Http/Action/Security/SsoRedirectAction.php
```diff
@@ -0,0 +1,158 @@
...
+    public function __construct(
+        Sentinel $sentinel,
+        AccountManagement $accounts,
+        Routing\Generator\UrlGeneratorInterface $urlGenerator,
+        int $clientId,
+        string $clientSecret,
+        string $redirectUri,
+        string $resourceUri,
+        string $tokenUrl,
```
localheinz: Not necessary for this PR, but maybe it makes sense to extract (and subsequently, inject here) a configuration object instead?

src/Http/Action/Security/SsoRedirectAction.php
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * @param HttpFoundation\Request $request
+     *
+     * @throws \OpenCFP\Infrastructure\Auth\UserExistsException
+     * @throws \Symfony\Component\Routing\Exception\RouteNotFoundException
+     * @throws \Symfony\Component\Routing\Exception\MissingMandatoryParametersException
+     * @throws \Symfony\Component\Routing\Exception\InvalidParameterException
+     * @throws \OpenCFP\Infrastructure\Auth\UserNotFoundException
+     * @throws \InvalidArgumentException
```
localheinz: Not sure if any of these exceptions are actually thrown (and not caught) anywhere in the scope of this method, can you check?

src/Http/Action/Security/SsoRedirectAction.php
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        /**
+         * What do we do with the access token?
+         * Check if the user exists
+         * -> if they don't, generate a random password and then create and activate an account
+         * manually log them in
+         * redirect them to their dashboard
+         */
+        $details = \json_decode((string) $response->getBody(), true);
```
localheinz: Should we check here whether the response body actually contained valid JSON, before moving on and assuming it does?

src/Http/Action/Security/SsoRedirectAction.php
```diff
@@ -0,0 +1,158 @@
...
+
+        $userResponse = $this->httpClient->get($this->resourceUri, [
+            'headers' => [
+                'Accept'        => 'application/json',
+                'Authorization' => 'Bearer ' . $details['access_token'],
```
localheinz: Similarly, should we verify whether the response body actually contains an `access_token` field?

src/Http/Action/Security/SsoRedirectAction.php
```diff
@@ -0,0 +1,158 @@
...
+            ],
+        ]);
+        $userDetails = \json_decode((string) $userResponse->getBody(), true);
```
localheinz: Again, should we check whether the response body actually contains valid JSON?

src/Http/Action/Security/SsoRedirectAction.php
```diff
@@ -0,0 +1,158 @@
...
+
+        try {
+            /** @var SentinelUser $user */
+            $user = $this->accounts->findByLogin($userDetails['email']);
```
localheinz: Should we also verify here whether the response body actually contains an `email` field, and that it is a `string`?

localheinz:
Thank you, @chartjes!

PR #1165, opened 2018-10-02
Allow for periods to be present in the joind.in username

### LINKED ISSUE #1161: Joind.in regex doesn't allow full stop in username

1. Have full stop in joind.in username
2. Try to enter joind.in URL on profile page

# Expected Result

*  No error

# Actual Result

* Error will be displayed "You did not enter a valid joind.in URL"

chrisforrence:
This PR fixes #1161 by permitting periods to be present in the joind.in link.

### DIFF
--- a/src/Http/Form/SignupForm.php
+++ b/src/Http/Form/SignupForm.php
@@ -203,7 +203,7 @@ class SignupForm extends Form
 
     public function validateUrl(): bool
     {
-        if (\preg_match('/^https:\/\/joind\.in\/user\/[a-zA-Z0-9\-_]{1,100}$/', $this->cleanData['url'])
+        if (\preg_match('/^https:\/\/joind\.in\/user\/[a-zA-Z0-9\-_\.]{1,100}$/', $this->cleanData['url'])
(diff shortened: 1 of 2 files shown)


src/Http/Form/SignupForm.php
```diff
@@ -203,7 +203,7 @@ public function validateLastName(): bool
 
     public function validateUrl(): bool
     {
-        if (\preg_match('/^https:\/\/joind\.in\/user\/[a-zA-Z0-9\-_]{1,100}$/', $this->cleanData['url'])
+        if (\preg_match('/^https:\/\/joind\.in\/user\/[a-zA-Z0-9\-_\.]{1,100}$/', $this->cleanData['url'])
```
localheinz: This will also allow users to provide user names which only consist of `.`s, are we ok with that?

localheinz: Well, I just registered at https://joind.in using `.` as a username.

Of course I can sign in, but not view my profile.

I guess we can merge this in, then.

localheinz:
Thank you, @chrisforrence!

PR #1167, opened 2018-10-03
Enhancement: Run composer validate and composer normalize in vendor target

localheinz:
This PR

* [x] runs `composer validate` and `composer normalize` in `vendor` target


chartjes:
Thanks @localheinz

localheinz:
Thank you, @chartjes!

Issue #1091, opened 2018-01-27
If a user tries to login with no account, redirect them to the create account page.


matthewtrask:
# Steps required to reproduce the problem

1. Attempt to Login
2. No account is found
3. Get flash banner

# Expected Result

* It would be nice if a user attempts to login in (misclick or something to the login page) to be redirected to the create account page rather then lingering on the login page which has no menu.

# Actual Result

* I stay on the login page with no menu/nav options.

PR #1168, opened 2018-10-04
Redirects users with emails not in the system to the signup page

chrisforrence:
This PR addresses #1091 by redirecting users who attempt to log in with an email address that is not present in the system to the signup page, while also pre-populating the email field. It also validates that some semblance of an email address is provided, and keeps the user on the login page if not.

### DIFF
--- a/resources/views/security/signup.twig
+++ b/resources/views/security/signup.twig
@@ -10,7 +10,9 @@
 
         <form method="post" action="{{ url('user_create') }}">
             <label class="text-white text-sm">Email</label>
-            <input type="email" name="email" placeholder="you@example.com">
+            <input type="email" name="email"
+                   value="{% if flash.old_email %}{{ flash.old_email | escape }}{% endif %}"
+                   placeholder="you@example.com">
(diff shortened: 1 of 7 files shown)


chrisforrence:
So just to clarify the new behavior, when a user attempts to log in with an email address that's not in the system, they'll actually get redirected to the registration page and the email address they had entered on the login page will get automatically added in the appropriate place on the registration form.

That being said, your point's definitely still valid; while it does provide a nicer user experience for good actors, it does pose a security problem in being able to see if a user has already signed up on that instance (and potentially sign up on their behalf).

As a potential way to mitigate the security risk, what if we added in an email confirmation step before a registration is complete? That way, the email address's owner would get a say in whether or not the registration attempt is valid or not, and that would prevent someone from spamming the registration form with people's known email addresses.

chartjes:
We used to have email confirmation as part of the process but removed it a while ago as it felt it was causing a bit of friction.

I am okay to merge this in and see if other folks think it is a problem.

PR #1171, opened 2018-10-09
Fix: Auto-wire Guzzle\ClientInterface implementation into SsoRedirectAction

localheinz:
This PR

* [x] auto-wires an implementation of `GuzzleHttp\ClientInterface` into `SsoRedirectAction`

Follows #1164.


resources/config/services/services.yml
```diff
@@ -38,7 +38,6 @@ services:
     bind:
       $clientId: '%opencfpcentral.clientId%'
       $clientSecret: '%opencfpcentral.clientSecret%'
-      $httpClient: '@http_client'
```
localheinz: We can safely remove the binding when we use an actual class (or interface) name for the service definition, as then the auto-wiring mechanism will use the service that corresponds to that class (or interface) name.

localheinz:
Thank you, @chartjes!

PR #1174, opened 2018-10-10
Split previous joind.in URL into joind.in username and separate URL

chrisforrence:
This PR addresses #1083 by splitting the joind.in URL field into 

- joind.in username, and
- URL

The URL field is no longer limited to joind.in domain.

This is the re-submission of #1166; I'll carry over the comments from the previous pull request.


migrations/20181002234013_create_user_joind_in_username_column.php
```diff
@@ -0,0 +1,73 @@
...
+    {
+        // Create joindin_username
+        $this->table('users')
+            ->addColumn('joindin_username', 'string', ['null' => true])
+            ->update();
+
+        // Go through each record in user, strip out (https://joind.in/user/) and copy to joindin_username
+        $joindInRegex = '/^https:\/\/joind\.in\/user\/(.{1,100})$/';
+
+        $users = EloquentUser::all();
```
chrisforrence: localheinz, 7 hours ago

> Personally, I'm not a fan of using an ORM in database migrations, as when the migration is run
> 
> * a model or entity may not even exist anymore
> * a model or entity might not even work with the schema at the state before the migration is run
> 
> Now, since I haven't used Laravel at all, and apparently models have been used in migrations within this project before, so I'm not going to make an issue out of it, but for the record, I would prefer not to use models or entities in a database migration.
> 
> * at the time the migration is run,

chartjes 6 hours ago

> We are already using Eloquent in other migrations, so I do not have a big problem with it.

src/Http/Form/SignupForm.php
```diff
@@ -201,15 +202,29 @@ public function validateLastName(): bool
         return $validationResponse;
     }
 
+    public function validateJoindInUsername(): bool
+    {
+        if (!isset($this->cleanData['joindin_username'])
+            || $this->cleanData['joindin_username'] == ''
+            || \preg_match('/^[a-zA-Z0-9\-_\.]{1,100}$/', $this->cleanData['joindin_username'])
```
chrisforrence: localheinz 7 hours ago

> Should we make sure we are dealing with a `string` here before applying `preg_match()`?

resources/views/admin/speaker/view.twig
```diff
@@ -15,6 +15,15 @@
             <div>
                 {% if speaker.company %}<span class="mr-3"><i class="fa fa-building"></i> {{ speaker.company }}</span>{% endif %}
                 {% if speaker.twitter %}<span class="mr-3"><a class="hover:text-brand" target="_blank" href="{{ speaker.twitterUrl }}" rel="noopener noreferrer"><i class="fa fa-twitter"></i> @{{ speaker.twitter }}</a></span>{% endif %}
+                {% if speaker.joindInUsername %}
+                    <span class="mr-3">
+                        <a class="hover:text-brand" target="_blank"
+                           href="{{ speaker.joindInUrl }}"
```
chrisforrence: localheinz 7 hours ago

> Unless I am terribly mistaken, and to answer your question from the body of the pull request, I think it should be possible to use
> 
> ```diff
> diff --git a/resources/views/admin/speaker/view.twig b/resources/views/admin/speaker/view.twig
> index ace150ce..3d9e23fe 100644
> --- a/resources/views/admin/speaker/view.twig
> +++ b/resources/views/admin/speaker/view.twig
> @@ -18,9 +18,9 @@
>                  {% if speaker.getJoindInUsername() %}
>                      <span class="mr-3">
>                          <a class="hover:text-brand" target="_blank"
> -                           href="https://joind.in/user/{{ speaker.getJoindInUsername() }}"
> +                           href="https://joind.in/user/{{ speaker.joindInUsername }}"
>                             rel="noopener noreferrer">
> -                            <i class="fa fa-commenting-o"></i> {{ speaker.getJoindInUsername() }}
> +                            <i class="fa fa-commenting-o"></i> {{ speaker.joindInUsername }}
>                          </a>
>                      </span>
>                  {% endif %}
> ```
> 
> here instead. Have you tried?
> 
> Maybe it also makes sense to extract a method and shift the construction of the URL to `SpeakerProfile`, what do you think?

chartjes 6 hours ago

> I like the idea of having `SpeakerProfile` generate this too since I recently merged in a PR that moved the generation of Twitter URL's to `SpeakerProfile` as well

chrisforrence 6 hours ago

> Haha, I spent half an hour on stream trying to figure out why speaker.joindInUsername hadn't been working :p I can certainly give it another crack tonight though!

> I can definitely be on board with either constructing the URL through SpeakerProfile or on a JoindIn helper (and something similar for Twitter as well)

chrisforrence 6 hours ago

> Ah, well then, that answers the second part!

tests/Unit/Domain/Speaker/SpeakerProfileTest.php
```diff
@@ -338,6 +338,96 @@ public function getTwitterUrlReturnsTwitterUrlWhenTwitterPropertyIsNeitherHidden
...
+
+        $profile->getJoindInUsername();
+    }
+
+    /**
+     * @test
+     */
+    public function getJoindInUsernameReturnsJoindInUsernameIfPropertyIsNotHidden(): void
+    {
+        $joindinUsername = $this->faker()->userName;
```
chrisforrence: localheinz 7 hours ago

> Can we use `camelCase` here instead of `snake_case`?

chartjes 6 hours ago

> Yep, `camelCase` please

chrisforrence: Fixed

resources/views/admin/speaker/view.twig
chrisforrence: I've got the SpeakerProfile now creating a joind.in URL as a getter, and I figured out why I wasn't able to get it working before (I had been trying to refer to it with `joindin_username` with the underscore, not `joindInUsername`). Should be good now!

PR #1179, opened 2018-10-12
Require airport code entry if requesting travel assistance

iansltx:
Since flight prices vary significantly between a local hop and an intercontinental jaunt, requiring originating airport code will help conf organizers gauge what they can afford without having to manually ask speakers what their airport code is after the fact.

With the above said, I totally understand if we don't want this requirement to be on by default, in which case I can throw it behind a config variable (which we'll definitely have turned on for Longhorn).

Also s/Code of conduct/Code of Conduct; if I absolutely need to back out that change to get this PR through, I'll do it :)

### DIFF
--- a/src/Http/Form/SignupForm.php
+++ b/src/Http/Form/SignupForm.php
@@ -62,7 +62,25 @@ class SignupForm extends Form
             $validSpeakerBio = $this->validateSpeakerBio();
         }
 
-        return $this->validateEmail() && $validPasswords && $this->validateFirstName() && $this->validateLastName() && $this->validateUrl() && $validSpeakerInfo && $validSpeakerBio && $this->validateSpeakerPhoto() && $agreeCoc && $this->validateJoindInUsername();
+        return $this->validateEmail() && $validPasswords &&
+            $this->validateFirstName() && $this->validateLastName() && $this->validateUrl() &&
+            $validSpeakerInfo && $validSpeakerBio && $this->validateSpeakerPhoto() && $agreeCoc &&
+            $this->validateJoindInUsername() && $this->validateTransportationRequests();
+    }
+
+    public function validateTransportationRequests()
+    {
+        if (!isset($this->taintedData['transportation']) || !$this->taintedData['transportation']) {
+            return true;
+        }
+
+        if (!isset($this->taintedData['airport']) || !$this->taintedData['airport']) {
+            $this->addErrorMessage("Since you specified that you'll need help with transportation costs, we need to know which airport you'll be departing from. You may specify additional information (e.g. alternate airports) in the Additional Notes field.");
+
+            return false;
+        }
+
+        return true;
     }
 
     public function validateSpeakerPhoto(): bool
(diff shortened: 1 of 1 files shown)


src/Http/Form/SignupForm.php
```diff
@@ -62,7 +62,25 @@ public function validateAll(string $action = 'create'): bool
             $validSpeakerBio = $this->validateSpeakerBio();
         }
 
-        return $this->validateEmail() && $validPasswords && $this->validateFirstName() && $this->validateLastName() && $this->validateUrl() && $validSpeakerInfo && $validSpeakerBio && $this->validateSpeakerPhoto() && $agreeCoc && $this->validateJoindInUsername();
+        return $this->validateEmail() && $validPasswords &&
```
localheinz: Great idea to start wrapping here, the line has already become quite long.

Would you mind wrapping after each element?

```diff
diff --git a/src/Http/Form/SignupForm.php b/src/Http/Form/SignupForm.php
index eeea60b1..8d2630d6 100755
--- a/src/Http/Form/SignupForm.php
+++ b/src/Http/Form/SignupForm.php
@@ -62,10 +62,17 @@ class SignupForm extends Form
             $validSpeakerBio = $this->validateSpeakerBio();
         }

-        return $this->validateEmail() && $validPasswords &&
-            $this->validateFirstName() && $this->validateLastName() && $this->validateUrl() &&
-            $validSpeakerInfo && $validSpeakerBio && $this->validateSpeakerPhoto() && $agreeCoc &&
-            $this->validateJoindInUsername() && $this->validateTransportationRequests();
+        return $this->validateEmail()
+            && $validPasswords
+            && $this->validateFirstName()
+            && $this->validateLastName()
+            && $this->validateUrl()
+            && $validSpeakerInfo
+            && $validSpeakerBio
+            && $this->validateSpeakerPhoto()
+            && $agreeCoc
+            && $this->validateJoindInUsername()
+            && $this->validateTransportationRequests();
     }

     public function validateTransportationRequests()
```

src/Http/Form/SignupForm.php
```diff
@@ -62,7 +62,25 @@ public function validateAll(string $action = 'create'): bool
...
+            $this->validateFirstName() && $this->validateLastName() && $this->validateUrl() &&
+            $validSpeakerInfo && $validSpeakerBio && $this->validateSpeakerPhoto() && $agreeCoc &&
+            $this->validateJoindInUsername() && $this->validateTransportationRequests();
+    }
+
+    public function validateTransportationRequests()
```
localheinz: Can we add a return type declaration here?

src/Http/Form/SignupForm.php
```diff
@@ -298,7 +316,7 @@ private function validateAgreeCoc(): bool
             return true;
         }
 
-        $this->addErrorMessage('You must agree to abide by our code of conduct in order to submit');
+        $this->addErrorMessage('You must agree to abide by our Code of Conduct in order to submit talks.');
```
localheinz: A bit out of scope, but fine!

iansltx:
Yep, I'll see if I can handle all of the above over the weekend.

iansltx:
Welp, php-cs-fixer insists on one long line on the return rather than something sane, so I can't commit the multi-line version without failing Travis. Guess I'll stick with the one long line, as I don't want to break some other sniff by editing .php_cs.dist as part of this PR.

iansltx:
@localheinz back to you

PR #1176, opened 2018-10-12
Fix Twitter link

iansltx:
Fixes duplicate : in Twitter profile link that was causing Twitter profile links to not work.

Have a couple more PRs incoming for other minor quality of life fixes.

PR #1172, opened 2018-10-09
Enhancement: Move responsibilty to construct Twitter URL upstream

localheinz:
This PR

* [x] moves the responsibility to construct a Twitter URL from view scripts to `SpeakerProfile`

Related to the suggestion in https://github.com/opencfp/opencfp/pull/1166/files#r223808752.

/cc @chrisforrence

PR #1180, opened 2018-10-12
Fix: Add missing test annotation

localheinz:
This PR

* [x] adds a missing `@test` annotation

Follows #1172.
Follows #1176.


localheinz:
Thank you, @chartjes!

PR #1184, opened 2018-10-12
Enhancement: Extract AutoReview test-suite and split tests

localheinz:
This PR

* [x] extracts an `auto-review` test-suite and splits tests out of `ProjectCodeTest`

💁‍♂️ In addition to what we already have it could probably be useful to verify test code in regard to `public` methods that are not used as data providers and aren't prefixed with `test` (perhaps they should be tests and are missing a `@test` annotation)


localheinz:
Thank you, @chartjes!

PR #1185, opened 2018-10-12
Use global date format for CFP end date in submission email

iansltx:
Everything else uses the global date format, so it looks like this is a bit of an omission as it stands.

### DIFF
--- a/resources/config/config.yml
+++ b/resources/config/config.yml
@@ -25,6 +25,7 @@ wouterj_eloquent:
 twig:
   default_path: '%kernel.project_dir%/resources/views'
   debug: '%kernel.debug%'
+  base_template_class: 'OpenCFP\Infrastructure\Templating\Template'
   globals:
     site:       '%config.application%'
     talkHelper: '@OpenCFP\Http\View\TalkHelper'
--- a/resources/views/emails/talk_submit.twig
+++ b/resources/views/emails/talk_submit.twig
@@ -5,7 +5,7 @@
 
 {% block body_html %}
 <p>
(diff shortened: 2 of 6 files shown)


localheinz:
@chartjes @iansltx 

I have labeled this as a bug as I believe that an inconsistency in formatting dates is in fact a bug from the user’s perspective.

iansltx:
The reason I injected it at the controller level is I wasn't sure that Twig was able to access the date format string in that particular context. I'll check again shortly.

iansltx:
Looks like that date is globally injected, so I'm going to completely re-roll this.

iansltx:
False alarm; the "site" variables don't appear to exist in the Twig instance that's being used by the emails for some reason, so it's falling back to the Twig-declared date format. Tracking this down further now.

iansltx:
Okay; the issue here is that we're using an internal Twig method to render each block, rather than rendering the entire template at once, for the email. There are a few easier solutions than what I'm about to attempt:

1. Split the email into text and HTML components
2. Manually inject the date format, as I was doing before
3. Manually inject site globals, somewhat similarly to what I was doing before

What I'm going to try is to have a patched Twig_Environment that allows for rendering a block with the global context added in. Wish me luck.

iansltx:
Quite a bit more work involved in the above, but in return emails can now include global Twig vars in their blocks without explicitly passing things in, and those vars will actually be there (since emails are rendered one block at a time). I've confirmed that this works exactly as expected on our instance (the correct date format is respected).

localheinz:
@iansltx 

Don't worry too much, we can still follow up on this! 👍

iansltx:
@localheinz This now passes CI due to my fix of a flakey test elsewhere. What else do I need to do to get this merged?

PR #1187, opened 2018-10-12
Redirect admins and reviewers to their respective dashboards on-login

### LINKED ISSUE #1183: Enhancement: push admins to admin dashboard right after login
My thought here (from experience) is that admins aren't going to be submitting abstracts to their own conferences, so when they log in, the admin dashboard is a *much* more useful place to land.

I'm not sure of the effort required to make this tweak, but if the above sounds like something merge-able if I get a working PR for it, I'll definitely have a go at it, as I'll definitely benefit from the feature.

iansltx:
Resolves #1183

### DIFF
--- a/src/Http/Action/Security/LogInAction.php
+++ b/src/Http/Action/Security/LogInAction.php
@@ -61,6 +61,8 @@ final class LogInAction
                 $request->get('email'),
                 $request->get('password')
             );
+
+            $user = $this->authentication->user();
         } catch (UserNotFoundException $exception) {
             $flash = [
                 'type'  => 'error',
@@ -91,7 +93,7 @@ final class LogInAction
                 $content,
                 HttpFoundation\Response::HTTP_BAD_REQUEST
             );
-        } catch (Services\AuthenticationException $exception) {
+        } catch (Services\AuthenticationException | Services\NotAuthenticatedException $exception) {
             $flash = [
                 'type'  => 'error',
                 'short' => 'Error',
@@ -109,7 +111,13 @@ final class LogInAction
             );
         }
 
-        $url = $this->urlGenerator->generate('dashboard');
+        if ($user->hasAccess('admin')) {
+            $url = $this->urlGenerator->generate('admin');
+        } elseif ($user->hasAccess('reviewer')) {
+            $url = $this->urlGenerator->generate('reviewer');
+        } else {
+            $url = $this->urlGenerator->generate('dashboard');
+        }
 
         return new HttpFoundation\RedirectResponse($url);
     }
--- a/tests/Unit/Http/Action/Security/LogInActionTest.php
+++ b/tests/Unit/Http/Action/Security/LogInActionTest.php
@@ -110,14 +110,9 @@ final class LogInActionTest extends Framework\TestCase
         $this->assertSame($content, $response->getContent());
     }
 
-    /**
-     * Data provider for redirectsToDashboardIfAuthenticationSucceeded
-     *
-     * @return array
-     */
     public function userProvider(): array
     {
-        return [ // admin check, reviewer check, expected redirect
+        return [ // admin check result, reviewer check result (null means expected to not be checked), expected redirect
             [false, false, 'dashboard'],
             [false, true, 'reviewer'],
             [true, null, 'admin'],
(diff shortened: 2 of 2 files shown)


tests/Unit/Http/Action/Security/LogInActionTest.php
```diff
@@ -109,10 +110,29 @@ public function rendersLogInFormIfAuthenticationFailed()
...
     /**
      * @test
+     *
+     * @param bool      $isAdmin
+     * @param null|bool $isReviewer
+     * @param string    $shouldRedirectTo
+     * @dataProvider    userProvider
      */
-    public function redirectsToDashboardIfAuthenticationSucceeded()
+    public function redirectsToDashboardIfAuthenticationSucceeded($isAdmin, $isReviewer, $shouldRedirectTo)
```
localheinz: Can we add type and return-type declarations here?

src/Http/Action/Security/LogInAction.php
```diff
@@ -61,6 +61,8 @@ public function __invoke(HttpFoundation\Request $request): HttpFoundation\Respon
                 $request->get('email'),
                 $request->get('password')
             );
+
+            $user = $this->authentication->user();
```
localheinz: Can we move this line out of the `try`-and-`catch` block, perhaps just before we actually need the user?

tests/Unit/Http/Action/Security/LogInActionTest.php
```diff
@@ -109,10 +110,29 @@ public function rendersLogInFormIfAuthenticationFailed()
         $this->assertSame($content, $response->getContent());
     }
 
+    /**
+     * Data provider for redirectsToDashboardIfAuthenticationSucceeded
+     *
+     * @return array
+     */
```
localheinz: Can we remove the entire DocBlock?

When we rename the test method, it becomes outdated, and it should be clear from the test method itself that it is used there.

tests/Unit/Http/Action/Security/LogInActionTest.php
```diff
@@ -109,10 +110,29 @@ public function rendersLogInFormIfAuthenticationFailed()
...
+    public function userProvider(): array
+    {
+        return [ // admin check, reviewer check, expected redirect
+            [false, false, 'dashboard'],
```
localheinz: For sake of completeness, should we also provide data for another combination

* `[true, true, 'admin']`

?

tests/Unit/Http/Action/Security/LogInActionTest.php
```diff
@@ -109,10 +110,29 @@ public function rendersLogInFormIfAuthenticationFailed()
...
+            [false, true, 'reviewer'],
+            [true, null, 'admin'],
```
localheinz: I think this should be `[true, false, 'admin']` instead (see method signature of `User::hasAccess()`.

src/Http/Action/Security/LogInAction.php
iansltx: authentication->user() could theoretically an exception (hence the new exception I'm catching further down), hence putting this inside the try-catch. Storm yelled at me when I put it outside the try-catch, where you're suggesting.

tests/Unit/Http/Action/Security/LogInActionTest.php
iansltx: I was following what appeared to be a convention elsewhere. Easy enough though.

tests/Unit/Http/Action/Security/LogInActionTest.php
iansltx: Nope; see below.

tests/Unit/Http/Action/Security/LogInActionTest.php
iansltx: Negative. Because we're using an else-if on the actual impl, the is-a-reviewer check is never called. In the body of the test, the null here avoids telling Prophecy to expect a call to hasAccess for the reviewer role, because the admin role being there actually short-circuits that.

tests/Unit/Http/Action/Security/LogInActionTest.php
iansltx: Done.

iansltx:
Weird that the tests failed since I didn't touch anywhere near that part of the code. Just rebased on top of the latest; hopefully the error was intermittent as it's not obvious why it failed.

iansltx:
@localheinz Now that issues on master have been resolved, tests pass here. What remains to get this merged?

Issue #1194, opened 2018-10-19
Broken integration test suite has me scratching my head


chartjes:
# Steps required to reproduce the problem

1. Check out master branch
2. Run 'make integration'

# Expected Result

* All integration tests pass

# Actual Result

* Integration test suite has failures

I had to alter the phpunit.xml file to disable the recently-merged PHPUnit pretty-print results plugin as it was generating a sprintf() error, no idea what is up there. These are the tests that are currently failing on my laptop. Running PHP 7.2.11

There were 4 failures:

1) OpenCFP\Test\Integration\Http\Action\Talk\EditActionTest::seeEditPageWhenAllowed
Failed asserting that the response body "<!DOCTYPE html>
<html>
    <head>
        <meta charset="UTF-8" />
        <meta http-equiv="refresh" content="0;url=/talk/92" />

        <title>Redirecting to /talk/92</title>
    </head>
    <body>
        Redirecting to <a href="/talk/92">/talk/92</a>.
    </body>
</html>" contains "Dicta debitis accusamus in corrupti et.".
Failed asserting that '<!DOCTYPE html>\n
<html>\n
    <head>\n
        <meta charset="UTF-8" />\n
        <meta http-equiv="refresh" content="0;url=/talk/92" />\n
\n
        <title>Redirecting to /talk/92</title>\n
    </head>\n
    <body>\n
        Redirecting to <a href="/talk/92">/talk/92</a>.\n
    </body>\n
</html>' contains "Dicta debitis accusamus in corrupti et.".

/Users/chartjes/Sites/opencfp/tests/Helper/ResponseHelper.php:60
/Users/chartjes/Sites/opencfp/tests/Integration/Http/Action/Talk/EditActionTest.php:99

2) OpenCFP\Test\Integration\Infrastructure\Auth\SentinelUserTest::checkResetPasswordCodeWorks
Failed asserting that false is true.

/Users/chartjes/Sites/opencfp/tests/Integration/Infrastructure/Auth/SentinelUserTest.php:78

3) OpenCFP\Test\Integration\Infrastructure\Auth\SentinelUserTest::getResetPassWordCodeWorks
Failed asserting that false is true.

/Users/chartjes/Sites/opencfp/tests/Integration/Infrastructure/Auth/SentinelUserTest.php:94

4) OpenCFP\Test\Integration\Infrastructure\Auth\SentinelUserTest::attemptResetPasswordWorks
Failed asserting that false is true.

[...]

PR #1196, opened 2018-10-19
Set testing CFP close date to far into future

chrisforrence:
This PR addresses #1194 by setting the CFP closing date for testing far into the future (which was causing issues when testing if talks could be edited).

cc @chartjes

### DIFF
--- a/config/testing.yml.dist
+++ b/config/testing.yml.dist
@@ -4,7 +4,7 @@ application:
   email: contact@openconf.com
   eventurl: http://localhost
   event_location: Miami, FL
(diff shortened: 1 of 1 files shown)


config/testing.yml.dist
```diff
@@ -4,7 +4,7 @@ application:
   email: contact@openconf.com
   eventurl: http://localhost
   event_location: Miami, FL
-  enddate: Oct. 14th, 2018
```
localheinz: 😆

localheinz:
Thank you, @chrisforrence!

PR #1201, opened 2018-12-14
enable phpstan

smoench:
Enable phpstan in travis (ci builds) and be compatible with level 1.

I had to update infection as well. Is infection still in use or should be kept?


chartjes:
I don't think having phpstan running all the time is what I want for the project. I believe it is a great tool to help during development.

smoench:
Okay, fair enough. But what about the improvements I did?

chartjes:
If you resubmit the PR with enabling phpstan for Travis removed, I will take a look.

Makefile
```diff
@@ -26,15 +26,15 @@ database: test-env vendor
...
 
 integration: test-env vendor database cache
 	vendor/bin/phpunit --testsuite integration
 
 stan: vendor
-	vendor/bin/phpstan analyse --level=2 src tests
+	vendor/bin/phpstan analyse
 
-test: auto-review integration unit
+test: auto-review integration unit stan
```
localheinz: What do you think about adding `stan` to the `it` target instead?

Then it will be run as well when we run

```
$ make it
```

or 

```
$ make
```

chartjes:
After sleeping on the idea, I am okay with having phpstan be part of the Travis process. An extra set of eyes (as it were) seems like a good thing. I'm never mad when some other automated process tells me I missed something or made a mistake. That's what they are there for!

So @smoench, my apologies, and I will be happy to accept the original PR and allow it to be merged.

smoench:
@chartjes :) Sure, I will reapply my changes. Would you prefer an extra stage, like @localheinz proposed, or should it be a part of the `style`-stage?

chartjes:
I am okay with it being part of the style stage, since that is what we're trying to enforce here: writing code in a specific way via style enforcement and now adding some static code analysis

phpstan.neon.dist
```diff
@@ -0,0 +1,7 @@
+parameters:
+    level: 1
+    paths: [src, tests]
```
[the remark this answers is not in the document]
smoench: You are welcome :)

Makefile
smoench: done

PR #1218, opened 2019-10-01
Fixed joind in username column migration

carnage:
This PR

* [x] Fixes error when running migration if any of the speaker urls is null in the database


chartjes:
Thanks @carnage

PR #1220, opened 2019-10-22
Add raw filter to talk titles

### LINKED ISSUE #1217: Inconsistent encoding for talk titles between dahsboard and talk view
I noticed this minor display issue during a talk submission where I used the characters `<3` as a heart. On the dashboard and in the talk edit view this shows correctly, but on the talk view page it renders as `&lt;3`.

I haven't confirmed but I presume the database is storing it encoded, thus when `&lt;` is output with `|raw` it is rendered by the browser as `<`. On the view however the `&` gets encoded itself as `&amp;`, meaning the browser receives `&amp;lt;` and renders it as `&lt;`

https://github.com/opencfp/opencfp/blob/965a1210fa8b69713be0bd709e2d0ddd2be845b2/resources/views/talk/view.twig#L3

https://github.com/opencfp/opencfp/blob/965a1210fa8b69713be0bd709e2d0ddd2be845b2/resources/views/dashboard.twig#L82

The most obvious solution here would be to make the view also do a `|raw` output, as it seems the back end validation both uses strip tags and also HTML encodes before saving.

I can submit this as a PR if you're happy for it?

chrisforrence:
This PR simply adds a raw filter to the talk titles where it is missing.

Fixes #1217.

### DIFF
--- a/resources/views/admin/talks/view.twig
+++ b/resources/views/admin/talks/view.twig
@@ -39,7 +39,7 @@
             {% if talk.otherTalks | length > 0 %}
                 <h3><i class="fa fa-bullhorn"></i> Other Talks...</h3>
                 {% for talk in talk.otherTalks %}
-                    <p><a class="text-brand underline" href="{{ url('admin_talk_view', { id: talk.id }) }}">{{ talk.title }}</a></p>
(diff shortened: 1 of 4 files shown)


chartjes:
Thanks @chrisforrence

Issue #1199, opened 2018-12-04
Enhancement: Store original-size (or at least larger) version of speaker profile photo


iansltx:
Right now we drop all photos to 250px, and from a quick look around it appears that that's not readily configurable?

In any case, with modern high-res phones/computers, we want higher-res photos, if not for use in OpenCFP then for pulling into conf websites etc.

My thought here would be to save two versions of an image on upload: one at the current resize (250px), one at either full resolution or scaled down less aggressively, e.g. down to 2048px on a side or similar.

If that sounds reasonable (and I'm not missing something that's already implemented) I should have time to work on this this coming weekend. I should've figured this out before our CFP close drew near, but better late than never I guess.

chartjes:
Sorry for the delay in responding, that looks like a good enhancement

PR #1221, opened 2019-10-22
Adds the ability to store larger uploaded files

chrisforrence:
This PR removes the hard-coded image resize (250px) and allows it to be configured at the application configuration level. Furthermore, this also allows the application to generate 240x240 thumbnails to be used on the dashboard (instead of cramming a potentially-large file into a 120x120 space). If a thumbnail is not available for a particular image, an onerror handler swaps the image source out with the non-thumbnail version.

Related to #1199.


chartjes:
Thanks @chrisforrence

PR #1229, opened 2019-11-18
Fix github actions badge

linuxjuggler:
This PR fixes the broken badge URL and use the name of the action instead of the file name. Also, remove the line break between the badges.

PS: generating GH Actions badge is a slow process, thats why the readme file may take a bit to load.


chartjes:
Thank you @linuxjuggler

PR #1236, opened 2019-11-30
Add PHP 7.4 to the test matrice

tdutrion:
This PR adds PHP 7.4 to the test matrice on Gitlab Actions


chartjes:
The test suite currently generates a "message bus" error when run under 7.4.1 available on my laptop via homebrew. Waiting for a version of PHP 7.4 that my tests run under before I add it in.

PR #1240, opened 2019-12-22
Refactor http foundation calls

N-M:
This PR

- removes 2 unused classes

- changes the usage of get method of Request class. Change of this increases the readability and speed of getting the parameters. This is provided in the description of the method 

> "This method is mainly useful for libraries that want to provide some flexibility. If you don't need the flexibility in controllers, it is better to explicitly get request parameters from the appropriate
> public property instead (attributes, query, request)."

@chartjes if you agree with this PR I will fix remaining tests that are failing.


chartjes:
@N-M Please go ahead and fix all the broken tests! Greatly appreciated

chartjes:
Thanks so much @N-M for the refactor

PR #1242, opened 2019-12-31
Update verbiage on speaker package page

cookieguru:
I can't remember the last time I saw a VGA cable connected to a projector, or a laptop that had a VGA port or adapter.  I also haven't seen a conference organizer change this sentence _ever_.

And maybe check back in a few years to see how USB-C/Thunderbolt 3 is doing.


chartjes:
Thanks @cookieguru

PR #1243, opened 2020-01-16
Fix configs to use credentials defined in parameters instead of hardcoded ones

samdark:
Currently speakers list doesn't work if database isn't default one.


chartjes:
Thanks @samdark

PR #1246, opened 2020-03-11
Implements and tests anonymized reviews setting

### LINKED ISSUE #1075: enable blind review
this is a feature request to make it possible to do blind reviews of talks, ie. where the reviewer of submissions doesn’t see who submitted the talk. note I don’t run any conference myself but I think it would be good to make it easy for conferences to be able to do this.

that being said I know some conferences that require speakers to do 2 talks in order to be able to afford paying for travel expenses, maybe this should also be taken into account, ie. making it still possible to see if a prospective speaker submitted two interesting talks while making the final decisions on the schedule. but this is obviously a second step in this feature and maybe its not even relevant enough to even bother.

SandyS1:
Some conferences have attempted to remove a source of bias in speaker selection by hiding the names of speakers from reviewers. This PR introduces a new configuration option to enable this behavior. The default behavior is unchanged, so it is an opt-in solution for users.

Fixes #1075.

### DIFF
--- a/config/ci.yml.dist
+++ b/config/ci.yml.dist
@@ -15,6 +15,7 @@ application:
   coc_link: http://confcodeofconduct.com
   venue_image_path: /assets/img/venue.jpg
   user_image_size: 300
+  anonymized_reviews: false
 
 database:
   host: localhost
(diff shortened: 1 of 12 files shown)


SandyS1:
The tests pass on my own system, so I assume this is something that may succeed if the tests are re-run and the mock uses different sample text.

chartjes:
I have noticed this happens once in a while with some other tests I wrote -- I need to investigate those failures and see if I can fix it.

chartjes:
Thanks @SandyS1

SandyS1:
Of course, quite welcome!

PR #1255, opened 2020-10-09
Feat: Remake the dockerfile configuration

BoutValentin:
Need to work on nginx compatibilities

This PR

* [x] 
* [ ] 
* [ ] 

Follows #.
Related to #.
Fixes #.


chartjes:
Thanks for cleaning this up.

toofff:
NOOOOOOOOO, this PR is Draft, is not finish

chartjes:
Ah, it's okay! Just keep working on it and I will deal with anyone else complaining about it not working.

toofff:
@chartjes OK we continue but we still had a lot of work to do on it, it would be better to do a revert of commit on your side, because it will take us a little more time.

